PR #3, opened 2025-07-16
Enable Windows in CI, upgrade to Swift 6.2, add devcontainer support

tothambrus11:
Now Swift 6.2 is officially released, so we can use it here.

### DIFF
--- a/Sources/FrontEnd/Syntax/Builtins/BuiltinFunction.swift
+++ b/Sources/FrontEnd/Syntax/Builtins/BuiltinFunction.swift
@@ -1562,7 +1562,7 @@ private func + <A, B>(
 
 /// Returns a parser that returns an instance of `T` if it can be built by consuming the next
 /// element in the stream.
-private func take<T: RawRepresentable>(
+private func take<T: RawRepresentable & SendableMetatype>(
   _: T.Type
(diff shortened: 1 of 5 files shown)


Sources/FrontEnd/Syntax/Builtins/BuiltinFunction.swift
```diff
@@ -1569,13 +1569,13 @@ private func + <A, B>(
...
 /// element in the stream.
-private func take<T: RawRepresentable>(
+private func take<T: RawRepresentable & Sendable>(
   _: T.Type
 ) -> BuiltinFunctionParser<T> where T.RawValue == String {
   { (stream: inout ArraySlice<Substring>) -> T? in
     stream.popFirst().flatMap({ (x) in T(rawValue: .init(x)) })
   }
-}
+}
```
kyouko-taiga, line 1578:
```suggestion
}
```

kyouko-taiga (commented):
Is this PR in addition with the other about LLVM, or as an alternative? It seems some of the changes here look like some of the changes there.

tothambrus11:
@kyouko-taiga yes, the LLVM PR is based on top of this, so I recommend reviewing only that one.

tothambrus11:
Update: I changed the order of commits, now this one can be merged before #5.

PR #5, opened 2025-07-17
Add LLVM dependency

tothambrus11:


Tests/CodeGenTests/ExampleCodeGenerationTests.swift
```diff
@@ -0,0 +1,480 @@
+// Copied from SwiftyLLVM test suite.
+import SwiftyLLVM
+import XCTest
+
+final class CodeGenerationTests: XCTestCase {
```
kyouko-taiga: Why do we need to copy this file from SwiftyLLVM?

kyouko-taiga (changes requested):
Nice work. I don't understand why we have to copy test cases from SwiftyLLVM.

tothambrus11:
Superseded by this PR: https://github.com/hylo-lang/hylo-new/pull/14 (mostly the same content, with added comment)

PR #13, opened 2025-12-07
Language server support

tothambrus11:
This PR exposes the necessary internals from the FrontEnd for the language server.

- Expose the frontend and standard library as library products.
- Support source files with given in-memory contents.
- Expose source container properties for reading through accessors in `Program`.
- Expose querying givens of a type-checked program.
- Expose bundled standard library path (related PR: #75 )

### DIFF
--- a/Sources/FrontEnd/Files/SourceFile.swift
+++ b/Sources/FrontEnd/Files/SourceFile.swift
@@ -47,7 +47,18 @@ public struct SourceFile: Hashable, Sendable {
 
   /// Creates a source file with the contents of the file at `path`.
   public init(contentsOf path: URL) throws {
-    try self.init(name: .local(path), contents: .init(contentsOf: path, encoding: .utf8))
+    let contents = try String(contentsOf: path, encoding: .utf8)
+    self.init(localPath: path, contents: contents)
+  }
+
+  /// Creates a source file with given `contents`, representing a local file at `p`.
+  ///
+  /// - Note: `p` can be non-existent or contain different contents on disk,
+  ///   e.g. when called by the LSP.
+  /// - Requires: `p` has a file name component.
+  public init(localPath p: URL, contents: String) {
+    precondition(!p.pathComponents.isEmpty)
+    self.init(name: .local(p), contents: contents)
   }
 
   /// Creates a virtual source file with the given contents.
--- a/Sources/FrontEnd/Files/SourceFile.swift
+++ b/Sources/FrontEnd/Files/SourceFile.swift
@@ -143,6 +143,11 @@ public struct SourceFile: Hashable, Sendable {
     return (lineNumber, columnNumber)
   }
 
+  /// Returns the index in `text` corresponding to the 1-based `line` and `column`.
+  public func index(line: Int, column: Int) -> Index {
+    text.index(properties.lineStarts[line - 1], offsetBy: column - 1)
+  }
+
   /// Calls `action` on each source file URL in `directory` having the extension `pathExtension`.
   public static func forEachURL(
     in directory: URL, withPathExtension pathExtension: String = "hylo",
--- a/Sources/FrontEnd/Files/SourceFile.swift
+++ b/Sources/FrontEnd/Files/SourceFile.swift
@@ -4,12 +4,16 @@ import Foundation
 import Utilities
 
 /// A source file.
+/// 
+/// - Invariant: `name` contains an absolute path.
+/// - Note: A local file name doesn't imply that the file actually exists on disk or that we can 
+///   re-read its contents. It might be supplied by the LSP without having the file saved.
 public struct SourceFile: Hashable, Sendable {
 
   /// The internal representation of a source file.
   private final class Properties: Hashable, Sendable {
 
-    /// The name of the file that the source came from.
+    /// The absolute name of the file that the source came from.
     let name: FileName
 
     /// The contents of the file.
@@ -40,32 +44,27 @@ public struct SourceFile: Hashable, Sendable {
   /// The properties of `self`.
   private let properties: Properties
 
-  /// Creates a source file with the given name and contents.
-  private init(name: FileName, contents: String) {
-    self.properties = .init(name: name, text: contents)
+  /// Creates a source file with given `contents`, associated with the file `n`.
+  ///
+  /// - Requires: `n` has a file name component.
+  public init(name n: FileName, contents: String) {
+    precondition(n.url.pathComponents.count >= 2) // The 0th component is `/`, followed by the real components.
+    self.properties = .init(name: n.absolute, text: contents)
   }
 
-  /// Creates a source file with the contents of the file at `path`.
+  /// Creates a local source file with the contents of the file at `path`.
   public init(contentsOf path: URL) throws {
     let contents = try String(contentsOf: path, encoding: .utf8)
-    self.init(localPath: path, contents: contents)
+    self.init(name: .local(path), contents: contents)
   }
 
-  /// Creates a source file with given `contents`, representing a local file at `p`.
-  ///
-  /// - Note: `p` can be non-existent or contain different contents on disk,
-  ///   e.g. when called by the LSP.
-  /// - Requires: `p` has a file name component.
-  public init(localPath p: URL, contents: String) {
-    precondition(!p.pathComponents.isEmpty)
-    self.init(name: .local(p), contents: contents)
-  }
-
-  /// Creates a virtual source file with the given contents.
+  /// Creates a virtual source file with the given `contents`.
   public init(contents: String) {
     var hasher = FNV()
     hasher.combine(contents)
-    self.init(name: .virtual(hasher.state), contents: contents)
+    self.init(
+      name: .virtual(URL(string: "virtual:///\(String(UInt(bitPattern: hasher.state), radix: 36))")!),
+      contents: contents)
   }
 
   /// The name of the file that the source came from.
@@ -75,12 +74,7 @@ public struct SourceFile: Hashable, Sendable {
 
   /// The name of the source file, sans path qualification or extension.
   public var baseName: String {
-    switch name {
-    case .local(let u):
-      return u.deletingPathExtension().lastPathComponent
-    case .virtual(let i):
-      return String(UInt(bitPattern: i), radix: 36)
-    }
+    name.url.deletingPathExtension().lastPathComponent
   }
 
   /// The contents of the file.
--- a/Sources/FrontEnd/Module.swift
+++ b/Sources/FrontEnd/Module.swift
@@ -347,6 +347,13 @@ public struct Module: Sendable {
     sources.values.map(\.topLevelDeclarations).joined()
   }
 
+  /// Returns the identity of a contained source file named `f`, if any.
+  public func sourceFile(named f: FileName) -> SourceFile.ID? {
+    if let i = sources.index(forKey: f) {
+      SourceFile.ID(module: identity, offset: i)
+    } else { nil }
+  }
+
   /// Projects the source file identified by `f`.
   internal subscript(f: SourceFile.ID) -> SourceContainer {
     _read {
--- a/Sources/FrontEnd/Module.swift
+++ b/Sources/FrontEnd/Module.swift
@@ -269,6 +269,8 @@ public struct Module: Sendable {
   }
 
   /// Adds a source file to this module.
+  /// 
+  /// - Requires: File name `s.name` is not already present in the program.
   @discardableResult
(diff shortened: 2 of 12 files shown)


Sources/FrontEnd/Files/SourceFile.swift
```diff
@@ -50,6 +50,11 @@ public struct SourceFile: Hashable, Sendable {
     try self.init(name: .local(path), contents: .init(contentsOf: path, encoding: .utf8))
   }
 
+  /// Creates an in-memory source file that represents a possibly non-existent file at `path`.
+  public init(representing path: URL, inMemoryContents: String) {
```
tothambrus11:
```suggestion
  public init(id: URL, contents: String) {
```

change docs

tothambrus11: delegate other to this

tothambrus11: mention this is for LSP-like

Sources/FrontEnd/Files/SourceFile.swift
```diff
@@ -148,6 +148,13 @@ public struct SourceFile: Hashable, Sendable {
     return (lineNumber, columnNumber)
   }
 
+  /// Returns the index in `text` corresponding to the 1-based `line` and `column`.
+  ///
+  /// - Requires: `line` and `column` describe a valid position in `self`.
```
tothambrus11: No need for requires

Sources/FrontEnd/Module.swift
```diff
@@ -30,7 +30,7 @@ public struct Module: Sendable {
     internal var roots: [DeclarationIdentity] = []
 
     /// The top-level declarations in `self`.
-    internal var topLevelDeclarations: [DeclarationIdentity] = []
+    public internal(set) var topLevelDeclarations: [DeclarationIdentity] = []
```
tothambrus11: these can be removed, access through module

Sources/FrontEnd/Module.swift
```diff
@@ -216,7 +216,7 @@ public struct Module: Sendable {
   public private(set) var dependencies: [Module.Name] = []
 
   /// The source files in the module.
-  internal private(set) var sources: OrderedDictionary<FileName, SourceContainer> = [:]
+  public private(set) var sources: OrderedDictionary<FileName, SourceContainer> = [:]
```
tothambrus11: access through module

Tests/FrontEndTests/Files/SourceFileTests.swift
```diff
@@ -75,6 +75,26 @@ final class SourceFileTests: XCTestCase {
     XCTAssertEqual(p2.description, "virtual://350c8wstjkie0:2:6")
   }
 
+  func testIndexLineColumn() throws {
+    let f = SourceFile.helloWorld
+    try XCTSkipIf(f.lineCount != 2)
+
+    // Line 1, column 1 -> start of file.
```
tothambrus11: break up tests, make the names descriptive

Sources/FrontEnd/Syntax/SyntaxVisitor.swift
```diff
@@ -169,6 +169,13 @@ extension Program {
     visit(ps.usings, calling: &v)
   }
 
+  /// Visits the children of `sourceFile` in pre-order, calling back `visitor` when a node is entered or left.
```
tothambrus11: no need, just call visit(topleveldeclarations)

tothambrus11: improve visitor docs

Sources/FrontEnd/Typer/Typer.swift
```diff
@@ -3701,7 +3701,7 @@ public struct Typer {
   /// type is being computed.
   ///
   /// The result does not include built-in givens.
-  private mutating func givens(visibleFrom scopeOfUse: ScopeIdentity) -> [[Given]] {
+  public mutating func givens(visibleFrom scopeOfUse: ScopeIdentity) -> [[Given]] {
```
tothambrus11: Make a method in program, requiring program's typing is done

use withTyper

use `internal`

Sources/FrontEnd/Files/SourceFile.swift
```diff
@@ -42,19 +42,39 @@ public struct SourceFile: Hashable, Sendable {
...
+  public init(localPath p: URL, contents: String) {
+    self.init(name: .local(p), contents: contents)
+  }
+
+  /// Creates a source file with given `contents` and virtual path `p`.
+  /// 
+  /// - Requires: `p` has a file name component.
+  public init(virtualPath p: URL, contents: String) {
+    self.init(name: .virtual(p), contents: contents)
+  }
```
tothambrus11: Alternatively, we could make `init(name: FileName, contents: String)` public and delete these.

Sources/FrontEnd/Files/SourceFile.swift
```diff
@@ -42,19 +42,39 @@ public struct SourceFile: Hashable, Sendable {
...
+
+  /// Creates a virtual source file with the given `contents`.
   public init(contents: String) {
     var hasher = FNV()
     hasher.combine(contents)
-    self.init(name: .virtual(hasher.state), contents: contents)
+    self.init(
+      virtualPath: URL(string: "virtual:///\(String(UInt(bitPattern: hasher.state), radix: 36))")!,
+      contents: contents)
```
tothambrus11, line 67: Note: using `virtual:///` instead of `virtual://` because the segment after `://` is not interpreted as a `pathComponent` but as the domain. `virtual:///` is the first url with non-empty `pathComponents`.

Sources/FrontEnd/Syntax/SyntaxVisitor.swift
```diff
@@ -155,7 +155,7 @@ extension Program {
   }
 
   /// Visits `ns` and their children in pre-order, calling back `v` when a node is entered or left.
-  public func visit<T: SyntaxVisitor, U: Collection>(
+  public func visit<T: SyntaxVisitor, U: Sequence>(
```
tothambrus11, line 158: Relaxing the requirement, because `topLevelDeclarations` returns `some Sequence<DeclarationIdentity>`

Sources/FrontEnd/Files/SourceFile.swift
```diff
@@ -41,7 +41,10 @@ public struct SourceFile: Hashable, Sendable {
   private let properties: Properties
 
   /// Creates a source file with the given name and contents.
+  ///
+  /// - Requires: `name` has a file name component.
```
lucteo: I thought this is not required for `virtual` `FileName`s. Why this requirement?

(based on the description, I understand that one might call this for virtual filenames)

Sources/FrontEnd/Files/SourceFile.swift
```diff
@@ -41,7 +41,10 @@ public struct SourceFile: Hashable, Sendable {
...
   private init(name: FileName, contents: String) {
+    precondition(name.url.pathComponents.count >= 2) // The 0th component is `/`, followed by the real components.
     self.properties = .init(name: name, text: contents)
   }
```
lucteo, line 53: is this initializer adding any value compared to the ones below?

Sources/FrontEnd/Files/SourceFile.swift
lucteo: why do we care? I thought we can put any URL in the `virtual` case.

Tests/FrontEndTests/Files/SourceFileTests.swift
```diff
@@ -62,17 +62,17 @@ final class SourceFileTests: XCTestCase {
...
     let f = SourceFile.helloWorld
     let i1 = try XCTUnwrap(f.text.firstIndex(of: ","))
     let p1 = SourcePosition(i1, in: f)
-    XCTAssertEqual(p1.description, "virtual://350c8wstjkie0:1:6")
+    XCTAssertEqual(p1.description, "virtual:///350c8wstjkie0:1:6")
     let i2 = try XCTUnwrap(f.text.firstIndex(of: "!"))
     let p2 = SourcePosition(i2, in: f)
-    XCTAssertEqual(p2.description, "virtual://350c8wstjkie0:2:6")
+    XCTAssertEqual(p2.description, "virtual:///350c8wstjkie0:2:6")
   }
```
lucteo, line 77: can you please add a test in which the URL doesn't resemple a fully qualified path?
(not sure what are the limits of `URL`, but I would try to put something at its limits)

Sources/FrontEnd/Program.swift
```diff
@@ -1645,3 +1647,41 @@ fileprivate struct ChildrenEnumerator: SyntaxVisitor {
...
+extension Program {
+
+  /// Returns the identity of source file named `f`, if any.
+  public func sourceFile(named f: FileName) -> SourceFile.ID? {
+    modules.values.indices.firstNonNil { (m) in
+      if let i = self[m].sources.index(forKey: f) {
+        SourceFile.ID(module: m, offset: i)
+      } else { nil }
+    }
+  }
```
lucteo, line 1659: can we move these functions to the `Module` level? They feel more appropriate there.

Tests/FrontEndTests/ProgramTests.swift
```diff
@@ -63,6 +64,95 @@ final class ProgramTests: XCTestCase {
     }
   }
 
+  func testGivens() async throws {
```
lucteo: can you please document what the tests does?

we can check that we can parse givens, that we can type check givens, the fact that we can call `Program.givens` and returns something/anything, or that `Program.givens` returns the expected value.

In our case, it seems that we test that we get a user given that is a `ConformanceDeclaration` (but we are not testing that it's the one that we have in our small source file).

Same with other tests.

Sources/FrontEnd/Files/SourceFile.swift
tothambrus11: So that `baseName` on `SourceFile` can be implemented correctly, which requires at least one path component.

Sources/FrontEnd/Files/SourceFile.swift
tothambrus11: Same as above; So that `baseName` of SourceFile can be implemented easily, by taking the last path component and stripping it from any suffix.

Tests/FrontEndTests/Files/SourceFileTests.swift
tothambrus11: I added this test, which revealed that two SourceFiles were not equal when constructed with a relative vs absolute URL. I changed the constructor of SourceFile to take the `.absoluteURL` of any provided URL. 

```swift

/// Tests that when a file is created with a relative URL, it is equivalent 
  /// to creating it with an absolute URL.
  func testRelativePath() {
    let nonCanonical = URL(fileURLWithPath: "a/../README.md")
    let relative = URL(fileURLWithPath: "README.md")
    let absolute = relative.absoluteURL
    
    let sn = SourceFile(localPath: nonCanonical, contents: "hello")
    let sr = SourceFile(localPath: relative, contents: "hello")
    let sa = SourceFile(localPath: absolute, contents: "hello")

    XCTAssertEqual(sr.baseName, "README")
    XCTAssertEqual(sa.baseName, "README")
    XCTAssertEqual(sn.baseName, "README")

    XCTAssertEqual(sr, sa)
    XCTAssertEqual(sn, sa)
    XCTAssertEqual(sr.hashValue, sa.hashValue)
    XCTAssertEqual(sn.hashValue, sa.hashValue)

    XCTAssertEqual(sr.name.url, absolute)
    XCTAssertEqual(sn.name.url, absolute)
  }
```

Sources/FrontEnd/Program.swift
tothambrus11: I moved a part of the `sourceFile(named:)` function to Module, and called it from the program to find the program-wide first source file.

I can't move `givens`, because it requires a Program for `withTyper`.

The functions taking a `SourceFile.ID` argument are convenient to keep in program. `SourceFile.ID` identifies both the module and the source file, so when moving this to Module, use sites would look like this: `program[id.module].diagnostics(in: id)`. `id` is passed in twice, so it's a bit redundant or more error-prone than doing the lookup all at once.

Sources/FrontEnd/Files/SourceFile.swift
tothambrus11: Not much, but archiving required this constructor. I removed the other 2 in favor of this, since they were essentially just forwarding to this anyway.

tothambrus11:
@kyouko-taiga This is ready for a re-review when you have time.

PR #14, opened 2025-12-07
Add LLVM infrastructure support

tothambrus11:


kyouko-taiga (commented):
We should probably revert this set of changes and keep them on the side until we have discussed what we want to do for native code generation.

PR #28, opened 2026-02-05
Enhance `CommandLine` and `Driver` to support regular phases of the compiler

lucteo:
Allows the user to specify what should `hc` emit and also the output file. We have placeholders for emitting LLVM, Intel ASM and binary.

### DIFF
--- a/Sources/Driver/Driver.swift
+++ b/Sources/Driver/Driver.swift
@@ -85,7 +85,7 @@ public struct Driver {
     return (elapsed, program[module].containsError)
   }
 
-  /// Generate backend code for `module`.
+  /// Generates backend code for `module`.
   public mutating func generateCode(
     _ module: Module.ID
   ) async -> (elapsed: Duration, containsError: Bool) {
@@ -96,7 +96,7 @@ public struct Driver {
     return (elapsed, program[module].containsError)
   }
 
-  /// Generate executable from `module`.
+  /// Generates executable from `module`.
   public mutating func generateExecutable(
     _ module: Module.ID
   ) async -> (elapsed: Duration, containsError: Bool) {


Sources/hc/CommandLine.swift
```diff
@@ -42,9 +42,17 @@ import Utilities
   @Option(
     name: [.customLong("emit")],
     help: ArgumentHelp(
-      "Produce the specified output: Possible values are: ast, typed-ast",
+      "Produce the specified output: Possible values are: ast, typed-ast, ir, raw-ir, llvm, intel-asm, binary.",
```
```suggestion
      "Produce the specified output: Possible values are: \(OutputType.allValueStrings.joined(separator: ", ")).",
```

Sources/Driver/Driver.swift
```diff
@@ -73,6 +73,39 @@ public struct Driver {
     return (elapsed, program[module].containsError)
   }
 
+  /// Applies the transformation passes on the IR of `module`.
```
```suggestion
  /// Applies mandatory transformation passes on the IR of `module`.
```

kyouko-taiga (commented):
Could there be an option to write the result of typing to the console directly? I use that quite often during debugging.

Sources/hc/CommandLine.swift
```diff
@@ -212,6 +280,50 @@ import Utilities
...
+  /// Given the desired name of the compiler's product, returns the file to write when "intel-asm"
+  /// is selected as the output type.
+  private func intelASMFile(_ productName: Module.Name) -> URL {
+    outputURL ?? URL(fileURLWithPath: productName.description + ".s")
+  }
+
+  /// Given the desired name of the compiler's product, returns the file to write when "binary" is
+  /// selected as the output type.
+  private func binaryFile(_ productName: Module.Name) -> URL {
+    outputURL ?? URL(fileURLWithPath: productName.description)
```
tothambrus11: These functions don't return a _file_ but a _file url_, so it would be nice to reflect that in the function name somehow, e.g. `astFileUrl(product:)`, `astFileTarget(for product:)`, `targetAstUrl(for product)`

tothambrus11 (changes requested):
Useful changes 👍 

Some suggestion for reducing code duplication for CLI argument names, and improving function names.

Sources/hc/CommandLine.swift
lucteo: I've copied these functions from the old compiler.
Plus, one good thing I've learned from Dave is that we should not reflect types in the names of functions / parameters.

Sources/hc/CommandLine.swift
```diff
@@ -42,9 +42,17 @@ import Utilities
   @Option(
     name: [.customLong("emit")],
     help: ArgumentHelp(
-      "Produce the specified output: Possible values are: ast, typed-ast",
+      "Produce the specified output: Possible values are: \(OutputType.allValueStrings.joined(separator: ", ")).",
```
```suggestion
      "Produce the specified output: \(OutputType.allValueStrings.joined(separator: ", ")).",
```
I think there should not be two colons in the message.

Sources/hc/CommandLine.swift
```diff
@@ -91,24 +99,45 @@ import Utilities
     let sources = try sourceFiles(recursivelyContainedIn: inputs)
     await perform("parsing", { await driver.parse(sources, into: module) })
     await perform("scoping", { await driver.assignScopes(of: module) })
+    if outputType == .ast {
+      try emitAst(module, in: driver.program, name: product)
+      return
+    }
+
     await perform("typing", { await driver.assignTypes(of: module) })
```
kyouko-taiga: I am surprised that there is no early exit after each stage if an error occurred. Is that logic handles elsewhere?

kyouko-taiga (changes requested):

Sources/Driver/Driver.swift
```diff
@@ -73,6 +73,38 @@ public struct Driver {
...
+    _ module: Module.ID
+  ) async -> (elapsed: Duration, containsError: Bool) {
+    let elapsed = ContinuousClock().measure {
+      program.applyTransformationPasses(module)
+    }
+    return (elapsed, program[module].containsError)
+  }
+
+  /// Generate LLVM code for `module`.
+  public mutating func generateCode(
```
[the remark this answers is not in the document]
lucteo: I was preparing for the follow-up patch (or set of patches) that introduces the LLVM codegen. My hope is that this can be filled in soon.

I also pondered whether the code should commit to LLVM backend. I choose the answer to be "yes" because:
- we did the same thing in the old compiler
- at this point we only plan to support LLVM
- when we add the second backend, we need some abstraction over the backends, and we will change all these places anyways; at that point, we will properly update the documentation.

Sources/hc/CommandLine.swift
```diff
@@ -119,6 +148,44 @@ import Utilities
...
+  private func emitAst(
+    _ module: Module.ID, in program: Program, name: Module.Name
+  ) throws {
+    let target = astFile(name)
+    let c = treePrinterConfiguration(for: treePrinterFlags)
+    let a = program.select(from: module, .satisfies({ program.parent(containing: $0).isFile }))
+    let parts = a.map({ (d) in
+      program.show(d, configuration: c)
+    }).lazy
+    try parts.joined(separator: "\n").write(to: target, atomically: true, encoding: .utf8)
```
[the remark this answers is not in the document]
lucteo: Thank you! I now understand now why this pattern is not ideal. I was assuming that Swift will know to do the right thing, but apparently my assumption was wrong.

Will change (probably going to make a function for this)

lucteo:
not sure why some of the builds are failing. Investingating...

Sources/Driver/Driver.swift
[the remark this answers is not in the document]
kyouko-taiga: SGTM

lucteo:
> Could there be an option to write the result of typing to the console directly? I use that quite often during debugging.

Sorry, @kyouko-taiga, haven't seen this one so far. I was also thinking about this, but then I dropped it. I will add such an option.

Sources/Driver/Driver.swift
```diff
@@ -74,6 +74,39 @@ public struct Driver {
...
+  public mutating func applyTransformationPasses(
+    _ module: Module.ID
+  ) async -> (elapsed: Duration, containsError: Bool) {
+    let elapsed = ContinuousClock().measure {
+      program.applyTransformationPasses(module)
+    }
+    return (elapsed, program[module].containsError)
+  }
+
+  /// Generate backend code for `module`.
```
```suggestion
  /// Generates backend code for `module`.
```

Sources/Driver/Driver.swift
```diff
@@ -74,6 +74,39 @@ public struct Driver {
...
+    _ module: Module.ID
+  ) async -> (elapsed: Duration, containsError: Bool) {
+    let clock = ContinuousClock()
+    let elapsed = clock.measure {
+      // TODO
+    }
+    return (elapsed, program[module].containsError)
+  }
+
+  /// Generate executable from `module`.
```
```suggestion
  /// Generates executable from `module`.
```

kyouko-taiga (approved):

PR #45, opened 2026-02-22
Enforce formatting in CI

tothambrus11:
Added a CI step to enforce formatting within the repository, and the scripts `format.sh` and `lint.sh`.


tothambrus11:
Alright, you can check out the reformatting changes. Some of them I also don't think are improvements, so we can look into if we can fix them with customization options, and we can discuss how to proceed in the next meeting.

tothambrus11:
How does Xcode work with regards to formatting? Doesn't it also use the toolchain's built-in swift-format?

Sources/FrontEnd/Parser/Parser.swift
```diff
@@ -177,7 +177,9 @@ public struct Parser {
     }
 
     // None of the above.
-    else { throw expected("annotation argument") }
+    else {
+      throw expected("annotation argument")
+    }
```
tothambrus11: IMO the original version was fine.

Sources/FrontEnd/Parser/Parser.swift
```diff
@@ -423,7 +424,8 @@ public struct Parser {
       _ = try take(contextual: "is") ?? expected("'is'")
       let concept = try parseExpression(in: &file)
       let witness = file.desugaredConformance(of: lhs, to: concept)
-      let members = self.next(is: .leftBrace)
+      let members =
+        self.next(is: .leftBrace)
         ? try parseTypeBody(in: &file, accepting: \.isValidStructMember)
         : nil
```
tothambrus11: IMO this is not an improvement

PR #53, opened 2026-03-16
LLVM lowering

tothambrus11:
## Contents of this PR
- Added basic LLVM lowering of Hylo IR.
- Added behavioral testing utilities, asserting the exit code, and optionally the contents of stdout.
- A minimal standard library that's implemented in C and only used as declarations.

## Limitations
- Linkage not yet present in the IR -> everything becomes public
- Global variables are not yet transpiled. https://github.com/hylo-lang/hylo-new/issues/86
- Enums are not yet transpiled. We should decide on a memory layout and add this in a later PR.
- Mangling is mocked by prepending `hylo_` to the `show`n name of entities.
- Many builtin instructions are missing, we can find a strategy later to add them.
- No struct layout algorithm is implemented yet, we rely on LLVM's default struct layout. https://github.com/hylo-lang/hylo-new/issues/85
- Projections are not lowered yet, whenever there is a usage of givens, the compilation fails. I'm using the minimal standard library with C interop to work around unsupported features whose lowering is not supported yet. Currently, the standard library is not yet lowered and linked, instead a C program is linked and all funtions in the minimal standard library are extern.
- I changed the transpilation to be done within `CodeGenerationContext` rather than an extension of SwiftyLLVM.Module. My reasoning was that it can be useful to know if something is a SwiftyLLVM call or a transpilation function, so now everything is prepended with the llvm module: `llvm.function(named: program.llvmName(of: f))`. I don't have very strong opinions on this though, so if you think this is too verbose, we can change it back.


Sources/FrontEnd/Typer/Typer.swift
```diff
@@ -335,7 +335,7 @@ public struct Typer {
 
   /// Type checks `d`, which occurs as a condition item iff `conditional` is true.
   ///
-  /// If `conditional` is `true`, then `d` occurs in the condition of a 
+  /// If `conditional` is `true`, then `d` occurs in the condition of a
```
tothambrus11: It seems I accidentally formatted this file, nothing interesting here. I'll revert these changes asap.

Package.swift
```diff
@@ -137,3 +162,28 @@ let package = Package(
...
+  // Heuristic for common fle formats:
+  if path.hasSuffix(".hylo") || path.hasSuffix(".observed") || path.hasSuffix(".expected") ||
+    path.hasSuffix(".diagnostics") || path.hasSuffix(".c") {
+    return false
+  }
+
+  // Fallback to filesystem check
+  var isDirectory: ObjCBool = true
+  precondition(FileManager.default.fileExists(atPath: path, isDirectory: &isDirectory))
+  return isDirectory.boolValue
```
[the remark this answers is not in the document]
tothambrus11: I'll use `fatalError` instead.

lucteo:
The PR seems too big for me. It touches too many parts of the compiler and does things that seems unrelated (new additions to standard library, updating documentation, fixes in existing code, updates to CI, updates to testing framework, etc.)

Can we break it down into smaller PRs?

Sources/hc/CommandLine.swift
```diff
@@ -299,6 +330,9 @@ import Utilities
     /// Assembly.
     case asm = "asm"
 
+    /// Object files.
+    case objectFiles = "object-files"
```
lucteo: Why use plural here? Also, do we need the "-files" suffix? Wouldn't just `obj`/`object` be more consisntent with what we have in this enum?

Sources/hc/CommandLine.swift
```diff
@@ -327,10 +361,19 @@ import Utilities
     outputURL ?? URL(fileURLWithPath: productName.description + ".s")
   }
 
+  /// Returns the directory to write when "object-files" is selected as the output type.
+  private func objectFilesDirectory() throws -> URL {
+    guard outputURL?.relativePath != "-" else {
+      throw ValidationError("object files cannot be written to the standard output")
+    }
+    return outputURL ?? URL(fileURLWithPath: "./objects")
```
lucteo: (extra minor) If we want to keep the extra directory we should probably make it configurable. But this can be done in a later PR, IMHO

Sources/Driver/Driver.swift
```diff
@@ -86,27 +169,95 @@ public struct Driver {
...
+  /// - Requires: `lowerToLLVM(_:)` has been called for `module`.
   public mutating func generateExecutable(
-    _ module: Module.ID
-  ) async -> (elapsed: Duration, containsError: Bool) {
-    let clock = ContinuousClock()
-    let elapsed = clock.measure {
-      // TODO
+    for module: Module.ID,
+    writingTo output: URL? = nil
+  ) async throws -> (elapsed: Duration, containsError: Bool) {
```
lucteo: Does this need to be `async`?

lucteo (commented):
Only a partial review

lucteo:
Can we separate the renaming of the tests (with just the code changes that will enable tests to work), and submit that first?

CONVENTIONS.md
```diff
@@ -0,0 +1,26 @@
...
+- Use strong types where possible to enforce invariants and prevent invalid states.
+
+## Contracts
+- Create the strictest contracts possible, so long the client can reason about the preconditions locally.
+- If the contract seems to strict, you may relax it and make invalid inputs a reported error (preferably `throw`n).
+
+## Naming
+- In a small scope, single-letter variable names can be used.
+- No abbreviations shall be used in APIs, unless something is a universally known acronym.
+
```
lucteo: Can this file be put in a different PR?

Package.swift
```diff
@@ -15,7 +16,7 @@ let commonSwiftSettings: [SwiftSetting] = [
 let package = Package(
   name: "Hylo",
   platforms: [
-    .macOS(.v13)
+    .macOS(.v15)
```
lucteo, line 19: I would also prefer to have changing the MacOS version to be done independently of the LLVM code gen.

Package.swift
```diff
@@ -137,3 +162,29 @@ let package = Package(
...
+  guard let enumerator: FileManager.DirectoryEnumerator = FileManager.default.enumerator(atPath: directory) 
+  else { return [] }
+  
+  let l = enumerator.compactMap { $0 as? String }
+    .filter { !$0.hasSuffix(".swift") && !isDirectory(directory + "/" + $0) }
+
+  return l
+}
+
+func isDirectory(_ path: String) -> Bool {
```
lucteo, line 183: please add documentation.

Sources/BackEnd/LLVM/TypeLowering.swift
```diff
@@ -0,0 +1,84 @@
+import FrontEnd
```
lucteo, line 1: Another suggestion for splitting this PR is:
- have changes to link to SwiftyLLVM, wiring in the Driver & compilation in one PR
    - here, we generate a simple `int main() { return 0; }` LLVM program all the time, regardless of the input program.
- have the actual logic of LLVM codegen in another PR (that should be our last PR)

StandardLibrary/Full/Shims/shim.c
```diff
@@ -0,0 +1,5 @@
+#include <stdio.h>
+#include <stdint.h>
+
+// The full standard library doesn't yet support compilation due to conformances requiring projection lowering
+// Once that's resolved, this file can be used to provide C implementations of some stdlib functions.
```
lucteo, line 5: can we add this file in a later PR?

StandardLibrary/Minimal/Sources/Core/Numbers/Int.hylo
```diff
@@ -0,0 +1,72 @@
+/// A signed integer value.
+@_symbol("Int")
+public struct Int is Deinitializable {
```
lucteo, line 3: I'm not a fan of maintaining two versions of standard library, and copy a bunch of definitions between them.
Do we need to have this separation now?

tothambrus11: This is a temporary hack while the lowering of the standard library is incomplete. We don't need to maintain or update the minimal version, as long as it sufficiently allows testing the lowering.

Package.swift
tothambrus11: The CI setup requires changing the macOS target version because the LLVM static libraries are built for MacOS 15, and the versions must match. I can put this into the PR that sets up LLVM with a fixed emitted program.

Sources/BackEnd/LLVM/CodeGenerationContext.swift
```diff
@@ -0,0 +1,1290 @@
+import FrontEnd
+import SwiftyLLVM
+import Utilities
+
+extension SwiftyLLVM.OverflowBehavior {
+
+  /// An instance equivalent to `x`
```
kyouko-taiga, line 7:
```suggestion
  /// Creates an instance equivalent to `x`.
```

Sources/BackEnd/LLVM/CodeGenerationContext.swift
```diff
@@ -0,0 +1,1290 @@
...
+/// The LLVM lowering happens upon construction, after which you can extract the resulting module via `extractModule()`.
+private struct CodeGenerationContext: ~Copyable {
+
+  /// The program containing the module being lowered.
+  private let program: Program
+
+  /// The LLVM module being built.
+  private var llvm: SwiftyLLVM.Module
+
+  /// The module's identity that's being transpiled.
```
kyouko-taiga:
```suggestion
  /// The identity of the module being transpiled.
```

Sources/BackEnd/LLVM/CodeGenerationContext.swift
```diff
@@ -0,0 +1,1290 @@
...
+  private let moduleID: FrontEnd.Module.ID
+
+  /// The functions from the current module to transpile.
+  private let functions: [FrontEnd.IRFunction]
```
kyouko-taiga: Aren't those contained in the Hylo module already?

Sources/BackEnd/LLVM/CodeGenerationContext.swift
```diff
@@ -0,0 +1,1290 @@
...
+
+  /// `isFunctionTranspiled[i]` is `true` iff the `i`th function's transpilation has begun.
+  private var isFunctionTranspiled: [Bool]
+
+  /// Initialized a new instance from its parts, without lowering anything.
```
kyouko-taiga:
```suggestion
  /// Creates an instance from the given properties, without lowering anything.
```

The rephrasing is only for consistency with the rest of the codebase.

Sources/BackEnd/LLVM/CodeGenerationContext.swift
```diff
@@ -0,0 +1,1290 @@
...
+  /// The LLVM module being built.
+  private var llvm: SwiftyLLVM.Module
+
+  /// The module's identity that's being transpiled.
+  private let moduleID: FrontEnd.Module.ID
+
+  /// The functions from the current module to transpile.
+  private let functions: [FrontEnd.IRFunction]
+
+  /// `isFunctionTranspiled[i]` is `true` iff the `i`th function's transpilation has begun.
```
kyouko-taiga: `i` being the position of the function in `self.functions`, right? The precision would be useful I think.

But unless this array is consulted very often in hot loops I would probably define a set of `IRFunction.Name` instead. I feel like this representation is brittle, as it has to be synchronized with another stored property.

Sources/BackEnd/LLVM/CodeGenerationContext.swift
```diff
@@ -0,0 +1,1290 @@
...
+private struct CodeGenerationContext: ~Copyable {
+
+  /// The program containing the module being lowered.
+  private let program: Program
+
+  /// The LLVM module being built.
+  private var llvm: SwiftyLLVM.Module
+
+  /// The module's identity that's being transpiled.
+  private let moduleID: FrontEnd.Module.ID
```
kyouko-taiga:
```suggestion
  private let module: FrontEnd.Module.ID
```

For consistency with other data structures, e.g., `IREmitter`.

Sources/BackEnd/LLVM/CodeGenerationContext.swift
```diff
@@ -0,0 +1,1290 @@
...
+  /// Initialized a new instance from its parts, without lowering anything.
+  private init(
+    transpiling module: FrontEnd.Module.ID, in program: Program,
+    compilingFor targetMachine: consuming SwiftyLLVM.TargetMachine
+  ) throws {
+    self.llvm = try SwiftyLLVM.Module(
+      program.modules.elements[module].key, targetMachine: targetMachine)
+    self.program = program
+    self.moduleID = module
+    // FIXME: avoid copying the whole array a second time. FrontEnd could expose an array instead of an opaque collection.
```
kyouko-taiga: Where is the first copy? Also I don't understand yet why you need the function to be copied. In other places of the compiler the hope is that we can parallelize some work but here I'm not sure we can. And even if we do, I don't think you need to modify the Hylo functions, do you?

Note that there is a method `move` in `IRFunction` to make a non-destructive move of the function, which will be more efficient than a copy if you end up mutating the functions later on.

Sources/BackEnd/LLVM/CodeGenerationContext.swift
```diff
@@ -0,0 +1,1290 @@
...
+  }
+
+  /// Incorporates all IR entities from an IR module to an LLVM module.
+  public static func transpiling(
+    _ module: FrontEnd.Module.ID, in program: Program,
+    compilingFor targetMachine: consuming SwiftyLLVM.TargetMachine
+  ) throws -> Self {
+    var context = try Self(transpiling: module, in: program, compilingFor: targetMachine)
+    for f in context.functions.indices {
+      try context.incorporate(functionWithIndex: f)
```
kyouko-taiga: I'm not a fan here that functions get identified by an index. There are already two kinds of identities in the compiler (`IRFunction.Name` and `IRFunction.ID`). These indices are adding a third, which I think makes the whole system more complicated for dubious benefits.

Sources/BackEnd/LLVM/CodeGenerationContext.swift
```diff
@@ -0,0 +1,1290 @@
...
+    }
+    return context
+  }
+
+  /// Extracts the LLVM module while consuming the code generation context.
+  public consuming func extractModule() -> SwiftyLLVM.Module {
```
kyouko-taiga:
```suggestion
  public consuming func release() -> SwiftyLLVM.Module {
```

For consistency with other data structures, such as `Typer`.

Sources/BackEnd/LLVM/CodeGenerationContext.swift
```diff
@@ -0,0 +1,1290 @@
...
+    consume llvm
+  }
+
+  /// The IR of the module being lowered.
+  var sourceModule: FrontEnd.Module {
```
kyouko-taiga:
```suggestion
  private var sourceModule: FrontEnd.Module {
```
I added explicit access modifier to most structures.

kyouko-taiga: Now I realize `CodeGenerationContext` is `private`, meaning that I don't think it's useful to define any `public` method. Am I missing something?

Sources/BackEnd/LLVM/CodeGenerationContext.swift
```diff
@@ -0,0 +1,1290 @@
...
+    get { program.modules.elements[moduleID].value }
```
kyouko-taiga:
```suggestion
    program.modules.elements[moduleID].value
```

Sources/BackEnd/LLVM/CodeGenerationContext.swift
kyouko-taiga:
```suggestion
  var source: FrontEnd.Module {
```
Does that work? The fact that it is a module is already encoded in the type of the property.

Sources/BackEnd/LLVM/CodeGenerationContext.swift
```diff
@@ -0,0 +1,1290 @@
...
+  }
+
+  /// Transpiles function with index `i` if not already transpiled.
+  mutating func incorporate(functionWithIndex i: Int) throws {
+    /// Don't re-transpile functions. TODO see if this is actually necessary.
```
kyouko-taiga: Why would your driver call `incorporate(functionWithIndex:)` with the same argument twice?

Sources/BackEnd/LLVM/CodeGenerationContext.swift
```diff
@@ -0,0 +1,1290 @@
...
+
+    let f = functions[i]
+
+    /// Don't transpile generic functions.
+    if f.isGeneric { return }
+
+    // Don't transpile projections.
+    guard let _ = f.returnRegister else {
+      var p = TreePrinter(program: program)
+      print("WARNING: function \(p.show(f.name)) has no return register, probably it's a subscript that we didn't lower yet.")
```
kyouko-taiga: The idea is that eventually you expect subscripts to have been transformed using Lucian's work? If so I think the above should be a fatal error. We should not (almost silently) compile a program with missing symbols. The failure will be pushed to the linker, which is not very intuitive.

Sources/BackEnd/LLVM/CodeGenerationContext.swift
```diff
@@ -0,0 +1,1290 @@
...
+    transpile(contentsOf: f, into: transpiledFunction)
+
+    if program.isModuleEntry(f) {
+      defineMain(calling: f)
+    }
+  }
+
+  /// Inserts or retrieves the transpiled declaration of an IR function `f`.
+  mutating func declareFunction(transpiledFrom f: IRFunction) -> SwiftyLLVM.Function.UnsafeReference
+  {
```
kyouko-taiga:
```suggestion
  mutating func declareFunction
    transpiledFrom f: IRFunction
  ) -> SwiftyLLVM.Function.UnsafeReference {
```

Sources/BackEnd/LLVM/CodeGenerationContext.swift
```diff
@@ -0,0 +1,1290 @@
...
+
+    let transpilation = llvm.function(named: program.llvmName(of: f))!
+    llvm.setLinkage(.private, for: transpilation)
+
+    let int32 = program.standardLibraryType(.int32)
+
+    if let r = f.returnRegister,
+      let rt = f.result(of: r),
+      rt.type == int32
+    {
```
kyouko-taiga:
```suggestion
    if let r = f.returnRegister, let rt = f.result(of: r), rt.type == int32 {
```

Sources/BackEnd/LLVM/CodeGenerationContext.swift
```diff
@@ -0,0 +1,1290 @@
...
+      let (type, _) = f.result(
+        of: f.returnRegister
+          ?? fatalError("function expected to be a regular function, not a subscript"))
+    else {
+      var p = TreePrinter(program: program)
+      fatalError("Why did we not get a result type of \(f.show(using: &p))?")
+    }
+
+    _ = type
+    // FIXME cannot call mutating never() method
```
kyouko-taiga: I don't understand  this FIXME.

Sources/BackEnd/LLVM/CodeGenerationContext.swift
```diff
@@ -0,0 +1,1290 @@
...
+    configureFunctionAttributes(function: transpiledFunction, transpiledFrom: f, in: sourceModule)
+    configureParameterAttributes(
+      parameters: transpiledFunction.unsafe[].parameters, transpiledFrom: f,
+      in: sourceModule)
+
+    return transpiledFunction
+  }
+
+  /// Defines a "main" function calling the function `f`, which represents the entry point of the
+  /// entry module `m` of the program `ir`.
```
kyouko-taiga: Where are `m` and `ir` defined?

Sources/BackEnd/LLVM/CodeGenerationContext.swift
```diff
@@ -0,0 +1,1290 @@
...
+      let s = llvm.insertAlloca(t, at: p)
+      _ = llvm.insertCall(transpilation, on: (s), at: p)
+      llvm.insertReturn(llvm.i32.unsafe[].zero, at: p)
+    }
+  }
+
+  /// Adds the function attributes to `llvmFunction` implied by its IR form `f`.
+  private mutating func configureFunctionAttributes(
+    function llvmFunction: SwiftyLLVM.Function.UnsafeReference, transpiledFrom f: IRFunction,
+    in m: FrontEnd.Module
```
kyouko-taiga: Do we really need `m`? Also why do we need `f` as a value rather than an index into `self.functions` or better an ID into the Hylo module being compiled?

In `Emitter`, we avoid having to pass these parameters over and over again by making them properties of `self`, which greatly simplifies APIs.

Sources/BackEnd/LLVM/CodeGenerationContext.swift
```diff
@@ -0,0 +1,1290 @@
...
+    //   setLinkage(.private, for: llvmFunction)
+    // }
+
+    guard
+      let (type, _) = f.result(
+        of: f.returnRegister
+          ?? fatalError("function expected to be a regular function, not a subscript"))
+    else {
+      var p = TreePrinter(program: program)
+      fatalError("Why did we not get a result type of \(f.show(using: &p))?")
```
kyouko-taiga: Good question .. 

The whole guard is a bit difficult to read, which makes me think it should probably be its own function. What are you trying to achieve?

Sources/BackEnd/LLVM/CodeGenerationContext.swift
```diff
@@ -0,0 +1,1290 @@
...
+    // FIXME cannot call mutating never() method
+    // if type != program.types.never() {
+    //   addFunctionAttribute(functionAttribute(.noreturn), to: llvmFunction)
+    // }
+  }
+
+  /// Adds the attributes to each parameter in `llvmParameters` implied by their corresponding IR
+  /// form in `m[f].termParameters`.
+  private mutating func configureParameterAttributes(
+    parameters llvmParameters: SwiftyLLVM.Function.Parameters,
```
kyouko-taiga:
```suggestion
    parameters: SwiftyLLVM.Function.Parameters,
```

Sources/BackEnd/LLVM/CodeGenerationContext.swift
```diff
@@ -0,0 +1,1290 @@
...
+      }
+    }
+
+    llvm.insertBr(to: block[entry]!, at: llvm.endOf(prologue))
+
+    /// Inserts the transpilation of `i` at `insertionPoint`.
+    func insert(_ i: AnyInstructionIdentity) {
+      switch f.tag(of: i) {
+      case IRAccess.self:
+        insert(access: i)
```
kyouko-taiga, line 330: The reason why OFE used local functions was that I wanted to avoid passing too many parameters to each `insert` function. But the debugging experience is not ideal and I also suspect weird shenanigans happening with reference semantics when `self` is being captured mutably.

If we make the current function and module being compiled part of the context (by making them properties of `self`) then I think all these methods can become members, just like in `IREmitter`.

Sources/BackEnd/LLVM/CodeGenerationContext.swift
```diff
@@ -0,0 +1,1290 @@
...
+
+    /// Inserts the transpilation of `i` at `insertionPoint`.
+    func insert(subfield i: AnyInstructionIdentity) {
+      let s = f.at(i) as! IRSubfield
+
+      let base = llvmOperand(s.base)
+      let baseType = program.llvmType(from: f.result(of: s.base)!.type, in: &llvm)
+      let indices =
+        [llvm.i32.unsafe[].constant(0).erased]
+        + s.path.map { sf in llvm.i32.unsafe[].constant(UInt64(sf)).erased }
```
kyouko-taiga:
```suggestion
        + s.path.map { (sf) in llvm.i32.unsafe[].constant(UInt64(sf)).erased }
```
For consistency.

Sources/BackEnd/LLVM/TypeLowering.swift
```diff
@@ -0,0 +1,84 @@
...
+    -> SwiftyLLVM.StructType.UnsafeReference {
+    module.structType(
+      named: llvmName(of: structType.erased),
+      storedPropertyTypes(of: structType).map { llvmType(from: $0, in: &module) },
+      packed: false  // TODO: use our own layout algorithm, manually emitting padding bits.
+    )
+  }
+
+  /// The types of each stored property of a struct in declaration order.
+  func storedPropertyTypes(of structType: Struct.ID) -> [AnyTypeIdentity] {
```
kyouko-taiga: There's a method in `Program` or `Typer` to do that already.

Sources/Driver/Driver.swift
```diff
@@ -185,4 +314,151 @@ public struct Driver {
...
+    program.modules.elements[module].key
+  }
+
+  private static func decode(_ buffer: borrowing SwiftyLLVM.MemoryBuffer) -> String {
+    buffer.withUnsafeBytes({ (bytes) in
+      String(decoding: bytes.map(UInt8.init(bitPattern:)), as: UTF8.self)
+    })
+  }
+
+  private final class LLVMModuleBox {
```
kyouko-taiga, line 333: Why do we need such a box?

Sources/Driver/Driver.swift
```diff
@@ -185,4 +314,151 @@ public struct Driver {
...
+      self.module = module
+    }
+
+  }
+
+  private struct ExecutionResult {
+    let standardOutput: String
+    let standardError: String
+    let terminationStatus: Int32
+  }
```
```suggestion
  private struct ExecutionResult {

    let standardOutput: String

    let standardError: String

    let terminationStatus: Int32

  }
```
Please add docs to these declarations.

Sources/FrontEnd/IR/IRFunction.swift
```diff
@@ -151,6 +151,11 @@ public struct IRFunction: Sendable {
     blocks.firstAddress
   }
 
+  /// `true` iff `self` takes some generic parameters.
```
kyouko-taiga:
```suggestion
  /// `true` iff `self` accepts generic type parameters. 
```

Sources/Driver/Driver.swift
```diff
@@ -185,4 +314,151 @@ public struct Driver {
...
+
+  private enum DriverFailure: LocalizedError, CustomStringConvertible {
+
+    case commandFailed(String, [String], Int32, String)
+    case missingBackendArtifact(String)
+    case missingExecutable(String)
```
```suggestion
    case commandFailed(String, [String], Int32, String)

    case missingBackendArtifact(String)

    case missingExecutable(String)
```

Sources/Driver/Driver.swift
```diff
@@ -185,4 +314,151 @@ public struct Driver {
...
+      }
+    }
+
+    var description: String { "\n\(errorDescription)" }
+
+  }
+
+}
+
+extension FileManager {
```
kyouko-taiga: There's a file named `FileManager+Extensions` where I think these declarations should go.

Sources/hc/CommandLine.swift
```diff
@@ -80,77 +101,98 @@ import Utilities
...
-    }
 
-    // Create a module for the product being compiled.
-    let product = productName(inputs)
-    note("start compiling \(product)")
-    let module = driver.program.demandModule(product)
-    if !noStandardLibrary {
-      driver.program[module].addDependency(Module.standardLibraryName)
-    }
+      // Create a module for the product being compiled.
```
kyouko-taiga, line 155: All of this code has to be under the `do` block because one method can throw? I would rather catch the exceptions thrown by that method more locally.

Sources/hc/CommandLine.swift
```diff
@@ -80,77 +101,98 @@ import Utilities
...
+        try emitLLVM(module, from: driver, name: product)
+        return
+      }
+      if outputType == .asm {
+        try write(driver.assembly(of: module), to: asmFile(product))
+        return
+      }
+      if outputType == .object {
+        let modules = Array(driver.program.moduleIdentities)
+        for dependency in modules where dependency != module {
```
kyouko-taiga: My plan was that dependencies would be compiled (to IR) when they are loaded into the module. This code should not be here.

Sources/hc/CommandLine.swift
```diff
@@ -80,77 +101,98 @@ import Utilities
...
+          await perform("lowering", for: dependency, { await driver.lower(dependency) })
+          await perform(
+            "normalization",
+            for: dependency,
+            { await driver.applyTransformationPasses(dependency) })
```
```suggestion
          await perform("normalization", for: dependency) {
            await driver.applyTransformationPasses(dependency)
          }
```

kyouko-taiga: I don't like the fact that `dependency` is passed twice. If we need `perform` to accept a module, then it should project that module in the lambda running its body.

kyouko-taiga (changes requested):
Very nice work!

This PR is quite long so I only made a first pass mostly on the code generator and the driver.

There are some changes that I think should not be made yet. In particular I think the way dependencies are handled should not change so dramatically. The overall design is meant to compile dependencies in `load` rather than at the end of the pipeline, like it is done here. Once archiving will be handled completely there should be no need to type and lower dependencies, for example

I am also not in a hurry to merge changes to the structure of the standard library. This PR makes many changes to just get to a binary, which I guess is fine for a PoC, but I would rather avoid adding things like `Simplint` and make it possible to compile our actual `Int` type instead. I understand getting there requires some work on the main branch to lower the standard library completely but then I think we should keep this branch parallel to main and rebase it until we can merge a backend that doesn't add too many hacks to the rest of the codebase.

Sources/BackEnd/LLVM/CodeGenerationContext.swift
tothambrus11: They are exposed as opaque collections that can't be indexed easily with the same index as the corresponding  `isFunctionTranspiled` list, so I copied all entries from the `some Collection<IRFunction>` returned by Module's `functions`. We could simplify things by returning the concrete array from `functions` rather than `some Collection<IRFunction>`, but I would still save the result here so we avoid recomputing the array on every access.

Alternatively we could set up `isFunctionTranspiled` as a hashmap from `Collection.Index` to `Bool`, but that would likely be more expensive.

Sources/BackEnd/LLVM/CodeGenerationContext.swift
tothambrus11: I didn't find any way to access the ID of the `Never` type. `program.types.never()` requires mutable access to types, which at this point is already finalized. I don't think we should create another typer here just for this, so I guess we could save the ID of never in a convenient place. Do you have any recommendations where it would be suitable to save?

Sources/BackEnd/LLVM/CodeGenerationContext.swift
tothambrus11: We won't mutate anything, the second copy (or rather, map) is just required because Module exposes the functions via an opaque collection rather than an array. We need array for fast random-access lookup, and for checking whether something has been transpiled already.

Sources/BackEnd/LLVM/CodeGenerationContext.swift
tothambrus11: Yes, that's right.

Sources/Driver/Driver.swift
tothambrus11: Because LLVMModule is non-copyable, and we want to maintain a collection of them. We may be able to remove this collection if we don't perform the object file generation at a different stage than LLVM lowering, but for now these are decoupled.

CONVENTIONS.md
tothambrus11: https://github.com/hylo-lang/hylo-new/pull/102

Sources/BackEnd/LLVM/CodeGenerationContext.swift
tothambrus11: No longer saving these because we could eliminate `isFunctionTranspiled`

Sources/BackEnd/LLVM/CodeGenerationContext.swift
tothambrus11: Removed `isFunctionTranspiled`.

Sources/BackEnd/LLVM/CodeGenerationContext.swift
tothambrus11: I changed it to this temporary allowlist instead:

```swift
      // FIXME: we need to gracefully ignore some comformance declarations until we implement the transformation to continuations.


      /// Trivial deinitializable is never called.
      if name == "SimpleInt.$<ConformanceDeclaration at builtin-function-calls:4.28>" ||
         name == "Triple.$<ConformanceDeclaration at print-struct-elements:3.18>" {
        return
      }
      
      unreachable("function \(name) has no return register, probably it's a subscript that we didn't lower yet. See: \(p.show(f))")
    
```

These deinitializable conformances never seem to be called, so I guess DCE would get rid of them.

Sources/BackEnd/LLVM/CodeGenerationContext.swift
tothambrus11: @kyouko-taiga Do you have any suggestions here?

kyouko-taiga: I would store it in the `program` and ensure that is has been created after we've loaded standard library.

Sources/BackEnd/LLVM/CodeGenerationContext.swift
tothambrus11: I think it's unreachable, the return register's type is never `IRType.nothing`, right? (I lowered functions returning `Void`, that worked.)

I refactored this into a separate function:
```swift
  /// Returns the result type of a Hylo function (not a subscript).
  private func functionResultType(of f: IRFunction) -> AnyTypeIdentity {
    let (type, _) = f.result(
      of: f.returnRegister
        ?? unreachable("function expected to be a regular function, not a subscript"))
      ?? unreachable("The function's return register must produce a value: \(program.show(f))")
    
    return type
  }
```

Sources/BackEnd/LLVM/TypeLowering.swift
tothambrus11: It's in Typer. I'm not sure if it's worth re-wrapping program in Typer just so we can invoke this method.

```swift
  /// Returns the types of stored parts of `t`.
  private mutating func storage(of t: Struct.ID) -> [AnyTypeIdentity] {
    program.storedProperties(of: program.types[t].declaration).map { (v) in
      typeOfName(referringTo: .init(v), statically: false)
    }
  }
```

It also seems to be doing more things with `typeOfBoundMember` because of `statically: false`, so I expect it to do some additional checks before it realizes that it's actually a struct.

Sources/BackEnd/LLVM/CodeGenerationContext.swift
tothambrus11: The never type is not related to the standard library, the type store can already resolve it. This seems to work:

```swift
struct Program {
  [...]

  /// The `Never` type of Hylo.
  public let never: AnyTypeIdentity

  /// Creates an empty program.
  public init() {
    self.never = types.never().erased
  }
```

tothambrus11:
Superseded by #176

PR #75, opened 2026-04-03
Build hc distributable

tothambrus11:
This PR adds a build step that packages the release build of the `hc` product along with statically linked Swift standard library.

I changed the Hylo standard library loading to reference the bundled standard library by default. In the old compiler we had [an extra option to use the local source path of the standard library](https://github.com/hylo-lang/hylo/blob/main/StandardLibrary/StandardLibrary.swift#L9) during development when the 
`HYLO_USE_BUILTIN_STDLIB` environment variable is set to true. I think this might be unnecessary if, when developing the standard library itself, we could invoke the compiler with `--no-std` and pass in the standard library module's directory as an input. But if needed, I can add the old behavior.

(Also pinning the Windows SDK version, based on https://github.com/hylo-lang/hylo/pull/1861)

### DIFF
--- a/Tests/CompilerTests/StandardLibraryLoadingTests.swift
+++ b/Tests/CompilerTests/StandardLibraryLoadingTests.swift
@@ -1,0 +1,21 @@
+import FrontEnd
+import XCTest
+import Driver
+import StandardLibrary
+
+final class StandardLibraryLoadingTests: XCTestCase {
+  func testStandardLibraryLoading() async throws {
+    var driver = Driver()
+    try await driver.loadStandardLibrary()
+  }
+
+  func testStandardLibraryLoadingBundled() async throws {
+    var driver = Driver()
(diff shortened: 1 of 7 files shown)


Tests/CompilerTests/StandardLibraryLoadingTests.swift
```diff
@@ -0,0 +1,21 @@
...
+  func testStandardLibraryLoadingBundled() async throws {
+    var driver = Driver()
+    try await driver.load(Module.standardLibraryName, withSourcesAt: bundledStandardLibrarySources)
+  }
+
+  func testStandardLibraryLoadingLocal() async throws {
+    var driver = Driver()
+    try await driver.load(Module.standardLibraryName, withSourcesAt: localStandardLibrarySources)
+  }
+}
```
kyouko-taiga:
```suggestion

}
```

kyouko-taiga (approved):
Nice! I made some suggestions to improve the docs but otherwise LGTM.

PR #76, opened 2026-04-04
Add missing integer types to standard library, create generator script

tothambrus11:
This PR adds the missing integer types to the standard library, along with a generator that runs as part of the SPM build process and generates a read-only file inside the build folder. This generated file is always used through the bundle, as it's not supposed to be modified by hand. A later PR can add `//#source:` directive emission to the generator script so that it's easier to locate where something comes from. #301 

This change had significant impact on the test suite's performance, as type checking depends on how many conformances are in scope. For this reason, I only enabled generating conformances for `Int`. Everything else still gets an implementation and `Movable` + `ExpressibleByIntegerLiteral` conformances.


kyouko-taiga (changes requested):
What is the intended workflow here? Do we have to manually run the generator? If so when?

I think there should be a utility that we can run from SPM, the same way we can use `hc-tests` for manual actions related to test case generation.

StandardLibrary/Sources/Core/Numbers/Int.hylo
```diff
@@ -1,4 +1,6 @@
-/// A signed integer value.
+// Generated by ../../../generator.swift.
```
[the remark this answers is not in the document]
tothambrus11: The intent is to prevent people from _accidentally_ modify this file. Not fool-proof indeed.

Sources/FrontEnd/Typer/Typer.swift
```diff
@@ -4570,14 +4591,9 @@ public struct Typer {
...
-    case standardLibraryType(.int),
-        standardLibraryType(.int32),
-        standardLibraryType(.int64):
-      return true
-    default:
-      return false
-    }
+
+    let d = program.types.dealiased(t)
+    return Program.StandardLibraryEntity.allIntegerTypes.contains { standardLibraryType($0) == d }
```
tothambrus11: This has a cost of about 10 hasmap lookups. We could add a cached version of the list of standard library integer type _declarations_ in Program. Do you think that would be worth the extra top-level stored property in Program?

StandardLibrary/Sources/Core/Numbers/Int.hylo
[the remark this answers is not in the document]
tothambrus11: I added a README.md in the Generated/ folder instead of this comment.

Sources/FrontEnd/Typer/Typer.swift
```diff
@@ -4592,9 +4592,12 @@ public struct Typer {
...
     switch program.types.dealiased(t) {
-    case standardLibraryType(.int),
-        standardLibraryType(.int32),
-        standardLibraryType(.int64):
+    case //
+      standardLibraryType(.int), standardLibraryType(.uint),
+      standardLibraryType(.int8), standardLibraryType(.uint8),
+      standardLibraryType(.int16), standardLibraryType(.uint16),
+      standardLibraryType(.int32), standardLibraryType(.uint32),
+      standardLibraryType(.int64), standardLibraryType(.uint64):
```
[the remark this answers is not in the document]
tothambrus11: See the new way in one of the comments below.

tothambrus11:
Can we merge this and resolve the performance issues in later PR? If this lands, I would like to add floating point types too.

To lessen the performance regression, we can generate a subset of to-be-supported integer types, e.g. whatever we had before

kyouko-taiga:
Dave expressed concerns about the workflow. I didn't keep track of the discussion but we have to clear that up before we can merge this patch.

tothambrus11:
@kyouko-taiga @dabrahams The workflow is updated to generate `Generated.hylo` automatically using an SPM plugin that calls a CLI executable target. This change slowed down the type-checking of the standard library astronomically, so I'll try to profile this on different workloads and see what's the performance curve as we add more types.

StandardLibrary/Sources/Core/Numbers/Int.hylo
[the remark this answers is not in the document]
dabrahams: Generators should always make their generated files read-only.  Are you doing that?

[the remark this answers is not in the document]
dabrahams: Oh, no, it seems you can't do that because SPM wants to copy the file into a generated bundle, and if the bundle already exists, the copy onto a write-only file will fail.  **this is terrible**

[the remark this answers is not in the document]
dabrahams: This is part of why we need the generator to insert `#sourceLocation` directives.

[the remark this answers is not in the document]
dabrahams: …which we don't support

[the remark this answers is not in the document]
tothambrus11: I just implemented emitting read-only files, it seems to work. The build system probably copies the file with permissions reset. During regeneration, I make the file writable, then write it, and make it read-only again.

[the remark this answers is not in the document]
tothambrus11: The `#sourceLocation` directives will be the next step after this.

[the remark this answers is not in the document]
dabrahams: I discarded it because the directive isn't implemented but I wrote a `.withSourceLocation(file: StaticString = #filePath, line: UInt = #line)` method for `String` that will add a directive at the top with the correct line number.  Just subtract the number of newlines.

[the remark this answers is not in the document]
dabrahams: > I just implemented emitting read-only files, it seems to work.

Build, make a change to the generator script, and rebuild.  For me that would fail with a permissions error.

kyouko-taiga:
> This change had significant impact on the test suite's performance, as type checking conformances is currently O(n^2.3). For this reason, I only enabled the generation of integer types we actually need at the moment, and inserted dummy structs for the rest.

Can we get profiling data about the performance?

I'm not necessarily surprised about the quadratic complexity but I'd like to know what `n` is, and my intuition is that it is not the number of types. There's nothing expensive about type checking a type declaration like `struct S {}` and we could probably go over a million of them in very little time. What's expensive is to resolve conformances, which is work that is driven by conformance declarations.

Standard library is a kind of worst case scenario because we define many types with many conformances. So if my intuition is correct, standard library will be significantly more expensive to type check than regular code. But since we archive modules after typing, maybe that is not so bad.

Or maybe I'm completely wrong. In any case, we need more precise data to know.

Significantly increasing the time it takes to run the test is concerning a valid ground for blocking this PR until we can figure things out. If there is really nothing we can do about compilation time, then we should consider improving the support for the "minimal" standard library so that we can run the vast majority of tests with just the 10-ish symbols that are necessary to go through lowering.

dabrahams:
The standard library being slow to compile makes standard library development (and testing of Hylo in general) slow.  It's a bad thing but I'm not yet sure it warrants removing any integer types before we  try to optimize the typer.

tothambrus11:
I'll keep this PR open until we optimize the typer.

dabrahams:
Just a note that since my work is based on this branch, please do not rewrite its public history (e.g. by rebasing).  Thanks

kyouko-taiga:
We'll have to merge this sooner than later but I wish we would have approached this issue with a smaller hammer. I don't see the point merging 10 new types, with a very large APIs, while I'm struggling to make the few things we have working. That goes against iterative design.

We could have implemented just `UInt32` and `UInt16`, the 4 operations that we need on these types, and then we wouldn't be wondering whether it's a good bet to slow down every single run of the test suite to unblock much needed but mostly unrelated work.

Sources/hc-generate-stdlib/GenerateIntegerTypes.swift
```diff
@@ -0,0 +1,247 @@
...
+      }
+
+      /// Multiplies `self` by `other`.
+      public fun infix*= (other: Self) inout {
+        &self = self * other
+      }
+
+    }
+
+
```
kyouko-taiga, line 387: Too many newlines?
```suggestion
```

Plugins/GenerateHyloStdLibPlugin/GenerateHyloStdLibPlugin.swift
```diff
@@ -0,0 +1,23 @@
+import Foundation
+import PackagePlugin
+
+/// The SPM plugin generating repetitive parts of the standard library using `hc-generate-stdlib`.
+@main
+struct GenerateHyloStdLibPlugin: BuildToolPlugin {
```
kyouko-taiga:
```suggestion
struct GenerateStandardLibraryPlugin: BuildToolPlugin {
```

We're probably generating standard library for only one language, and we're generally avoiding abbreviations.

Sources/Driver/Driver.swift
```diff
@@ -247,9 +247,11 @@ public struct Driver {
   /// fingerprint matches the fingerprint of the source files in `root`. Otherwise, the module is
   /// compiled from sources and an archive is stored at `moduleCachePath`. If `moduleCachePath` is
   /// not set, the module is unconditionally compiled from sources and no archive is stored.
-  public mutating func load(_ module: Module.Name, withSourcesAt root: URL) async throws {
+  public mutating func load(
+    _ module: Module.Name, withSourcesAt root: URL, additionalSources: [SourceFile] = []
```
kyouko-taiga, line 284:
```suggestion
    _ module: Module.Name, withSourcesAt root: URL, generatedSources: [SourceFile] = []
```
Those additional sources are always generated, right?

Sources/FrontEnd/Program.swift
```diff
@@ -1796,8 +1828,9 @@ extension Program {
         let b = select(from: a, .symbol(n.rawValue)).uniqueElement,
         let d = castToDeclaration(b)
       else {
-         precondition(allowPartialStandardLibrary, "missing or corrupt standard library")
-         continue
+        precondition(allowPartialStandardLibrary,
+          "missing or corrupt standard library; missing \(n.rawValue)")
```
kyouko-taiga:
```suggestion
        precondition(
          allowPartialStandardLibrary,
          "missing or corrupt standard library; missing \(n.rawValue)")
```

Sources/hc-generate-stdlib/GenerateIntegerTypes.swift
```diff
@@ -0,0 +1,247 @@
...
+      documentation:
+        "/// A signed integer type that can represent any object size on the target platform.",
+      builtinType: "word", .signed, .fittingLargestObjectSize))
+
+  // types.append(
+  //   .init(
+  //     name: "UInt",
+  //     documentation:
+  //      "/// An unsigned integer type that can represent any object size on the target platform.",
+  //     builtinType: "word", .unsigned, .fittingLargestObjectSize))
```
kyouko-taiga: What's this?

Sources/hc-generate-stdlib/GenerateIntegerTypes.swift
```diff
@@ -0,0 +1,247 @@
...
+  return types.map(\.description).joined(separator: "\n")
+}
+
+extension IntegerTypeDefinition {
+
+  /// Creates a fixed-size integer type's definition.
+  static func fixedSize(bits: Int, _ signedness: SignednessPrefix) -> IntegerTypeDefinition {
+    .init(
+      name: signedness == .signed ? "Int\(bits)" : "UInt\(bits)",
+      documentation: "/// \(signedness.prefixedDescription) \(bits)-bit integer.",
```
kyouko-taiga:
```suggestion
      documentation: "/// A \(bits)-bit \(signedness.prefixedDescription) integer.",
```

Sources/hc-generate-stdlib/GenerateIntegerTypes.swift
```diff
@@ -0,0 +1,247 @@
+import Foundation
+
+/// Returns the Hylo source code declaring the standard library's integer types.
+func generateIntegerTypes() -> String {
```
kyouko-taiga:
```suggestion
private func generateIntegerTypes() -> String {
```
I guess. But maybe `internal`.

Sources/hc-generate-stdlib/GenerateIntegerTypes.swift
```diff
@@ -0,0 +1,247 @@
...
+    .init(
+      name: signedness == .signed ? "Int\(bits)" : "UInt\(bits)",
+      documentation: "/// \(signedness.prefixedDescription) \(bits)-bit integer.",
+      builtinType: "i\(bits)", signedness, .fixedWidth(bits: bits))
+  }
+
+}
+
+/// Whether an integer type is signed or unsigned.
+enum SignednessPrefix: String {
```
kyouko-taiga:
```suggestion
private enum SignednessPrefix: String {
```
Same guess as before.

Sources/hc-generate-stdlib/GenerateIntegerTypes.swift
```diff
@@ -0,0 +1,247 @@
...
+}
+
+/// Information for emitting a standard library integer type.
+struct IntegerTypeDefinition: CustomStringConvertible {
+
+  let declarationName: String
+  let declarationDocumentation: String
+  let builtinType: String
+  let signedness: SignednessPrefix
+  let size: Size
```
kyouko-taiga, line 108: I don't find these properties obvious enough to not deserve any documentation.

StandardLibrary/Sources/Core/Numbers/MockIntegerType.hylo
```diff
@@ -0,0 +1,15 @@
+@_symbol("Int8")
```
kyouko-taiga: How does this file work?

StandardLibrary/StandardLibrary.swift
```diff
@@ -15,3 +15,7 @@ public let localStandardLibrarySources = URL(fileURLWithPath: #filePath)
 /// its sources are compiled with the flag `USE_BUNDLED_STANDARD_LIBRARY` set.
 public let bundledStandardLibrarySources = Bundle.module.url(
   forResource: "Sources", withExtension: nil)!
+
+/// The bundled path of the source file containing the generated parts of the standard library.
+public let generatedStandardLibrarySource = Bundle.module.url(
+  forResource: "Generated.hylo", withExtension: nil)!
```
kyouko-taiga, line 21: Are all the generated types in a single source file? If so, why?

kyouko-taiga (changes requested):
Thanks for this work. The most important changes we have to make are:
- [x] declare all methods of in the types themselves and then have empty given declarations. 
- [x] mark all these methods `@inline(always)`.
- [x] call `Builtin.trap()` directly instead of `fatal_error()`.

These changes are necessary for my inliner.

I added a few comments with questions and further suggestions.

One thing in particular bothers me a little. From what I understood, we are generating one giant source file with all definitions. I'm not a fan if that's the case and would rather generate one file per type.

Regarding performance, I don't think it is necessary to make all types conform to all traits right aways. We could have the structs and their members (see above) and introduce givens later. Better yet we could keep a branch with these givens and use it to monitor performance improvements on `Typer`.

tothambrus11:
Yeah, we can downscale the hammer. We can change the generator to emit a version without conformances but with all methods implemented, so we can use the type as a concrete type. We can get away without `Copyable`/`Movable` conformances by adding `.zero()` to the type we want to copy. We can keep the conformances emitted for `Int`.

StandardLibrary/StandardLibrary.swift
tothambrus11: Having them in separate files would be indeed nicer to see. Unfortunately, that requires some complexity in our SPM build plugin.

It's because resources generated in a build plugin don't preserve the folder layout, so in the resources they end up as a flat set of files, mixed in with whatever resources the `StandardLibrary` had natively (whose folder layout is in fact preserved properly). It is easier to enumerate which files to use if there is just a single file.

The build plugin also needs to know the set of outputs that are to be generated by the tool call (before even calling the tool - it asks for the command to perform during build planning), so we would have to bake in the set of file names in the build plugin, which would make our generator less flexible in terms of which set of types it generates. Having a single `Generated.hylo` file makes things simpler for now from a compiler complexity standpoint, even though it's not an ideal output. I think it's a good tradeoff for now.

StandardLibrary/Sources/Core/Numbers/MockIntegerType.hylo
tothambrus11: This was needed because the generator script didn't generate all types, so the typer failed to prefill its standard library integer type cache.

Sources/hc-generate-stdlib/GenerateIntegerTypes.swift
tothambrus11: Now we need to worry about A/An depending on the number, or write `A(n)`.

Sources/hc-generate-stdlib/GenerateIntegerTypes.swift
tothambrus11: Disabled for avoiding needless performance hit, I'll change this to just use the different (dummy) emission strategy.

Sources/Driver/Driver.swift
tothambrus11: For now yes, but I don't want to bake this fact into the API of `load`, since it is not concerned about why those files are added. 

In the future we may also remove `withSourcesAt` in favor of an explicit set of source files that the build system supplies or we gather explicitly at the call site.

StandardLibrary/Sources/Core/Numbers/MockIntegerType.hylo
tothambrus11: It's now deleted.

tothambrus11:
Thanks for the feedback! I applied the changes, and now it's only a 3x slowdown of the test suite.

Now we generate
- 10 integer types with full native implementations
- `Int` with all the integer-related trait conformances (same as before)
- `Int32` with `Movable` and `ExpressibleByIntegerLiteral` conformances (same as before)

The main change is that there are now more types and that the methods are implemented on all types rather than only on `Int`.

The new version runs the CompilerTests in 60s in single-threaded debug mode, while the old suite run in about 21 seconds.

PR #78, opened 2026-04-04
Add basic Int32 type to the standard library

tothambrus11:
The LLVM lowering of `main()` requires the `Int32` type. The more comprehensive solution with an automatic generator can be seen in https://github.com/hylo-lang/hylo-new/pull/76 although it slows down the test suite by a factor of 5-8.


lucteo (commented):
LGTM

PR #81, opened 2026-04-07
Fix typos in standard library

tothambrus11:
Fixes some typos and some inaccuracies in the docs of Comparable.

### DIFF
--- a/StandardLibrary/Sources/Core/Comparable.hylo
+++ b/StandardLibrary/Sources/Core/Comparable.hylo
@@ -5,8 +5,8 @@
 /// relation that satisfies the following for all instances `a`, `b`, and `c`:
 ///
 /// - `a < a` is always `false` (irreflexivity)
-/// - if `a < b` then `b < a` is `false` (antisymmetry)
-/// - if `a < b` and `b < b` then `a < c` (transitivity)
+/// - if `a < b` then `b < a` is `false` (asymmetry)
+/// - if `a < b` and `b < c` then `a < c` (transitivity)
 /// - if `a == b` is `false` then either `a < b` or `b < a` (totality)
 ///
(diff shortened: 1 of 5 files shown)


StandardLibrary/Sources/Core/Comparable.hylo
```diff
@@ -5,8 +5,8 @@
 /// relation that satisfies the following for all instances `a`, `b`, and `c`:
 ///
 /// - `a < a` is always `false` (irreflexivity)
-/// - if `a < b` then `b < a` is `false` (antisymmetry)
-/// - if `a < b` and `b < b` then `a < c` (transitivity)
+/// - if `a < b` then `b < a` is `false` (asymmetry)
```
kyouko-taiga: It should be antisymmetry, no asymmetry. https://en.wikipedia.org/wiki/Antisymmetric_relation

kyouko-taiga (changes requested):

StandardLibrary/Sources/Core/Comparable.hylo
tothambrus11: [Asymmetry](https://en.wikipedia.org/wiki/Asymmetric_relation): `if aRb then not bRa`
Antisymmetry: `if (aRb and a!=b) then not bRa`

I guess it's implicitly implied by `a < b` that `a != b`, should we define this explicitly?

kyouko-taiga: I think that is covered by totality.

tothambrus11:
```suggestion
/// - if `a < b` then `b < a` is `false` (antisymmetry)
```

PR #89, opened 2026-04-13
Rename test expectations from .exepected to .diagnostics.expected

tothambrus11:
This prepares the the naming scheme for more kinds of expectations in the future - we no longer just have `.expected` but also an assertion kind qualification like `.diagnostics.expected` or `.stdout.expected` (future).


Tests/CompilerTests/CompilerTests.swift
```diff
@@ -12,6 +12,17 @@ import XCTest
 /// manually invoking `hc-tests`.
 final class CompilerTests: XCTestCase {
 
+  /// Cleans up observation files from previous test runs.
+  /// 
+  /// This is done at most once, during the first test execution.
```
lucteo: Note: if we have multiple types of tests that need standard library, this may not be the best place to put this (I'm adding such a test for checking the mangling of stdlib entries)

lucteo (commented):
LGTM

Tests/CompilerTests/CompilerTests.swift
tothambrus11: How would the `.observed` files interact with your test case? If you add a new test type but you don't include a `_ = initialObservationCleanup` in your test function, it wouldn't trigger the cleanup (if you only run that subset of tests).

lucteo: my tests are not in `CompilerTests.swift`, they have their own dedicated file, as the process is different.

BTW, what happens if I run the test concurrently, and we call the cleanup function while other tests is loading the standard library?

tothambrus11: The cleanup function only deletes `.observed` files within the test folder. If your tests don't emit `.observed` files, you probably don't need to clean them up. The intention behind cleanup is so that if we rename some test fixtures or if failures go away, we no longer have obsolete/outdated `.observed` files laying around.

> BTW, what happens if I run the test concurrently, and we call the cleanup function while other tests is loading the standard library?

Loading the standard library collects `*.hylo` files within the standard library root, while the `*.observed` files are put in the disjoint folder besides the test case's `.hylo` file.

tothambrus11: According to the meeting yesterday, we can remove this cleanup altogether, since hc-tests already has a functionality for this (manual action, more predictable)

PR #90, opened 2026-04-13
Save observed IR on test failure

tothambrus11:
This PR adds functionality to save the intermediate compilation artifacts (raw IR, transformed IR, llvm IR) on test failure so we don't need to recompile or rerun the tests, or run `hc` driver with the burden of setting up the same flags.

Initially I also saved the artifacts when the test cases passed, but this proved to have a large overhead. Now I gated it with a constant flag `alwaysSaveArtifacts = false` that can be used to enable it when needed.

Initially I saved the IR per file, but it ended up being more complicated and incomplete since there are some entities without a clear file origin. Now the IR is saved on a module level, including the global variables and functions.


Sources/FrontEnd/Program.swift
```diff
@@ -1632,3 +1632,17 @@ fileprivate struct ChildrenEnumerator: SyntaxVisitor {
...
+extension Program {
+
+  /// The textual representation of the IR of `module`.
+  public func irCode(_ module: Module.ID) -> String {
+    [
+      self[module].variables.map { g in show(g) },
+      self[module].functions.map { f in show(f) },
+    ]
+    .flatMap({ $0 })
+    .joined(separator: "\n\n")
```
lucteo: (minor)
Previously used a single newline between functions. Do we need to have two?

Tests/CompilerTests/CompilerTests.swift
```diff
@@ -127,16 +127,29 @@ final class CompilerTests: XCTestCase {
 
     // IR Lowering.
     if await driver.lower(m).containsError { return done() }
+    try saveObservedIR(of: m, in: driver.program, root: input.root, extension: "lowered-ir.observed")
```
lucteo: I'm not a fan of this. Why spend the extra time and write the two IR files for each tests, when 99% of time we don't need it? We can always invoke `hc` to produce the right IR whenever we need it.

Currently we don't have a lot of tests, but I would expect to have far more in the future.

That being said, I'm not opposed to this change if others are ok with it.

Tests/CompilerTests/CompilerTests.swift
```diff
@@ -127,16 +127,29 @@ final class CompilerTests: XCTestCase {
...
     return done()
   }
 
-  private func compile(_ m: Module.ID, driver: inout Driver) {
+  /// Saves the IR of module `m` for each of its source files separately.
+  ///
+  /// The outputs are written in the same directory as the source files, with the same name but
+  /// with the extension changed to `extension`.
+  private func saveObservedIR(of m: Module.ID, in program: Program, root: URL, extension: String) throws {
+    let ir = program.irCode(m)
```
lucteo: (minor) `ir` can be moved below near the usage. Also, we might not need the variable at all

lucteo (commented):
I have some concerns with the main objective of the PR, and a few minor comments.
Other than that, LGTM

Sources/FrontEnd/Program.swift
tothambrus11: It has a cleaner visual separation, we use the same convention between Swift functions too. I can revert to a single newline if there is an argument for compactness.

Tests/CompilerTests/CompilerTests.swift
tothambrus11: The motivation for emitting these is that it's somewhat cumbersome to use a CLI and set up the compilation the exact same way as the test is doing it. The test harness relies on logic that parses the manifest file, and based on that it configures a custom compilation pipeline similarly to how it's done in `hc`. I would spend 1-2 minutes to look up CLI flags and make sure everything in the test's manifest is mirrored correctly into `hc` CLI arguments. It's great if we can directly inspect the internals of the intermediate compilation stages upon a test failure without having to recompile or rerun the test with a debugger.

I thought about only saving the observed IR of failed tests by default, and emit specific successful test cases' IR only when opting in explicitly. This at least saves file IO, while still requiring TreePrinter to run on the compiled module anyway so we have the captured state by the time we sense failure.

I initially thought this would be a premature optimization, as I didn't notice significant slowdown of the test suite, but I should have made a performance comparison. It's indeed a big regression, especially in parallel test runs:

image

Now I changed it so that we only save on test failure. This needs to be extended by the later PR that adds behavioral test support, so that it would also perform the saving.

Sources/FrontEnd/Program.swift
lucteo: sounds good.

Tests/CompilerTests/CompilerTests.swift
lucteo: 80% extra time in parrallel case seems significant.

My main argument though is that this schema creates complexity while we hardly use this feature. Now we have to think about ignoring many files, mixing user-provided files with ignored files, etc.

If we can have this feature ready (i.e., dump the `.ir` if the test fail) and manually enable it when a test fails, that would be better. Consider adding a boolean or some instructions in the code on how to enable this.

That would keep the feature to be used when we really need it, and at the same time it won't create ignored files all the time, and won't consume testing time.

tothambrus11: > 80% extra time in parrallel case seems significant.

Maybe my phrasing was confusing, sorry about that. I completely agree, so I changed it to save only when tests fail (which happens much more rarely, and with much less files). When a test fails, we need as much information as possible ASAP, without having to recompile the compiler with an enable flag.

> Now we have to think about ignoring many files, mixing user-provided files with ignored files, etc.

I don't see much mental overhead here, we just ignore `*.observed` everywhere. IDEs usually display ignored files with a different color. Also, when enabling saving only for failures, we only see `.observed` files we probably care about.

PR #91, opened 2026-04-13
Improve diagnostic for failed module cache parsing

tothambrus11:
As breaking changes to the serialization of the module cache are being introduced, the caches may represent incorrect values or fail to get parsed (more likely). The latter problems are caught, but before this commit it was not obvious what's the root cause and how to fix it.

Note: the former kind of breaking changes cannot be diagnosed easily via this fix, and will need the more involved versioning system @kyouko-taiga mentioned, but it's not a priority due to most common cases being covered by parse errors.

I didn't add a death test for this case as it's not natively supported by XCTest.

### DIFF
--- a/Sources/Driver/Driver.swift
+++ b/Sources/Driver/Driver.swift
@@ -124,15 +124,24 @@ public struct Driver {
     }
 
     // Attempt to load the module from disk.
-    if cachingIsEnabled, let data = archive(of: module) {
-      let h = SourceFile.fingerprint(contentsOf: sources)
-      var a = ReadableArchive(data)
-      let (_, fingerprint) = try Module.header(&a)
-      if h == fingerprint {
-        a = ReadableArchive(data)
-        try program.load(module: module, from: &a)
-        return
+    do {
+      if cachingIsEnabled, let data = archive(of: module) {
+        let h = SourceFile.fingerprint(contentsOf: sources)
+        var a = ReadableArchive(data)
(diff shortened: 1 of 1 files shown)


Sources/Driver/Driver.swift
```diff
@@ -124,15 +124,21 @@ public struct Driver {
...
+        if h == fingerprint {
+          a = ReadableArchive(data)
+          try program.load(module: module, from: &a)
+          return
+        }
       }
+    } catch ArchiveError.invalidInput {
+      fatalError("Failed to parse module archive of '\(module)' at '\(moduleCachePath, default: "nil")'. "
+        + "Maybe the archive is compiled using a different version of the compiler.\n"
+        + "Try erasing the module cache.")
```
```suggestion
      let m = """
        Failed to parse the archive of '\(module)' at '\(moduleCachePath, default: "nil")'.
        Maybe the archive is compiled using a different version of the compiler.\n"
        Try erasing the module cache.
        """
      fatalError(m)
```

kyouko-taiga (approved):
LGTM. I made a small suggestion.

Sources/Driver/Driver.swift
tothambrus11: Interesting, it looks better in code but changes the displayed text has extra forced newlines instead of soft wrapping. YAML multiple kinds of string literals, some of which allow you to write multiline string where each line is just delimited by spaces.

PR #93, opened 2026-04-14
Platform utilities

tothambrus11:
Platform utilities for representing the architectures and operating systems at runtime. These additions will be useful for the driver to invoke the linker and the output executable.

The original version can be found at https://github.com/hylo-lang/hylo/blob/main/Sources/Utils/Host.swift and https://github.com/hylo-lang/hylo/blob/main/Sources/Utils/Platform.swift

The newly added utilities for resolving an executable on the PATH only supports resolving and running and resolving native executables, not arbitrary shell files that cmd or powershell could execute. For now, this is sufficient, as the only tool we depend on at runtime is the Clang driver. Adding support for ps1/cmd/bat files would require running them in a shell (they cannot be executed natively), so I wanted to avoid that extra complexity. With this constraint in mind, it's also sufficient to use the PATH based lookup on Windows instead of calling the `where.exe` subprocess.

### DIFF
--- a/Sources/Utilities/Host.swift
+++ b/Sources/Utilities/Host.swift
@@ -1,0 +1,95 @@
+import Foundation
+
+/// The platform on which the compiler or interpreter is running.
+public enum Host: Sendable {
+
+  #if os(macOS)
+    /// The host operating system.
+    public static let operatingSystem: Platform.OperatingSystem = .macOS
+  #elseif os(Linux)
+    /// The host operating system.
+    public static let operatingSystem: Platform.OperatingSystem = .linux
+  #elseif os(Windows)
+    /// The host operating system.
+    public static let operatingSystem: Platform.OperatingSystem = .windows
+  #else
+    #error("Unsupported host operating system")
+  #endif
+
+  #if arch(x86_64)
+    /// The host architecture.
+    public static let architecture: Platform.Architecture = .x86_64
+  #elseif arch(arm64)
+    /// The host architecture.
+    public static let architecture: Platform.Architecture = .arm64
+  #else
+    #error("Unsupported host architecture")
+  #endif
+
+  /// A view of the environment variables.
+  public struct Environment: Sendable {
+
+    /// Returns the value of the environment variable named `key`, if any.
+    ///
+    /// On Windows, the comparison is case-insensitive and takes linear time.
+    public subscript(_ key: String) -> String? {
+      #if os(Windows)
+        ProcessInfo.processInfo.environment
+          .first(where: { $0.key.caseInsensitiveCompare(key) == .orderedSame })?.value
+      #else
+        ProcessInfo.processInfo.environment[key]
+      #endif
+    }
+
+    /// Returns the value of the environment variable named `key` or `d` if it's not set.
+    ///
+    /// On Windows, the comparison is case-insensitive and takes linear time.
+    public subscript(_ key: String, default d: String) -> String {
+      self[key] ?? d
+    }
+
+  }
+
+  /// The environment variables of the current process.
+  public static let environment = Environment()
+
+  /// The locations of the executable search path (aka the `PATH` environment variable).
+  public static func searchPathLocations() -> [String] {
+    Host.environment["PATH", default: ""]
+      .split(separator: Host.searchPathSeparator)
+      .map(String.init)
+  }
+
+  /// The separator between individual locations in executable search path.
+  public static let searchPathSeparator: Character = Host.operatingSystem == .windows ? ";" : ":"
+
+  /// The suffix of native executables.
+  public static let nativeExecutableSuffix = operatingSystem == .windows ? ".exe" : ""
+
+  /// Returns the location of the native executable invoked as `name` using the search path or
+  /// throws `ExecutableNotFound` if no executable could be found.
+  ///
+  /// `name` shall be supplied without the native executable suffix, e.g. without `.exe` on Windows.
+  /// Only native executables are resolved. Script files such as `.cmd`, `.bat`, and `.ps1` are not.
+  public static func findNativeExecutable(invokedAs name: String) throws -> URL {
+   for base in searchPathLocations() {
+     let p =  URL(fileURLWithPath: base).appendingPathComponent(name + nativeExecutableSuffix)
+     if FileManager.default.isExecutableFile(atPath: p.path) { return p }
+   }
+   throw ExecutableNotFound(name: name)
+  }
+
+  /// Error thrown when an executable is not found on the PATH.
+  public struct ExecutableNotFound: Error, CustomStringConvertible {
+    
+    /// Name of the executable without native executable suffix.
+    public let name: String
+
+    /// A description of the error.
+    public var description: String {
+      "Executable not found on PATH: \(name)"
+    }
+
+  }
+
+}
--- a/Sources/Utilities/Platform.swift
+++ b/Sources/Utilities/Platform.swift
@@ -1,0 +1,32 @@
+/// The platform on which the compiler or interpreter is running/targeting.
+public enum Platform: Sendable {
+
+  /// An operating system on which the Hylo compiler toolchain can run as a host.
+  public enum OperatingSystem: String, Codable, CustomStringConvertible, Sendable {
+
+    case macOS = "macOS"
+
+    case linux = "Linux"
+
+    case windows = "Windows"
+
+    /// A description of `self`.
+    public var description: String {
+      rawValue
+    }
+
+  }
+
+  /// An architecture on which the Hylo compiler toolchain can run as a host.
+  public enum Architecture: String, Codable, CustomStringConvertible, Sendable {
+
+    case x86_64, arm64
+
+    /// String representation.
+    public var description: String {
+      return rawValue
+    }
+
+  }
+
+}
--- a/Sources/Utilities/Process+Extensions.swift
+++ b/Sources/Utilities/Process+Extensions.swift
@@ -1,0 +1,100 @@
+import Foundation
+
+extension Process {
+
+  /// The error thrown when a process exits with a non-zero status.
+  public struct NonzeroExit: Error {
+
+    /// The exit code of the process.
+    public let exitCode: Int32
+
+    /// The data written to the standard output of the process.
+    public let standardOutput: String
+
+    /// The data written to the standard error of the process.
+    public let standardError: String
+
+    /// The path to the executable ran by the process.
+    public let executable: URL
+
+    /// The arguments passed to the process.
+    public let arguments: [String]
+
+  }
+
+  /// Runs `executable` with `arguments` and returns the data written to the standard output.
+  ///
+  /// Throws a `NonzeroExit` upon terminating with non-zero exit code.
+  public static func executionOutput(
+    _ executable: URL, arguments: [String] = []
+  ) throws -> String {
+    let process = Process()
+    let standardOutput = Pipe()
+    let standardError = Pipe()
+    process.executableURL = executable
+    process.arguments = arguments
+    process.standardOutput = standardOutput
+    process.standardError = standardError
+    try process.run()
+    process.waitUntilExit()
+
+    let output = String(decoding: standardOutput.fileHandleForReading.readDataToEndOfFile(), as: UTF8.self)
+    let error = String(decoding: standardError.fileHandleForReading.readDataToEndOfFile(), as: UTF8.self)
+
+    if process.terminationStatus != 0 {
+      throw NonzeroExit(
+        exitCode: process.terminationStatus,
+        standardOutput: output,
+        standardError: error,
+        executable: executable,
+        arguments: arguments)
+    }
+
+    return output
+  }
+
+  /// Runs `executable` with `arguments` and returns its execution report.
+  public static func execute(
+    _ executable: URL, arguments: [String] = []
+  ) throws -> ExecutionReport {
+    let process = Process()
+    let standardOutput = Pipe()
+    let standardError = Pipe()
+    process.arguments = arguments
+    process.executableURL = executable
+    process.standardOutput = standardOutput
+    process.standardError = standardError
+    try process.run()
+    process.waitUntilExit()
+
+    let output = String(decoding: standardOutput.fileHandleForReading.readDataToEndOfFile(), as: UTF8.self)
+    let error = String(decoding: standardError.fileHandleForReading.readDataToEndOfFile(), as: UTF8.self)
+
+    return .init(
+      standardOutput: output,
+      standardError: error,
+      exitCode: process.terminationStatus)
+  }
+
+
+  /// The result of executing a process.
+  public struct ExecutionReport {
+
+    /// The data written to the standard output of the process.
+    public let standardOutput: String
+
+    /// The data written to the standard error of the process.
+    public let standardError: String
+
+    /// The exit code of the process.
+    public let exitCode: Int32
+
+    /// Creates an instance from its parts.
+    public init(standardOutput: String, standardError: String, exitCode: Int32) {
+      self.standardOutput = standardOutput
+      self.standardError = standardError
+      self.exitCode = exitCode
+    }
+
+  }
+}
--- a/Tests/UtilitiesTests/HostTests.swift
+++ b/Tests/UtilitiesTests/HostTests.swift
@@ -1,0 +1,78 @@
+import Foundation
+import Utilities
+import XCTest
+
+typealias Host = Utilities.Host
+
+
+final class HostTests: XCTestCase {
+
+  func testFindNativeExecutableThrowsForUnknownCommand() throws {
+    XCTAssertThrowsError(
+      try Host.findNativeExecutable(invokedAs: "randomNotFoundExecutable"), "",
+      { (error) in
+        if let e = error as? Host.ExecutableNotFound {
+          XCTAssertEqual(e.name, "randomNotFoundExecutable")
+          XCTAssertEqual(e.description, "Executable not found on PATH: randomNotFoundExecutable")
+        } else {
+          XCTFail("Expected Host.ExecutableNotFound, got \(error)")
+        }
+      })
+  }
+
+  #if os(Windows)
+    func testFindNativeExecutableFindsAndExecutesWhereExe() throws {
+      let whereExe = try Host.findNativeExecutable(invokedAs: "where")
+      XCTAssertEqual(whereExe.lastPathComponent.lowercased(), "where.exe")
+
+      let output = try Process.executionOutput(whereExe, arguments: ["cmd"])
+      XCTAssertTrue(output.lowercased().contains("cmd.exe"))
+    }
+  #else
+    func testFindNativeExecutableFindsAndExecutesBash() throws {
+      let bash = try Host.findNativeExecutable(invokedAs: "bash")
+      XCTAssertEqual(bash.lastPathComponent, "bash")
+
+      let output = try Process.executionOutput(bash, arguments: ["-lc", "printf '%s' bash-ok"])
+      XCTAssertEqual(output, "bash-ok")
+    }
+  #endif
+
+  func testExecutionOutputThrowsOnNonzeroExit() throws {
+    #if os(Windows)
+      let executable = try Host.findNativeExecutable(invokedAs: "cmd")
+      let arguments = ["/c", "exit", "42"]
+    #else
+      let executable = try Host.findNativeExecutable(invokedAs: "bash")
+      let arguments = ["-lc", "exit 42"]
+    #endif
+
+    XCTAssertThrowsError(
+      try Process.executionOutput(executable, arguments: arguments), "",
+      { (error) in
+        if let e = error as? Process.NonzeroExit {
+          XCTAssertEqual(e.exitCode, 42)
+          XCTAssertEqual(e.executable, executable)
+          XCTAssertEqual(e.arguments, arguments)
+        } else {
+          XCTFail("Expected Process.NonzeroExit, got \(error)")
+        }
+      })
+  }
+
+  func testExecuteReturnsReportOnNonzeroExit() throws {
+    #if os(Windows)
+      let executable = try Host.findNativeExecutable(invokedAs: "cmd")
+      let arguments = ["/c", "exit", "42"]
+    #else
+      let executable = try Host.findNativeExecutable(invokedAs: "bash")
+      let arguments = ["-lc", "exit 42"]
+    #endif
+
+    let r = try Process.execute(executable, arguments: arguments)
+    XCTAssertEqual(r.exitCode, 42)
+    XCTAssertEqual(r.standardOutput, "")
+    XCTAssertEqual(r.standardError, "")
+  }
+
+}
--- a/Tests/UtilitiesTests/PlatformTests.swift
+++ b/Tests/UtilitiesTests/PlatformTests.swift
@@ -1,0 +1,18 @@
+import Foundation
+import Utilities
+import XCTest
+
+final class PlatformTests: XCTestCase {
+
+  func testArchitectureDescription() {
+    XCTAssertEqual(Platform.Architecture.arm64.description, "arm64")
+    XCTAssertEqual(Platform.Architecture.x86_64.description, "x86_64")
+  }
+
+  func testOperatingSystemDescription() {
+    XCTAssertEqual(Platform.OperatingSystem.macOS.description, "macOS")
+    XCTAssertEqual(Platform.OperatingSystem.linux.description, "Linux")
+    XCTAssertEqual(Platform.OperatingSystem.windows.description, "Windows")
+  }
+
+}


Sources/Utilities/Host.swift
```diff
@@ -0,0 +1,94 @@
...
+    /// The host architecture.
+    public static let architecture: Platform.Architecture = .x86_64
+  #elseif arch(arm64)
+    /// The host architecture.
+    public static let architecture: Platform.Architecture = .arm64
+  #else
+    #error("Unsupported host architecture")
+  #endif
+
+  /// A subscriptable view of the environment variables, working around the lack of named subscripts.
```
kyouko-taiga:
```suggestion
  /// A view of the environment variables.
```

Is it relevant that the view is "subscriptable"?

Also there's no need to bash Swift in the comments of our compiler ;)

Sources/Utilities/Host.swift
```diff
@@ -0,0 +1,94 @@
...
+  public struct Environment: Sendable {
+
+    /// Looks up the value of the environment variable with the given `key`.
```
kyouko-taiga:
```suggestion
    /// Returns the value of the environment variable named `key`, if any.
```

Sources/Utilities/Host.swift
```diff
@@ -0,0 +1,94 @@
...
+    public subscript(_ key: String) -> String? {
+      #if os(Windows)
+        ProcessInfo.processInfo.environment
+          .first(where: { $0.key.caseInsensitiveCompare(key) == .orderedSame })?.value
+      #else
+        ProcessInfo.processInfo.environment[key]
+      #endif
+    }
+
+    /// Looks up the value of the environment variable with the given `key`.
```
kyouko-taiga:
```suggestion
    /// Returns the value of the environment variable named `key` or `d` if it is not set.
```

Sources/Utilities/Host.swift
```diff
@@ -0,0 +1,94 @@
...
+    ///
+    /// If the environment variable is not set, returns `default`.
```
kyouko-taiga:
```suggestion
```

Sources/Utilities/Host.swift
kyouko-taiga: I would probably add the mention here that name lookup is case insensitive and runs in linear time on Windows, and not repeat that elsewhere.

Sources/Utilities/Host.swift
```diff
@@ -0,0 +1,94 @@
...
+    /// On Windows, the comparison is case-insensitive and takes linear time.
+    public subscript(_ key: String, default d: String) -> String {
+      self[key] ?? d
+    }
+
+  }
+
+  /// A subscriptable view of the environment variables, using case-insensitive comparison on Windows.
```
kyouko-taiga:
```suggestion
  /// The environment variables of the current process.
```

Sources/Utilities/Host.swift
```diff
@@ -0,0 +1,94 @@
...
+  public static let environment = Environment()
+
+  /// The elements of the environment's executable search path.
+  public static func pathElements() -> [String] {
```
kyouko-taiga:
```suggestion
  /// The locations of the executable search path (aka the `PATH` environment variable).
  public static func searchPathLocations() -> [String] {
```

Sources/Utilities/Host.swift
```diff
@@ -0,0 +1,94 @@
...
+    Host.environment["PATH", default: ""]
+      .split(separator: Host.pathEnvironmentSeparator)
+      .map(String.init)
+  }
+
+  /// The separator between elements of the environment's executable search path.
+  public static let pathEnvironmentSeparator: Character = Host.operatingSystem == .windows ? ";" : ":"
```
kyouko-taiga:
```suggestion
  /// The separator between individual locations in executable search path.
  public static let searchPathSeparator: Character = Host.operatingSystem == .windows ? ";" : ":"
```

Sources/Utilities/Host.swift
```diff
@@ -0,0 +1,94 @@
...
+
+  /// The suffix of native executables.
+  public static let nativeExecutableSuffix = operatingSystem == .windows ? ".exe" : ""
```
kyouko-taiga, line 67: I remember comments from @dabrahams about this suffix. Have these comments been addressed elsewhere?

In case it may be useful here are links I found on Slack about this discussion:
- https://github.com/dabrahams/SPMBuildToolSupport/blob/c270d23a9c70aa015e09a9c0ea28e36d2801adc4/SPMBuildToolSupport.swift
- https://debbugs.gnu.org/cgi/bugreport.cgi?bug=67540

Sources/Utilities/Host.swift
```diff
@@ -0,0 +1,94 @@
...
+
+  /// Finds the native executable invoked as `name` in the environment's executable search path or throws `ExecutableNotFound` otherwise.
```
kyouko-taiga:
```suggestion
  /// Returns the location of the native executable invoked as `name` using the search path or
  /// throws `ExecutableNotFound` if no executablecould be found.
```

I'm not convinced it is necessary to specify the type of the exception being thrown but I wouldn't strongly object either.

Sources/Utilities/Host.swift
```diff
@@ -0,0 +1,94 @@
...
+  ///
+  /// On Windows, `name` shall be supplied without the `.exe` suffix.
+  /// Only native executables are resolved. Script files such as `.cmd`, `.bat`, and `.ps1` are not.
```
kyouko-taiga:
```suggestion
  /// On Windows, `name` shall be supplied sans `.exe` suffix. Only "native" executables are
  /// resolved; scripts suffixed by e.g. `.cmd`, `.bat`, or `.ps1` are ignored.
```

Is that true only on Windows? If not I would reorder the sentences.

Sources/Utilities/Host.swift
```diff
@@ -0,0 +1,94 @@
...
+  public static func findNativeExecutable(invokedAs name: String) throws -> URL {
+    try pathElements().firstNonNil { base in
+      let p = URL(fileURLWithPath: base).appendingPathComponent(name + nativeExecutableSuffix)
+      return FileManager.default.isExecutableFile(atPath: p.path) ? p : nil
+    }.unwrapOrThrow(ExecutableNotFound(name: name))
```
kyouko-taiga:
```suggestion
    for base in pathElements() {
      let p =  URL(fileURLWithPath: base).appendingPathComponent(name + nativeExecutableSuffix)
      if FileManager.default.isExecutableFile(atPath: p.path) { return p }
    }
    throw ExecutableNotFound(name: name)
```
I think this suggestion is much simpler than the one using closures all over the place.

But more generally:
- We should avoid calling methods on the result of a function applied with trailing closure syntax.
- We should wrap as little code as possible in `try` blocks.
- I insist on systematically adding parentheses around the arguments of closures.

Sources/Utilities/Host.swift
```diff
@@ -0,0 +1,94 @@
...
+  }
+
+  /// Error thrown when an executable is not found on the PATH.
+  public struct ExecutableNotFound: Error, CustomStringConvertible {
+    
+    /// Name of the executable without native executable suffix.
+    public let name: String
+
+    /// Human-friendly error message.
```
kyouko-taiga:
```suggestion
    /// A description of the error.
```

Sources/Utilities/Platform.swift
```diff
@@ -0,0 +1,32 @@
+/// The platform on which the compiler or interpreter is running/targeting.
+public enum Platform: Sendable {
+
+  /// An operating system on which the compiler/interpreter/target program can run.
```
kyouko-taiga: It took me a while to understand what "compiler/interpreter/target program can run". Can we reformulate? In particular I find it ambiguous to separate two-words phrases with a forward slash.

Sources/Utilities/Platform.swift
```diff
@@ -0,0 +1,32 @@
...
+  public enum OperatingSystem: Codable, CustomStringConvertible, Sendable {
+
+    case macOS, linux, windows
+
+    /// String representation that matches the possible values used for conditional compilation.
```
kyouko-taiga: That's a very verbose documentation.

More simply I think we could define this:
```
public enum OperatingSystem: String, CustomStringConvertible {

  case macOs = "macOS"

  case linux = "Linux"

  case windows = "Windows"

    rawValue
  }

}
```

But I'm not sure to understand why that is necessary at all that we have the textual description being the same as that of conditional compilation.

Sources/Utilities/Platform.swift
```diff
@@ -0,0 +1,32 @@
...
+      switch self {
+      case .macOS: return "macOS"
+      case .linux: return "Linux"
+      case .windows: return "Windows"
+      }
+    }
+
+  }
+
+  /// An architecture on which the compiler/interpreter/target program can run.
```
kyouko-taiga: Same comment as above regarding the slashes.

Sources/Utilities/Process+Extensions.swift
```diff
@@ -0,0 +1,55 @@
...
+    /// The exit code of the terminated process.
+    public let exitCode: Int32
+
+    /// The collected standard output of the terminated process.
+    public let standardOutput: String
+
+    /// The collected standard error of the terminated process.
+    public let standardError: String
+
+    /// The executable used to run the terminated process.
```
kyouko-taiga: This description makes me believe that we need to run an executable to terminate the process, which most likely isn't the right interpretation. I guess it is:

```suggestion
    /// The path to the executable ran by the process.
```

Sources/Utilities/Process+Extensions.swift
```diff
@@ -0,0 +1,55 @@
+import Foundation
+
+extension Process {
+
+  /// The error thrown when a process exits with a non-zero status.
+  public struct NonzeroExit: Error {
+
+    /// The exit code of the terminated process.
```
kyouko-taiga:
```suggestion
    /// The exit code of the process.
```
If the process exited then it has terminated.

Sources/Utilities/Process+Extensions.swift
```diff
@@ -0,0 +1,55 @@
...
+    public let exitCode: Int32
+
+    /// The collected standard output of the terminated process.
```
kyouko-taiga:
```suggestion
    /// The data written to the standard output of the process.
```

Sources/Utilities/Process+Extensions.swift
```diff
@@ -0,0 +1,55 @@
...
+    public let standardOutput: String
+
+    /// The collected standard error of the terminated process.
```
kyouko-taiga:
```suggestion
    /// The data written to the standard error of the process.
```

Sources/Utilities/Process+Extensions.swift
```diff
@@ -0,0 +1,55 @@
...
+    public let standardError: String
+
+    /// The executable used to run the terminated process.
+    public let executable: URL
+
+    /// The arguments passed to the executable.
```
kyouko-taiga:
```suggestion
    /// The arguments passed to the process.
```
The executable (or at least its path) is itself an argument.

Sources/Utilities/Process+Extensions.swift
```diff
@@ -0,0 +1,55 @@
...
+    public let arguments: [String]
+
+  }
+
+  /// Runs `executable` with `arguments` and returns its standard output.
```
kyouko-taiga:
```suggestion
  /// Runs `executable` with `arguments` and returns the data written to the standard output.
```

Sources/Utilities/Process+Extensions.swift
```diff
@@ -0,0 +1,55 @@
...
+  ///
+  /// Throws a `NonzeroExit` upon terminating with non-zero exit code.
```
kyouko-taiga, line 27: It seems to me that callers of this method are very likely to systematically handle the `NonzeroExit` exception. If my assumption is correct, I would make the return type a `Result` instead of throwing.

Tests/UtilitiesTests/HostTests.swift
```diff
@@ -0,0 +1,57 @@
+import Foundation
+import Utilities
+import XCTest
+typealias Host = Utilities.Host
```
kyouko-taiga, line 5:
```suggestion

typealias Host = Utilities.Host
```

Tests/UtilitiesTests/HostTests.swift
```diff
@@ -0,0 +1,57 @@
...
+
+final class HostTests: XCTestCase {
+
+  func testFindNativeExecutableThrowsForUnknownCommand() throws {
+    XCTAssertThrowsError(try Host.findNativeExecutable(invokedAs: "randomNotFoundExecutable")) { error in
+      guard let e = error as? Host.ExecutableNotFound else {
+        XCTFail("Expected Host.ExecutableNotFound, got \(error)")
+        return
+      }
```
kyouko-taiga:
```suggestion
      let e = try XCTUnwrap(error as? Host.ExecutableNotFound)
```

Tests/UtilitiesTests/HostTests.swift
```diff
@@ -0,0 +1,57 @@
...
+    #else
+      let executable = try Host.findNativeExecutable(invokedAs: "bash")
+      let arguments = ["-lc", "exit 42"]
+    #endif
+
+    XCTAssertThrowsError(try Process.executionOutput(executable, arguments: arguments)) { error in
+      guard let e = error as? Process.NonzeroExit else {
+        XCTFail("Expected Process.NonzeroExit, got \(error)")
+        return
+      }
```
kyouko-taiga:
```suggestion
      let e = try XCTUnwrap(error as? Process.NonzeroExit)
```

kyouko-taiga (changes requested):
Nice work. Request changes are in my comments.

Sources/Utilities/Host.swift
tothambrus11: Yes, I will remove these details from the docs of `environment`. There is another version of the subscript that takes a default, I would like to keep these details there, since it's on the same level from the user's point of view.

tothambrus11: It would be a nice language feature to reuse documentation fragments, so we don't need to maintain duplication.

Sources/Utilities/Host.swift
tothambrus11: The comment was that on Windows, typing `myProgram` in the terminal can resolve to any of `{ myProgram.exe, myProgram.bat, myProgram.cmd, myProgram.ps1 }`, and tool developers may decide to replace their executable with a wrapper script instead, so it would be nice to be resilient to this.

However, after implementing this, I realized that it brings a great deal of extra complexity, and given that `hc` currently has exactly 1 dependency (clang for linking), I don't think it's worth adding. My feeling is also that `clang.exe` will never suddenly become `clang.ps1`, since it has such a big user base, they wouldn't risk the effects of such breaking change.
 
<details>
<summary>Details of the complexity</summary>
On Windows, the only native executable suffix is `.exe`. Everything else is something that the shells are able to interpret. The set is different between CMD and PowerShell. Originally, I made a version that emulated the PowerShell's search algorithm for `findNativeExecutable` to match the behavior of the `where` utility on Windows, like in SPMBuildToolSupport. This had lots of additional complexity and supposedly performance penalties, due to having to launch a PowerShell child process to execute whatever executable or script has been found. It also has correctness/security implications, as we need to construct and sanitize input given to PowerShell, including any necessary quoting of the arguments. Launching a native `.exe` executable doesn't have this issue, as the arguments are directly passed to the executable one by one, rather than being parsed by a shell.
</details>

Sources/Utilities/Host.swift
tothambrus11: From the operating systems I know of, only Windows has a suffix for native executables, so it's mostly relevant on Windows, while being vacuously true on other platforms. I reordered the sentence to de-emphasize Windows.

*`name` shall be supplied without the native executable suffix, e.g. without `.exe` on Windows.
  Only native executables are resolved. Script files such as `.cmd`, `.bat`, and `.ps1` are not.*

Sources/Utilities/Host.swift
tothambrus11: Indeed, this unwrapOrThrow makes the control flow harder to understand, and necessitates `try` around a larger scope. I agree with the guidelines, added them to the conventions PR.

I would like to keep the use of an algorithm though, it documents the intent of searching. What do you think about the following version?

```swift
let match = searchPathLocations().firstNonNil { (base) in
  let p = URL(fileURLWithPath: base).appendingPathComponent(name + nativeExecutableSuffix)
  return FileManager.default.isExecutableFile(atPath: p.path) ? p : nil
}
if let match { return match }
    
throw ExecutableNotFound(name: name)
```

Sources/Utilities/Platform.swift
tothambrus11: Rephrased as: `An operating system on which the Hylo compiler toolchain can run as a host.`

Sources/Utilities/Platform.swift
tothambrus11: Yeah, this detail is not important indeed, I simplified the description. I was under the belief that raw values would increase the storage required for the enum, but now I checked. Swift is smart, it only stores what's needed, and gets the raw values on demand.

Tests/UtilitiesTests/HostTests.swift
tothambrus11: That's a nice change. Unfortunately the overload of `XCTAssertThrowsError` that accepts a throwing closure is missing in our CI on macOS.  I tried a couple of things to fix it, including installing a specific version of Xcode & SDK, but most likely the only fix *we* could currently do is to upgrade to macOS26 (which GitHub already supports). See the relevant reported issue: https://github.com/swiftlang/swift-corelibs-xctest/issues/543

If possible, I would wait with upgrading the macOS version, so we ensure the Hylo toolchain works on slightly older macOS versions too. If we only tested on later versions, we might accidentally use something that is not present in the old SDK.

Sources/Utilities/Process+Extensions.swift
tothambrus11: For executing the tests, we do indeed need to handle the result immediately including the checking of the exit code, for which I added a separate function (now I moved it into this PR).

This specific function is used in places where the users are interested in the output, and expect that the process succeeds during normal operations (the usage is invoking clang for linking). So it's convenient to keep a throwing version.

Tests/UtilitiesTests/HostTests.swift
kyouko-taiga: Then I have another suggestion.

```swift
func testFindNativeExecutableThrowsForUnknownCommand() throws {
    XCTAssertThrowsError(try Host.findNativeExecutable(invokedAs: "randomNotFoundExecutable")) { (e) in
      if let x = e as? Host.ExecutableNotFound {
        XCTAssertEqual(x.name, "randomNotFoundExecutable")
        XCTAssertEqual(x.description, "Executable not found on PATH: randomNotFoundExecutable") 
      } else {
        XCTFail("Expected 'Host.ExecutableNotFound', found '\(error)'")
      }
  }
```

In general I think we should avoid early returns. That is obviously not an absolute rule but unless other aspects of the code are significantly hindered, it is easier to make sense of control flow.

kyouko-taiga: By the way I think the code goes over 100 columns. I would fix by assigning `"randomNotFoundExecutable"` to a local variable.

Sources/Utilities/Host.swift
kyouko-taiga: SGTM

Sources/Utilities/Host.swift
kyouko-taiga: I still think it is more complicated than a for loop that returns early.

In terms of intent, it is not exactly clear what's going on because we're finding the first non-nil result of a notional collection that is built in the abstract, by wrapping concrete values into optionals. This indirection shows up twice. Once in the closure, in the form of a ternary, and once outside, in the form of a pattern match. That does not look good to me.

Besides, `if let match { return match }` looks silly IMO. I think we can write `return try match ?? ExecutableNotFound(name: name)`

Sources/Utilities/Host.swift
dabrahams: I think this argument is wrong. Two one-by-one subprocess calls, the first to where.exe and the second to its result, are all that's needed and no shell is involved. If you decide I'm wrong about something, raising it with me rather than simply going ahead with something else without telling me would be appreciated

Sources/Utilities/Host.swift
dabrahams: ? There are no try blocks in Swift and if you are arguing that one try should not be used for a large expression then you are arguing that granular try marking is important and I think we've agreed it's actually harmful.

Tests/UtilitiesTests/HostTests.swift
dabrahams: IMO trying to avoid early returns is counterproductive. I suppose it's _possible_ to use them in a confusing way… But why would you?

Sources/Utilities/Platform.swift
dabrahams: That phrasing implies the tool chain is the host, which I doubt is what you mean. Are the last three words adding anything useful?

Tests/UtilitiesTests/HostTests.swift
kyouko-taiga: The rationale is that in general

```swift
guard condition else { return b }
return a
```

or

```swift
if condition { return a }
return b
```

are both less straightforward than

```swift
if condition { return a } else { return b }
```

Early returns should be used if the alternative makes the code look worse, either because of excessive indentation or because it makes it difficult to bind variables. Otherwise I think a more straightforward control flow should always be preferred.

In my experience, early returns are often a sign that some abstraction is missing. We can find many real examples throughout `hc`'s codebase.

For example, I consistently refactor the pattern below.

```swift
func f() -> T? {
  let x: A
  if let y = g() {
    if let z = h(y) { x = A(z) } else { return nil }
  } else {
    x = A()
  }

  someMoreCode(using: x)
}
```

into the following:

```swift
func f() -> T? {
  if let x = someMeaningfulAbstraction() {
    return someMoreCode(using: x)
  } else {
    return nil
  }
}
```
*(also fine with `map` here, but I wouldn't necessarily recommend to replace it for all `if` statements in general)*

There two main issues with the first version IMO:

1. The fact that the first block of code may return early is not easy to spot at a first glance.
2. The fact that we need complex logic to initialize `x` heavily suggests that an abstraction is missing.

Using a guard is not much better. While it addresses (1), it also introduces another problem: the local namespace of `f` will be polluted with `y`, whose only utility is to initialize of `x`. No other code in `f` is using it.

I'm not suggesting there are no use cases for early returns (see above) but I maintain that they should be avoided unless there's a compelling reason to do otherwise.

kyouko-taiga: Come to think of it, that's sort of my own "no raw loops" argument, only applied to control flow ;)

tothambrus11: This example is clearly more complex, but that's not inherent to all early returns. I agree that `if condition { return a } else { return b }` is simpler when both branches are roughly equal in size, and small.

I think `guard`-style early returns are helpful especially when the normal path has more code, and a regular `if` would separate the `else` branch from its condition to a degree where it's not obvious what condition's negation it's handling:

```swift
func f() -> T? {
  if let x = someMeaningfulAbstraction() {
     for i in 0..<x {
       heatCPU(i)
     }
     return T()
  } else {
    return nil
  }
}
```
vs
```swift
func f() -> T? {
  guard let x = someMeaningfulAbstraction() else { return nil }
  // No longer worry about unhappy path...

  for i in 0..<x {
    heatCPU(i)
  }
  return T()
}
```

When possible, I try avoiding control flow and return statements, for which `if` expressions are perfect. Also, in my PR snippet's case, by using if/else, you could completely avoid using return statements, which made things look simpler for me.

kyouko-taiga: I don't think having more code is problematic. The issue is when avoiding early returns introduces excessive indentation.

I prefer your first version to the second. We can add the same comment:

```swift
func f() -> T? {
  if let x = someMeaningfulAbstraction() {
    // No longer worry about unhappy path...
    for i in 0..<x { heatCPU(i) }
    return T()
  } else {
    return nil
  }
}
```

If the two branches have vastly different number of lines, I suspect an abstraction I missing.

Also I think we are overly attached to newlines when we use `for` loops. No one is shocked when they see `xs.map((x) => f(x))` yet we tend to add needless newlines to write `for x in xs { f(x) }`.

dabrahams: Your rationale posits a single form for early returns that doesn't match the majority of what you see in practice.  Early returns should be used the same way exceptions are: almost no logic in the guard block—preferably just `return`, and significant logic in the function following that block.  And that's mostly what I see.

dabrahams: I very much agree with quickly eliminating unhappy paths and leaving the majority of a function reading as the happy path.  That is one readability principle behind the use of exceptions, and it works.

kyouko-taiga: Well, then we disagree (mainly on the fact that most early returns are just `else { return }`) and I guess it will be dealer's choice. I will keep suggesting eliminating most early returns in my code reviews but won't block merging if people feel strongly against it.

PR #102, opened 2026-04-16
Add coding conventions

tothambrus11:
- Introduces CONVENTIONS.md with guidelines for documentation, contracts, errors, testing, and formatting.
- Updates CONTRIBUTING.md to reference the new conventions and a copy of the LLM use section from our website.
- Adds .github/copilot-instructions.md to direct Github Copilot reviews to follow the conventions.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -4,3 +4,6 @@ publish a release when editing the draft.
 
 The conventional tag name for releases is `vX.Y.Z`, with an optional suffix for pre-releases (like `v0.0.2-test-lsp`).
 The release workflow only triggers for a pattern matching `v*`.
+
+## Coding Conventions
+Consult [CONVENTIONS.md](CONVENTIONS.md) for coding conventions used in this project.
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -7,3 +7,17 @@ The release workflow only triggers for a pattern matching `v*`.
 
 ## Coding Conventions
 Consult [CONVENTIONS.md](CONVENTIONS.md) for coding conventions used in this project.
+
+## LLM usage
+If you use an LLM to generate code, please disclose this in your Pull Request description.
+
+While we embrace tools that improve efficiency, we must avoid committing code we do not understand. If your PR includes
+LLM-generated code that is complex or opaque to you:
+
+1. Stop
+2. Understand and document what it is doing
+3. Simplify it so that others can follow.
+
+If you are unable to do these things, bring your problem up on Slack and work through a resolution with the community.
+
+Ultimately, you are responsible for the code you commit. Ensure you review and test generated code thoroughly, and always alert reviewers to it.
--- a/CONVENTIONS.md
+++ b/CONVENTIONS.md
@@ -33,6 +33,8 @@ or usages to infer meaning.
 - Conformance implementations are exempt from documentation when nothing useful can be added beyond the protocol
   requirement itself.
 - Test cases don't need to be documented, but should have a descriptive function name.
+- For types, the summary should say what the type *is*. For functions and subscripts, it should say what they *do* (e.g. returns, yields ...).
+- Describe if there is a significance in the element order of an array.
 
 ## Contracts
 
--- a/CONVENTIONS.md
+++ b/CONVENTIONS.md
@@ -79,3 +79,19 @@ Avoid typed `throws` unless you are confident that the immediate caller will be 
 - Use explicit named parameters in parentheses for multi-statement closures: `{ element in ... }`. Use `$0` shorthand only in short, single-expression closures.
 - Use `// MARK:` comments to organize sections in large files.
 - Use `self.` in initializers when assigning to stored properties.
+
+## File names
+
+All Swift source files end with the extension `.swift` and all Hylo source files end with the extension `.hylo`.
+
+In general, a file is named after the main entity that it defines. A file that extends an existing type with a protocol (Swift) or trait (Hylo) conformance is named with a combination of the type name and the protocol or trait name, joined with a plus (`+`) sign. For more complex situations, exercise your best judgment.
+
+For example:
+
+- A Swift file defining a type named MyType is named MyType.swift.
+- A Swift file defining how MyType conforms to Equatable is named MyType+Equatable.swift.
+- Retroactive conformances can be added in `MyType+Extensions.swift`.
+
+Avoid defining multiple types, protocols, or traits in the same file unless they are scoped by a main entity or meant to
+be used only in that file. Usually, conformances and custom error types are small and are defined in the same file as 
+the type to which they apply.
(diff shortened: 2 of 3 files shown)


tothambrus11:
@copilot Can you read the CONVENTIONS.md when the insructions.md points you to it during reviews? (even when it's not part of the diff)

CONVENTIONS.md
```diff
@@ -0,0 +1,98 @@
+# Coding Guidelines
+
+## Documentation
+
+Documentation enables local reasoning - it's a shortcut for understanding so readers can avoid looking up implementation or usages to infer meaning.
```
dabrahams: This should start with the documentation guidance given in the better code contracts and errors chapters, and only add stuff that departs from the book.

CONVENTIONS.md
```diff
@@ -0,0 +1,98 @@
+# Coding Guidelines
```
dabrahams, line 1: This should start with the Swift API Guidelines as a basis, noting that we're documenting departures from those guidelines.

CONTRIBUTING.md
```diff
@@ -3,4 +3,21 @@ Release drafts are created whenever pushing to a tag in the repository. You can
...
 The conventional tag name for releases is `vX.Y.Z`, with an optional suffix for pre-releases (like `v0.0.2-test-lsp`).
-The release workflow only triggers for a pattern matching `v*`.
\ No newline at end of file
+The release workflow only triggers for a pattern matching `v*`.
+
+## Coding Conventions
+Consult [CONVENTIONS.md](CONVENTIONS.md) for coding conventions used in this project.
+
+## LLM usage
+If you use an LLM to generate code, please disclose this in your Pull Request description.
```
dabrahams, line 12: IIUC Dimi is absolutely opposed to having any such code in the project, and I feel almost as strongly about it.  This guidance is not consistent with that.

CONVENTIONS.md
```diff
@@ -0,0 +1,98 @@
...
+Documentation enables local reasoning - it's a shortcut for understanding so readers can avoid looking up implementation or usages to infer meaning.
+
+- Document every declaration with at least a summary sentence, using `///` style comments.
+- Start with a summary sentence fragment describing what it *does* (functions) or what it *is* (types, properties). End with a period.
+- Omit needless words: don't repeat the receiver's type, don't write `the`, `given`, `of self`, `of the current object` when context makes these obvious.
+- Use `iff` instead of `if` where applicable.
+- Use `<...>, if any.` for optional values where the absence reason is obvious. Otherwise: `<...> if <condition>, nil otherwise.`
+- Document preconditions with `- Requires:`. If the precondition is obvious from the summary, it need not be repeated.
+- Multiple preconditions should be in a markdown list below `- Requires:`.
+- Document performance of every operation that doesn't run in constant time and space.
```
dabrahams: I think we're probably going to be a little loose about that.  An operation that “increments every element of the array” is somewhat obviously O(N).  But I'm happy to start with this rule and see what can't reasonably be done that way.

CONVENTIONS.md
```diff
@@ -0,0 +1,98 @@
...
+- Conformance implementations are exempt from documentation when nothing useful can be added beyond the protocol requirement itself.
+
+## Contracts
+
+- Create the strictest contracts possible, so long as the client can reason about the preconditions locally.
```
dabrahams: I don't understand what the part after the comma means.

tothambrus11: Sometimes we have preconditions about an object's internal state which is infeasible to reason about locally. See this API with too strong preconditions:

```swift
struct LLVMModule {
   ...
   
   /// Declares a function with `name` and `type`.
   ///
   /// - Requires: function with `name` doesn't exist yet.
   func declareFunction(name: String, type: LLVMType) -> LLVMFunction

   /// Returns the function named `name`, if any.
   func function(named name: String) -> LLVMFunction?

}
```

When there are multiple sources that can lead to the creation of LLVM functions, it's burdensome to use the above API, as we usually don't have local information about the preconditions, so we have to perform a defensive check outside the call:

```swift
let f: LLVMFunction
if let g = m.function(named: name) {
  f = g
} else {
   f  = declareFunction(named: name, type: .init(from: [], to: .int))
}
```

When the majority of use sites of a function look like the above, I would argue the API is hard to use correctly. If the function with strong precondition + the defensive checking is an efficient basis, we may leave them in the API (maybe adding a longer name for the one with overly strong preconditions, or making it private), and just add a third convenience function that does the single right thing. Otherwise, we could write a more efficient version by using the internals of the type, and we could remove the overly strict API if it's never used.

```swift
/// Declares a function with `name` and `type`, if not yet declared.
///
/// - Requires: If the function with `name` is already declared, it must have the same type as `type`. 
func demandFunction(named name: String, type: LLVMType) -> LLVMFunction
```

Do you agree with this recommendation? If yes, do you see a better way of phrasing it?

tothambrus11: @dabrahams ping

dabrahams: > When the majority of use sites of a function look like the above, I would argue the API is hard to use correctly.

When the majority of use sites look like the above, someone should make a function that encapsulates the pattern.

Also, that hardness is dependent on a particular usage pattern.  The same could be said of any precondition that depends on the state of an object.  Array indexing requires that the element exists.  You can easily create usage patterns where that is not known ahead of time. 

If we wanted to keep this caveat, you would need to show examples, because the statement is not sufficient for a reader to understand what you're talking about.  But I don't think we want to keep this caveat.

The better code book states that a precondition must be something that is practically possible and efficient for a client to ensure, and that ensuring it _often_ ends up being by construction.  If you pick a usage pattern that doesn't ensure it by construction, that's on you.

If the API author can reasonably predict that a check will be needed every time, well, it's not just a problem of preconditions; they have likely designed the wrong API.

dabrahams:
Instead of writing a bunch of stuff on your own, how about you start with the Swift API Guidelines and the book, and raise for discussion in slack anything you think needs to be amended or added?  Most of what you've written seems to be redundant with that material, and could be omitted, but picking it apart to find any small differences is not productive.

CONVENTIONS.md
tothambrus11: > You can easily create usage patterns where that is not known ahead of time.

Yeah, that's a fair point.

> a precondition must be something that is practically possible and efficient for a client to ensure, and that ensuring it often ends up being by construction. ... If the API author can reasonably predict that a check will be needed every time, well, it's not just a problem of preconditions; they have likely designed the wrong API.

That's a really nice way of phrasing this. Upon a quick search, I didn't find this in the book, maybe it should be written there explicitly. I would also add this as a point in our guidelines instead of my phrasing.

dabrahams: This is just a specific case of “think about how your API will be used,” which is applicable to all API design and I'm not sure it deserves special mention.  It is certainly a theme throughout the API guidelines.

PR #106, opened 2026-04-18
Reserve memory for the type store to amortize reallocation costs

kyouko-taiga:
Profiling suggests that the compiler spends a lot of time reallocating the internal storage of the type store. This PR attempts to fix this problem.

### DIFF
--- a/Sources/FrontEnd/Types/TypeStore.swift
+++ b/Sources/FrontEnd/Types/TypeStore.swift
@@ -30,6 +30,16 @@ public struct TypeStore: Sendable {
     self.nextFreshIdentifier = 0
   }
 
+  /// Returns a number less than the number of types created with this store.
+  public var underestimatedCount: Int {
+    types.count
+  }
+
+  /// Reserves enough space to store `minimumCapacity` types without allocating new storage.
+  public mutating func reserveCapacity(_ minimumCapacity: Int) {
+    self.types.reserveCapacity(minimumCapacity)
+  }
+
   /// Returns the identity of a fresh type variable.
   public mutating func fresh() -> TypeVariable.ID {
     defer { nextFreshIdentifier += 1 }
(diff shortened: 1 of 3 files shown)


Sources/FrontEnd/Types/TypeStore.swift
```diff
@@ -30,6 +30,16 @@ public struct TypeStore: Sendable {
     self.nextFreshIdentifier = 0
   }
 
+  /// Returns a number less than the number of types created with this store.
```
tothambrus11, line 33:
```suggestion
  /// Returns a number at most the number of types created with this store.
```
Is equality possible?

kyouko-taiga:
```suggestion
  /// Returns a number less than or equal to the number of types created with this store.
```

tothambrus11:
I tried reproducing the parallel speedup locally, it seems my machine has very different characteristics than the MacBook, so I didn't see significant improvements.

There were great improvements on sequential builds though.

image

The CI also confirmed that this is generally an improvement in sequential builds, though it only showed 10-20% improvement. Github CI runners are incapable of proper parallel execution, passing `--parallel` just made things 5-10x slower.

kyouko-taiga:
The 5x I observed on my machine are for debug builds, which probably explains why we can't reproduce such a dramatic improvement on the CI.

I think it is still relevant to speedup debug builds, though, because we typically run the tests quite often during development. Maybe something to consider if we want to start collecting performance measurement on nightlies or something like that.

Otherwise, do you think there's a good reason to keep this PR waiting? I wouldn't mind merging it, even if it only improves the developer/debugger experience.

tothambrus11:
Now I measured 1.74x speedup in sequential and 1.07x speedup in parallel, in debug builds on my local ubuntu machine. That's clearly a nice win, so let's merge. :D

PR #116, opened 2026-04-26
Add floating point literals and machine types

### LINKED ISSUE #114: Add float support
- float parsing
- float machine types
- ExpressibleByFloatingPointLiteral

tothambrus11:
This PR adds Float32, Float64 and the necessary conversions, similarly to how it's done for integer literals. A current limitation is that we can't use an integer literal for floating point types, we must explicitly add the `.0` suffix.

LLVM code generation is already implemented and tested in the #53 branch.

The floating point literal values are stored in a string after lexing and parsing (within the FloatLiteral expression), then during IR lowering, we strip all `_` symbols so the literal becomes parsable by standard float parsers. In the LLVM lowering branch, I used this string value and parsed by the corresponding LLVM floating point type's constant creation function that parses the value straight to the destination type. This may be slightly fragile, as we depend on LLVM float parsing to match our expectations, but we can save ourselves from an extra BigDecimal library. We can also mitigate unexpected parsing divergences by writing some integration tests exercising the edge cases of float syntax.

Some of the floating point builtin cases are not yet covered by tests because they are not yet used in the standard library, and I don't think it's worth testing the fact that lowering them succeeds without a crash. LMK if you disagree, I can add these tests (just calling `Builtin.f*` and ignoring the result).

Closes #114 and #53 .

For the lucky reviewer: https://open.spotify.com/track/0r8SR1mxjyN42Kz8NfQJl9?si=fe2fcc8f3ee048aa

### DIFF
--- a/Sources/FrontEnd/IR/IREmitter.swift
+++ b/Sources/FrontEnd/IR/IREmitter.swift
@@ -558,6 +558,8 @@ internal struct IREmitter {
       lower(store: program.castUnchecked(e, to: InoutExpression.self), to: target)
     case IntegerLiteral.self:
       lower(store: program.castUnchecked(e, to: IntegerLiteral.self), to: target)
+    case FloatingPointLiteral.self:
+      lower(store: program.castUnchecked(e, to: FloatingPointLiteral.self), to: target)
     case NameExpression.self:
       lower(store: program.castUnchecked(e, to: NameExpression.self), to: target)
     case SyntheticExpression.self:
@@ -683,6 +685,11 @@ internal struct IREmitter {
 
   /// Implements `lower(store:to:)` for integer literals.
   private mutating func lower(store e: IntegerLiteral.ID, to target: IRValue) {
+    unreachable()
+  }
+  
+  /// Implements `lower(store:to:)` for integer literals.
+  private mutating func lower(store e: FloatingPointLiteral.ID, to target: IRValue) {
     unreachable()
   }
 
@@ -1116,9 +1123,12 @@ internal struct IREmitter {
     case .expressibleByIntegerLiteralInit:
       let s = program.cast(source.value, to: IntegerLiteral.self)!
       return loweredBuiltinIntegerLiteralConversion(from: s, to: target)
+    case .expressibleByFloatingPointLiteralInit:
+      let s = program.cast(source.value, to: FloatingPointLiteral.self)!
+      return loweredBuiltinFloatingPointLiteralConversion(from: s, to: target)
 
     default:
-      program.unexpected(e)
+      unreachable("unexpected call to '\(program.show(e))' applied to '\(f)'")
     }
   }
 
@@ -1141,7 +1151,28 @@ internal struct IREmitter {
     case program.standardLibraryType(.int64):
       return .integer(value, program.types.demand(MachineType.i(64)))
     default:
-      program.unexpected(source)
+      program.unexpected(target)
+    }
+  }
+
+  /// Returns the value denoted by `source` interpreted as the floating point
+  /// type `target`, defined in the standard library.
+  ///
+  /// Standard library floating point types are thin wrappers around a machine type. For instance, `Float32` wraps
+  /// a single `Builtin.float32` property. This method returns the value of that property converted from
+  /// a floating point number literal.
+  private mutating func loweredBuiltinFloatingPointLiteralConversion(
+    from source: FloatingPointLiteral.ID, to target: AnyTypeIdentity
+  ) -> IRValue {
+    let value = program[source].value.sans("_")
+
+    switch target {
+    case program.standardLibraryType(.float32):
+      return .floatingPoint(literal: value, program.types.demand(.float32))
+    case program.standardLibraryType(.float64):
+      return .floatingPoint(literal: value, program.types.demand(.float64))
+    default:
+      program.unexpected(target)
     }
   }
 
--- a/Sources/FrontEnd/IR/IRValue.swift
+++ b/Sources/FrontEnd/IR/IRValue.swift
@@ -12,6 +12,12 @@ public enum IRValue: Hashable, Sendable {
   /// A constant integer.
   indirect case integer(BigInt, MachineType.ID)
 
+  /// A constant floating point number.
+  ///
+  /// `literal` contains the string representation of the floating point number
+  /// in the format `[+-]? [0-9]+ "." [0-9]* ([eE] [+-]? [0-9]+)?`.
+  indirect case floatingPoint(literal: String, MachineType.ID)
+
   /// A reference to a lowered function.
   indirect case function(IRFunction.Name, AnyTypeIdentity)
 
@@ -70,6 +76,8 @@ extension IRValue: Showable {
     case .register(let i):
       return "%r\(i.address.rawValue)"
     case .integer(let n, let t):
+      return "\(printer.show(t)) \(n)"
+    case .floatingPoint(let n, let t):
       return "\(printer.show(t)) \(n)"
     case .function(let n, _):
       return printer.show(n)
--- a/Sources/FrontEnd/Syntax/Builtins/BuiltinFunction.swift
+++ b/Sources/FrontEnd/Syntax/Builtins/BuiltinFunction.swift
@@ -1105,7 +1105,7 @@ extension BuiltinFunction {
       self = .markUninitialized
 
     //    case "advanced":
-    //      guard let ((_, _), t) = (exactly("by") ++ exactly("bytes") ++ machineType)(&tokens)
+    //      guard let ((_, _), t) = (exactly("by") + exactly("bytes") + machineType)(&tokens)
     //      else { return nil }
     //      self = .advancedByBytes(byteOffset: t)
 
@@ -1124,19 +1124,19 @@ extension BuiltinFunction {
     //      self = .shl(p, t)
     //
     //    case "udiv":
-    //      guard let (p, t) = (maybe("exact") ++ machineType)(&tokens) else { return nil }
+    //      guard let (p, t) = (maybe("exact") + machineType)(&tokens) else { return nil }
     //      self = .udiv(exact: p != nil, t)
     //
     //    case "sdiv":
-    //      guard let (p, t) = (maybe("exact") ++ machineType)(&tokens) else { return nil }
+    //      guard let (p, t) = (maybe("exact") + machineType)(&tokens) else { return nil }
     //      self = .sdiv(exact: p != nil, t)
     //
     //    case "lshr":
-    //      guard let (p, t) = (maybe("exact") ++ machineType)(&tokens) else { return nil }
+    //      guard let (p, t) = (maybe("exact") + machineType)(&tokens) else { return nil }
     //      self = .lshr(exact: p != nil, t)
     //
     //    case "ashr":
-    //      guard let (p, t) = (maybe("exact") ++ machineType)(&tokens) else { return nil }
+    //      guard let (p, t) = (maybe("exact") + machineType)(&tokens) else { return nil }
     //      self = .ashr(exact: p != nil, t)
     //
     //    case "urem":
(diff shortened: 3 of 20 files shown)


Sources/FrontEnd/IR/IREmitter.swift
```diff
@@ -1116,9 +1123,11 @@ internal struct IREmitter {
     case .expressibleByIntegerLiteralInit:
       let s = program.cast(source.value, to: IntegerLiteral.self)!
       return loweredBuiltinIntegerLiteralConversion(from: s, to: target)
-
+    case .expressibleByFloatingPointLiteralInit:
+      let s = program.cast(source.value, to: FloatingPointLiteral.self)!
+      return loweredBuiltinFloatingPointLiteralConversion(from: s, to: target)
```
kyouko-taiga, line 1128:
```suggestion
      return loweredBuiltinFloatingPointLiteralConversion(from: s, to: target)

```

Sources/FrontEnd/IR/IREmitter.swift
```diff
@@ -1116,9 +1123,11 @@ internal struct IREmitter {
...
     default:
-      program.unexpected(e)
+      unreachable("Call: \(program.show(e)), applied with: \(f)")
```
kyouko-taiga: What's wrong with `unexpected`?

```suggestion
      unreachable("unexpected call to '\(program.show(e))' applied to '\(f)'")
```

Sources/FrontEnd/IR/IREmitter.swift
```diff
@@ -1141,7 +1150,28 @@ internal struct IREmitter {
     case program.standardLibraryType(.int64):
       return .integer(value, program.types.demand(MachineType.i(64)))
     default:
-      program.unexpected(source)
+      program.unexpected(target)
+    }
+  }
+
+  /// Returns the value denoted by `source` interpreted the contents of the floating point type `target`,
+  /// defined in the standard library.
```
kyouko-taiga:
```suggestion
  /// Returns the value denoted by `source` interpreted as the integer type `target`, defined in
  /// the standard library.
```

Sources/FrontEnd/IR/IRValue.swift
```diff
@@ -12,6 +12,11 @@ public enum IRValue: Hashable, Sendable {
   /// A constant integer.
   indirect case integer(BigInt, MachineType.ID)
 
+  /// A constant floating point number.
+  ///
+  /// `literal` contains the string representation of the floating point number without underscores.
+  indirect case floatingPoint(literal: String, MachineType.ID)
```
kyouko-taiga, line 19:
```suggestion
  /// A constant floating point number.
  indirect case floatingPoint(String, MachineType.ID)
```

It's pretty common to store the values of floating point numbers as character strings. No one would be surprised. The comment about the underscores is also not really necessary as one does not care how the literal was expressed in the first place. We don't preserve this formatting information for integers either.

Sources/FrontEnd/IR/IRValue.swift
```diff
@@ -71,6 +76,8 @@ extension IRValue: Showable {
       return "%r\(i.address.rawValue)"
     case .integer(let n, let t):
       return "\(printer.show(t)) \(n)"
+    case .floatingPoint(let f, let t):
+      return "\(printer.show(t)) \(f)"
```
kyouko-taiga:
```suggestion
    case .floatingPoint(let n, let t):
      return "\(printer.show(t)) \(n)"
```

Sources/FrontEnd/Syntax/Builtins/BuiltinFunction.swift
```diff
@@ -1095,7 +1105,7 @@ extension BuiltinFunction {
       self = .markUninitialized
 
     //    case "advanced":
-    //      guard let ((_, _), t) = (exactly("by") ++ exactly("bytes") ++ machineType)(&tokens)
+    //      guard let ((_, _), t) = (exactly("by") + exactly("bytes") + machineType)(&tokens)
```
kyouko-taiga, line 1108: Is that a preemptive change? If so I would not make it part of this patch.

Sources/FrontEnd/Syntax/Expressions/FloatingPointLiteral.swift
```diff
@@ -0,0 +1,27 @@
+import Archivist
+
+/// The expression of a floating point number literal.
+@Archivable
+public struct FloatingPointLiteral: Expression {
```
kyouko-taiga: The name of the file doesn't match. Is there a good reason? Otherwise we should maintain the convention.

Sources/FrontEnd/Syntax/SyntaxVisitor.swift
```diff
@@ -94,7 +94,7 @@ extension Program {
       break
     case InoutExpression.self:
       traverse(castUnchecked(n, to: InoutExpression.self), calling: &v)
-    case IntegerLiteral.self:
+    case IntegerLiteral.self, FloatingPointLiteral.self:
```
kyouko-taiga:
```suggestion
    case IntegerLiteral.self
      break
    case FloatingPointLiteral.self:
```
I avoid matching with more than one pattern in general.

Sources/FrontEnd/Typer/Typer.swift
```diff
@@ -2267,6 +2271,45 @@ public struct Typer {
     }
   }
 
+  /// Returns the inferred type of `e`.
+  private mutating func inferredType(
```
kyouko-taiga: This method is very, very close to the one for integer literals. That triggers my DRY response. I think we can abstract over the two implementations.

Let me know if you need help doing that.

Sources/FrontEnd/Typer/Typer.swift
```diff
@@ -4601,6 +4644,21 @@ public struct Typer {
     }
   }
 
+  /// Returns `true` iff `t` is a standard library floating point type (e.g., `Hylo.Float32`).
+  ///
+  /// The module containing the standard library must have been loaded in the `self.program`, or
+  /// `self.module` is the standard library.
+  private mutating func isStandardLibraryFloatingPointType(_ t: AnyTypeIdentity) -> Bool {
+    guard program.containsStandardLibrary else { return false }
+    switch program.types.dealiased(t) {
```
kyouko-taiga, line 4617: Nothing wrong with the functionality but I find this code "ugly" (and the one I write for integers too).

What if we tried something like this:

```swift
private mutating func isStandardLibraryFloatingPointType(_ t: AnyTypeIdentity) -> Bool {
  isStandardLibraryType(t, in: [.float32, .float64])
}

private mutating func isStandardLibraryType(
  _ t: AnyTypeIdentity, in ts: [Program.StandardLibraryEntity]
) -> Bool {
  guard program.containsStandardLibrary else { return false }
  let u = program.types.dealiased(t)
  return ts.contains(where: { (e) in standardLibraryType(e) == u  })
}
```

I didn't check but I strongly suspect the loop will get unrolled and inlined. The array will get statically allocated, if not completely inlined after unrolling.

IIRC you mentioned that this method is expensive elsewhere and I agree. I think we can cache the result of `standardLibraryType(_:)` in the typer. But maybe you'd like to do that as a separate patch.

kyouko-taiga (changes requested):
This PR looks very, very good! I enjoyed reviewing it.

I didn't carefully reviewed the definitions of `Float[32|64]` but I suppose they were pretty much copied and adapted from `Int`.

My change request is about the code duplication in typer. Let me know if you need help to write something more generic.

Sources/FrontEnd/IR/IREmitter.swift
tothambrus11: It didn't report enough useful context on the crash, IIRC it only printed "Call".

Sources/FrontEnd/IR/IRValue.swift
tothambrus11: It's kind of important to know what format they are in because we plug it into LLVM's `llvmType.constant(parsing:)` function, which doesn't work with underscores. If it wasn't LLVM but some generic float parsing library, I would even question if it wants `,` or `.` for the decimal separator, which is sometimes locale-dependent.

Sources/FrontEnd/Syntax/Builtins/BuiltinFunction.swift
tothambrus11: I added these in a separate commit (so that the next person won't need to go on a trip to look for the definition of `++` in the old compiler).

Sources/FrontEnd/Syntax/SyntaxVisitor.swift
tothambrus11: What's the motivation? It's more convenient to write this if the bodies are the same.

kyouko-taiga: It's narrows the context in diffs and it makes the switch very regular. There is one case per thing that is handled, in alphabetical order. Where canI find `IntegerLiteral`? Exactly where one can predict it will be.

Sources/FrontEnd/IR/IRValue.swift
kyouko-taiga: But almost no one stores underscores and I would be kind of shocked if someone found it weird that they are gone by the time we get to IR.

kyouko-taiga: If you really care about describing the format, then you should say way much more than "a string representation without underscores". `pi` and `quarterOfOne` satisfy your specification.

Sources/FrontEnd/IR/IREmitter.swift
kyouko-taiga: I don't think we should start getting into the business of describing very precisely what is unexpected. The whole point of `unexpected` is to have one uniform way of marking paths that do not handle some specific nodes in this syntax tree, so we don't have to bike shed strings that are supposed to never show up. You get a line number and a tag. That's plenty enough to go to the actual source and look for more context.

Add more precise messages only if there is something subtle that you want the next person to know if they get there. That's not the case here. And even if it was, I would add a comment on top `unexpected`. If anyone triggers the call, someone will have to look at the code anyway.

tothambrus11: In general, my goal when writing assertions is to dump enough context so the programmer doesn't need to recompile or rerun the process with a debugger attached to at least get a rough idea of how the condition failed.

I would still look at the code, but wouldn't need to rerun it or evaluate expressions in the debug repl. This can be helpful especially when running the compiler via `swift run hc ...`, where we don't have a debugger attached. I think `unexcpected` would have been given enough information, the main problem was that we didn't pass the switched-on value to it, but something that can be the same regardless whether it's a valid or invalid call. `f` is not a Syntax but StandardLibraryEntity, so I need to use the raw `unreachable`.

Sources/FrontEnd/Typer/Typer.swift
tothambrus11: Thanks, I'll try that. I suggest keeping the ugly version for this PR, and we can experiment with https://github.com/hylo-lang/hylo-new/issues/118 separately.

Sources/FrontEnd/Typer/Typer.swift
tothambrus11: I added a new protocol `LiteralExpression` that abstracts over the differences using associated value requirements.

Sources/FrontEnd/Syntax/Expressions/LiteralExpression.swift
```diff
@@ -0,0 +1,13 @@
+/// A literal expression.
+public protocol LiteralExpression: Expression {
```
kyouko-taiga, line 2:
```suggestion
public protocol ScalarLiteralExpression: Expression {
```
I don't think array and map literals will be able to conform to this protocol.

Sources/FrontEnd/Syntax/Expressions/IntegerLiteral.swift
```diff
@@ -17,6 +17,16 @@ public struct IntegerLiteral: Expression {
 
 }
 
+extension IntegerLiteral: LiteralExpression {
```
kyouko-taiga: I would probably declare conformance in the primary declaration. I'm not a fan of refining conformances in extensions, especially in cases where it feels intuitively like the type is implementing an interface, in the sense of traditional object-orientation.

tothambrus11:
> I didn't carefully reviewed the definitions of Float[32|64] but I suppose they were pretty much copied and adapted from Int.

That's exactly what happened. We can create a more fully featured API when the generator is done.

Sources/FrontEnd/Syntax/Expressions/IntegerLiteral.swift
```diff
@@ -17,6 +17,16 @@ public struct IntegerLiteral: Expression {
...
+
+  public static let kind: LiteralType = .integer
```
kyouko-taiga: The term `kind` is reserved to mean the kind of a type. We cannot use it in any other way in the entire codebase.

Sources/FrontEnd/Syntax/Expressions/LiteralExpression.swift
tothambrus11: Thanks, indeed. Now I considered two alternatives: either
 - make a protocol `PrimtiveLiteralExpression: LiteralExpression`, or
 - only keep the `literalType` requirement in `LiteralExpression` protocol, and require the other values as a parameter in Typer's function.

I chose the latter, as I'm not yet sure what's the best taxonomy and hierarchy of future types, so  introducing extra abstractions could lead to overfitting. It's also the only place where these values would be likely used, so maybe it's better to define the values inline rather than at a distant place.

If you have more confifent/different predictions, LMK.

kyouko-taiga: Option 2 sounds good to me.

PR #122, opened 2026-04-28
Fix illegal access in self-reference in assignments (#121)

### LINKED ISSUE #121: illegal access in self-reference in assignment: `&x = x - 1`
This function currently fails access checking:
```hylo
fun f() {
    var x = 10
    &x = x - 1
}
```

```
loop.hylo:4.9-15: error: illegal 'sink' access
    &x = x - 1
        ~~~~~~
loop.hylo:4.9-15: error: illegal 'set' access
    &x = x - 1
        ~~~~~~
```

However, the issue is solved when adding `0` to the rhs expression:
```
fun f() {
    var x = 10

    &x = (x - 1) + 0
}
```

tothambrus11:
Closes #121

### DIFF
--- a/Sources/FrontEnd/Program.swift
+++ b/Sources/FrontEnd/Program.swift
@@ -707,13 +707,12 @@ public struct Program: Sendable {
     self[n.module].declaration(referredToBy: n) ?? unreachable("untyped node at \(self[n].site)")
   }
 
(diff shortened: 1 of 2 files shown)


kyouko-taiga:
I have a competing PR that improves the efficiency of `Program.occurs(_:in:)` and also tests for cases where the LHS of the assignment is a more complex expression.

I'm sorry for the double work, but I think we should close this PR.

PR #127, opened 2026-04-29
ci: add dependabot config for github-actions updates

### LINKED ISSUE #97: Add dependabot
We should add dependabot so that we can upgrade our dependencies easily. See the old compiler repository for reference: https://github.com/hylo-lang/hylo/blob/main/.github/dependabot.yml

mvanhorn:
Resolves #97.

Mirror the existing dependabot config from the legacy [hylo-lang/hylo](https://github.com/hylo-lang/hylo/blob/main/.github/dependabot.yml) repo (linked in the issue) so the GitHub Actions versions pinned in `.github/workflows/` stay current.

```yaml
version: 2
updates:
  - package-ecosystem: "github-actions"
    directory: "/"
    schedule:
      interval: "weekly"
```

Weekly cadence matches the legacy repo. The repo currently has three workflows (`build-and-test`, `release`, `spell-check`) that pin actions like `actions/checkout@v4` — dependabot will open PRs as those publish new versions.

Swift Package Manager dependencies live in `Package.swift` / `Package.resolved`, but dependabot's swift-package-manager support is still preview-only and the legacy repo did not enable it either. Keeping this config narrow to `github-actions` for parity; SPM updates can be added in a follow-up if the team wants them.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,6 @@
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "weekly"


tothambrus11:
Thanks for looking into this! I think we are fine without always updating our SPM dependencies. It's useful for all those javascript based tools that turn out to contain vulnerabilities every other day. Speaking of which, we may want to look into our set of dependencies (after we ship 1.0 and have real users), so that we can minimize the risk of a supply chain attack. Odin's creator likes to read the source code of all of his dependencies, I hope we won't need to get that far...

PR #131, opened 2026-04-29
Fix the typing of memberwise initializers

kyouko-taiga:
This patch fixes the typing of memberwise initializers for structs having a binding declaration that involves more than one variable.


tothambrus11:
Thanks, LGTM!

Issue #117, opened 2026-04-27
Add while statement


tothambrus11:
We need loops for practical programming with limited stack space.

kyouko-taiga:
You probably don't need to justify the need for `while` loops ;)

PR #132, opened 2026-04-30
Add while loop and relax IR invariants

tothambrus11:
This PR adds:
- parsing, type checking and lowering of while statement #117 
- relaxing the invariants of IRAlloca declarartion in the abstract interpreter

The static allocas are already hoisted during LLVM lowering, but not the dynamic ones (e.g. for stack-allocating generic return value).

Please review carefully especially the IREmitter and abstract context changes. I had to change many usages of AbstractContext's `declare` to wire a `ControlFlowGraph` to `declare`. I saw a similar pattern already for passing in `Typer!`, but it still feels a bit too intrusive, so let me know if you have a better idea.

I added a check that repeated allocations for the same instructions must only happen if the instruction is in a loop (using the control flow graph). I couldn't yet add the check that would ensure that upon reentering the same allocation, the value at that place should have been already deinitialized. IIUC this check needs to reside in lifetime normalization, but `declare` is used there in a number of places, so I was not sure where to add this check. I'm also not confident at upholding the invariants of lifetime normalization/abstract interpretation yet, so I think I shouldn't attempt a potential refactoring / changing of invariants myself. But LMK if there would be a straightforward path, I can try then. We could also merge the PR without this extra guardrail, and add it later.

### DIFF
--- a/Sources/FrontEnd/IR/AbstractInterpreter/AbstractContext.swift
+++ b/Sources/FrontEnd/IR/AbstractInterpreter/AbstractContext.swift
@@ -171,10 +171,15 @@ internal struct AbstractContext<Domain: AbstractDomain>: Hashable, Sendable {
   /// to contain the new object and the register is assigned to that place.
   ///
   /// The new object is defined as a uniform value `v`.
+  /// `g` is the control-flow graph of `f`.
   internal mutating func declare<T: InstructionIdentity>(
-    _ i: T, from f: IRFunction, initially v: Domain
+    _ i: T, from f: IRFunction, controlFlow g: ControlFlowGraph, initially v: Domain
   ) {
-    assert(locals[.register(i.erased)] == nil, "register is already assigned")
+    guard locals[.register(i.erased)] == nil else {
+      assert(isInCycle(i, in: f, g), 
+        "Register \(i) is already defined, which is only acceptable in a loop.")
+      return // Don't redefine if it's already there.
+    }
 
     // Create a new object.
     let t = f.resolved(f.at(i.erased).type)!
@@ -182,7 +187,6 @@ internal struct AbstractContext<Domain: AbstractDomain>: Hashable, Sendable {
 
     // If the register defines an address, create a new place and assigns it the new object.
     if t.isPlace {
-      assert(memory[.register(i.erased)] == nil, "storage already exists")
       memory[.register(i.erased)] = .init(type: t.type, value: .uniform(v))
       locals[.register(i.erased)] = .place(.root(.register(i.erased)))
     }
@@ -259,4 +263,12 @@ extension AbstractContext.Locals: Showable {
     }
   }
 
+}
+
+/// Returns `true` iff `i` is defined in a block that is its own (transitive) predecessor.
+private func isInCycle(
+  _ i: some InstructionIdentity, in f: IRFunction, _ g: ControlFlowGraph
+) -> Bool {
+  let b = f.block(defining: i)
+  return g.predecessors(of: b).contains(b)
 }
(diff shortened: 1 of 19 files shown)


Sources/FrontEnd/IR/IREmitter.swift
```diff
@@ -536,6 +538,32 @@ internal struct IREmitter {
...
+    insertionContext.point = .end(of: head)
+
+    // Lower the conditions.
+    for c in program[s].conditions {
+      insertionContext.point = .end(of: 
+        lowerCondition(c, onFailure: tail))
+    }
+    
+    // Lower the body.
+    if lower(program[s].body) == .next{
```
dabrahams:
```suggestion
    if lower(program[s].body) == .next {
```

Sources/FrontEnd/IR/IREmitter.swift
```diff
@@ -536,6 +538,32 @@ internal struct IREmitter {
     return .return(s)
   }
 
+  /// Generates the IR of `s`.
+  private mutating func lower(_ s: While.ID) -> ControlFlow {
+    
+    let head = insertionContext.function!.addBlock()
```
tothambrus11:
```suggestion
    let conditionsStart = insertionContext.function!.addBlock()
```

Sources/FrontEnd/IR/IREmitter.swift
```diff
@@ -536,6 +538,32 @@ internal struct IREmitter {
...
+    let tail = insertionContext.function!.addBlock()
```
tothambrus11:
```suggestion
    let afterLoop = insertionContext.function!.addBlock()
```

Sources/FrontEnd/IR/Transforms/IRFunction+StaticAllocationHoisting.swift
```diff
@@ -0,0 +1,53 @@
+import Utilities
+
+extension IRFunction {
+
+  /// Moves all static stack allocations to the entry block, updating all usages.
+  internal mutating func hoistStackAllocationsToEntryBlock() {
+    guard let entry else { return }
```
tothambrus11:
```suggestion
    guard let e = entry else { return }

File: Sources/FrontEnd/IR/Transforms/IRFunction+StaticAllocationHoisting.swift [hunk 5]
```diff
@@ -0,0 +1,53 @@
...
+
+    let n = instructions().count
```
tothambrus11:
```suggestion
    let initialInstructionCount = instructions().count
```

Sources/FrontEnd/IR/AbstractInterpreter/AbstractContext.swift
```diff
@@ -171,18 +171,22 @@ internal struct AbstractContext<Domain: AbstractDomain>: Hashable, Sendable {
...
   ///
   /// The new object is defined as a uniform value `v`.
+  /// `g` is the control-flow graph of `f`.
   internal mutating func declare<T: InstructionIdentity>(
-    _ i: T, from f: IRFunction, initially v: Domain
+    _ i: T, from f: IRFunction, controlFlow g: ControlFlowGraph, initially v: Domain
   ) {
-    assert(locals[.register(i.erased)] == nil, "register is already assigned")
+    guard locals[.register(i.erased)] == nil else {
+      assert(isInCycle(i, in: f, g),
```
kyouko-taiga, line 179: I think this test is worsening the API of the function too much. The fact that redefinitions can only happen in cycles should be a wellformedness check that doesn't have to run in the middle of hot passes like lifetime normalization.

kyouko-taiga:
Superseded by #293

Issue #126, opened 2026-04-28
Unnecessary temporary created on non-overlapping assignment


tothambrus11:
```hylo
//! stage:lowering

/// Regression test for https://github.com/hylo-lang/hylo-new/issues/126.
///
/// Assigning a value to a stored property of `self` should build the value in place rather than
/// storing it into a temporary and moving it. In particular, the type of the assigned value need
/// not conform to `Movable`, as demonstrated here by omitting a `Movable` witness for `Wrapper`.

struct Wrapper is Deinitializable {
  var x: Int
  var y: Int
  var z: Int

  public init(x: Int, y: Int, z: Int) {
    &self.x = x + 0
    &self.y = y + 0
    &self.z = z + 0
  }

  public fun manualCopy() -> Wrapper {
    return Wrapper(
      self.x + 0,
      self.y + 0,
      self.z + 0)
  }
}

struct Box is Deinitializable {
  public var payload: Wrapper

  /// The RHS is an in-place constructor call that does not mention `self`.
  public init() {
    &self.payload = Wrapper(0, 0, 0)
  }

  /// The RHS is a non-constructor call whose arguments do not mention `self`.
  public init(_ payload: Wrapper) {
    &self.payload = payload.manualCopy()
  }
}

```
In both initializers, the compiler complains that `Wrapper` is not `Movable`, indicating that it emitted a temporary before the assignment. This temporary is created because of a too conservative check to see if the left and right hand side of an assignment overlap.

kyouko-taiga:
`Copyable` is a refinement of `Movable` so you'll beed a conformance to the latter anyway.

tothambrus11:
Here is a partial fix, which is already less conservative, but not as liberal as we would want it still: https://github.com/hylo-lang/hylo-new/pull/136

kyouko-taiga:
An assignment is not an initialization. The semantics of an assignment is precisely implemented by `Movable` because`&lhs = rhs` desugars (at least notionally) to `&lhs.take_value(from: rhs)`, where `take_value(from:)` is a requirement of `Movable`.

We reserve the right to optimize the move away but the semantics still mandates a conformance to `Movable`.

Initializing a variable at declaration is different. That is, `var x = A()` is **not** equivalent to `var x: A ; &x = A()`. Rather, it is `var x: A ; A.init(self: &x)`, which incidentally is the way one should initialize immovable things.

tothambrus11:
[...]

PR #136, opened 2026-05-01
Don't emit temporary upon assignment when lhs and rhs roots don't overlap

tothambrus11:
This is a partial solution to #126. It still doesn't account for the case when we know the memory layout of the left hand side, and we would be able to prove that the LHS and RHS don't overlap. 

E.g. `&self.x = self.y.copy()` doesn't work yet because the compiler has a too conservative understanding, and sees that `self`, the root of LHS is used by RHS.

Test case:

````
//! stage:lowering

struct Wrapper is Deinitializable {
  var x: Int
  var y: Int
  var z: Int

  public init(x: Int, y: Int, z: Int) {
    &self.x = x + 0
    &self.y = y + 0
    &self.z = z + 0
  }

  public fun manualCopy() -> Wrapper {
    return Wrapper(
      self.x + 0,
      self.y + 0,
      self.z + 0)
  }
}

struct Box is Deinitializable {
  public var w1: Wrapper
  public var w2: Wrapper

  /// The RHS is a non-constructor call whose arguments do not mention `self`.
  public init(_ payload: Wrapper) {
    &self.w1 = payload.manualCopy()
    &self.w2 = self.w1.manualCopy()
  }
}
```

### DIFF
--- a/Sources/FrontEnd/IR/IREmitter.swift
+++ b/Sources/FrontEnd/IR/IREmitter.swift
@@ -461,15 +461,12 @@ internal struct IREmitter {
       return .next
     }
 
-    // If the LHS does not occur in the RHS, we can build the RHS in place.
-    if let n = program.cast(target, to: NameExpression.self) {
-      if case .direct(let d) = program.declaration(referredToBy: n) {
-        if !program.occurs(referenceTo: d, in: program[s].rhs) {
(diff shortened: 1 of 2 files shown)


kyouko-taiga:
I don't think #126 is legit. This patch is surprising. We should discuss semantics before moving forward.

PR #144, opened 2026-05-03
Applies some optimizations on the typer

kyouko-taiga:


Sources/FrontEnd/Typer/Typer.swift
tothambrus11: Don't we want to rewrite this as well to a loop to avoid the extra  temporary array allocation for filter? Are lazy collections as good as the for loop?

Sources/FrontEnd/Typer/Typer.swift
```diff
@@ -3843,31 +3846,48 @@ public struct Typer {
...
-      // Is the given of the form `T ~ U`?
-      if let e = program.types.cast(u.head, to: EqualityWitness.self) {
-        let l = program.types.open(u.context.parameters, in: program.types[e].lhs)
-        let r = program.types.open(u.context.parameters, in: program.types[e].rhs)
+  /// Returns `(lhs, rhs)` where `lhs` (respectively `rhs`) is `true` iff a `g` might be used to
+  /// prove a coercion from  from `a` to `x` (respectively `x` to `b`) for some type `x`.
+  private mutating func canDeriveCoercion(
+    _ a: AnyTypeIdentity, _ b: AnyTypeIdentity, applying g: Given
+  ) -> (Bool, Bool) {
+    let p = (b.bits < a.bits) ? Pair(b, a) : Pair(a, b)
```
tothambrus11, line 3866: This is a bit tricky code. It could deserve some comment.

Sources/FrontEnd/Typer/Typer.swift
```diff
@@ -3843,31 +3846,48 @@ public struct Typer {
...
+      let l = program.types.open(u.context.parameters, in: program.types[e].lhs)
+      let r = program.types.open(u.context.parameters, in: program.types[e].rhs)
+
+      let lhs = unifiable(a, l) || unifiable(a, r)
+      let rhs = unifiable(b, l) || unifiable(b, r)
+      cache.canDeriveCoercion[g, default: [:]][p] = (lhs, rhs)
+      return (lhs, rhs)
+    }
+
+    // The given can't be use to form a coercion.
```
```suggestion
    // The given can't be used to form a coercion.
```

Sources/Utilities/Pair.swift
```diff
@@ -0,0 +1,22 @@
+/// A pair of elements.
+public struct Pair<T, U> {
```
tothambrus11, line 2: It's weird that Swift doesn't have a conditional Hashable conformance for tuples. Do you see any reason why they did that? IIUC adding post-hoc conformances to standard library types is not sound if we make libraries, so the best place for such conformance would be in the standard library.

Sources/FrontEnd/Typer/Typer.swift
```diff
@@ -3843,31 +3846,48 @@ public struct Typer {
...
+    assert(!lhs || !rhs)
+    return false
+  }
 
-      // Is the given of the form `T ~ U`?
-      if let e = program.types.cast(u.head, to: EqualityWitness.self) {
-        let l = program.types.open(u.context.parameters, in: program.types[e].lhs)
-        let r = program.types.open(u.context.parameters, in: program.types[e].rhs)
+  /// Returns `(lhs, rhs)` where `lhs` (respectively `rhs`) is `true` iff a `g` might be used to
+  /// prove a coercion from  from `a` to `x` (respectively `x` to `b`) for some type `x`.
```
tothambrus11: This sentence is a bit complex. Is "some type `x`" common for the lhs and rhs checks or separate? Maybe it's clearer to phrase it in first order logic:

```
 where 
   lhs = ...
   rhs = ...
```

Sources/FrontEnd/Typer/Typer.swift
```diff
@@ -57,6 +57,9 @@ public struct Typer {
     /// A table mapping identifiers to declarations.
     fileprivate typealias LookupTable = OrderedDictionary<String, [DeclarationIdentity]>
 
+    /// A pair of types.
```
tothambrus11, line 60: The documentation doesn't add any more information to what's obvious from the API. Normally, that's still useful if we want to imply that there is really not much more going on than what you see already. In this case, there is a role of the left and right side (I'm not exactly sure, depends on the semantics of `canDeriveCoercion(...)`. Or we could also leave TypePair's comment as it is, and extend `canDeriveCoercion` maps' docs.

tothambrus11:
Really nice speedup! I left some comments regarding docs (I had difficulties understanding what `canDeriveCoercion` is doing).

image

Sources/FrontEnd/Typer/Typer.swift
kyouko-taiga: No, there's really nothing more going on. It's a pair of type. The way it is _used_ by `canDeriveCoercion` is another thing. I considered creating a richer type here but that would be mostly noise.

Sources/FrontEnd/Typer/Typer.swift
kyouko-taiga: The sentence is constructed the way you suggest, only in a more compact way. Repeating the common parts does not really help because we need to juggle with 3 different subjects.

Using `x` as a placeholder for the concept of "some type that is not in the parameters" also helps keeping too much prosaic (and imprecise) text around.

Sources/Utilities/Pair.swift
kyouko-taiga: Yes, and also they cannot add conformance to non-nominal types. If they could they would need many more complicated rules to avoid unsoundness.

Sources/FrontEnd/Typer/Typer.swift
kyouko-taiga: Good point.

Sources/FrontEnd/Typer/Typer.swift
```suggestion
  /// Returns `(lhs, rhs)` where `lhs` (respectively `rhs`) is `true` iff `g` might be used to
  /// prove a coercion from from `a` to `x` (respectively `x` to `b`) for some type `x`.
```

tothambrus11: Thanks for the clarification!

kyouko-taiga, line 3861:
```suggestion
  /// Returns `(lhs, rhs)` where `lhs` (respectively `rhs`) is `true` iff a `g` might be used to
  /// prove a coercion from `a` to `x` (respectively `x` to `b`) for some type `x`.
```

PR #155, opened 2026-05-16
Implement IR lowering for bundles

kyouko-taiga:
This patch implements IR lowering for function and subscript bundles.

The feature is incomplete. In particular, this set of changes does not include access reification, which is necessary to support bundles featuring a `let` variant along with a `sink` one and/or a `set` variant along with an `inout` one.

### DIFF
--- a/Sources/FrontEnd/IR/IREmitter.swift
+++ b/Sources/FrontEnd/IR/IREmitter.swift
@@ -856,7 +856,7 @@ internal struct IREmitter {
   }
 
   /// Generates the IR using `d` as a callee that is optionally qualified by `qualification`,
-  /// achoring new instructions to `anchor`.
+  /// anchoring new instructions to `anchor`.
   ///
   /// This method implements `loweredCallee(_:output:)` for a use of `d` expressed explicitly in
   /// sources or synthesized during compilation. `mutationMarker` is, if defined, is the mutation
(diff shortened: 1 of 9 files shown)


Sources/FrontEnd/IR/IREmitter.swift
```diff
@@ -310,27 +314,41 @@ internal struct IREmitter {
...
+    }
   }
 
-  /// Generates the IR of `d` assuming the insertion context is clear.
-  private mutating func lowerInClearContext(_ d: FunctionDeclaration.ID) {
+  /// Generates the IR of `d`.
+  private mutating func lower(_ d: FunctionDeclaration.ID) {
     let f = demandLoweredDeclaration(functionOrConformance: .init(d))
-    assert(!program[module].ir[f].isDefined, "function already lowered")
+    defining(f) { (me) in
```
tothambrus11, line 332: @kyouko-taiga We should probably only define the function if it has a body. Now `@extern` functions fail at LLVM lowering because they have a block with no terminator.

We could add `precondition(f.body != nil)` to `defining` and wrap this call in an `if f.body != nil { ... }`.

kyouko-taiga: Sounds like a reasonable idea. I'll look into it.

PR #159, opened 2026-05-17
Fix testing on Xcode

kyouko-taiga:
It is not trivial to pass custom search paths to Xcode's test driver, which would be necessary to locate `lld` properly on arbitrary systems. As the use of `lld` is not a high-priority yet, it is simpler to just remove `lld` until we spend time on addressing this issue.

This patch also makes a few adjustments to the code to improve consistency.

### DIFF
--- a/Sources/Driver/Driver.swift
+++ b/Sources/Driver/Driver.swift
@@ -138,7 +138,7 @@ public struct Driver {
   /// - Requires:
   ///   - `module` has been lowered and all required transformation passes have been run.
   ///   - `module` has not been lowered to LLVM IR yet.
-  public mutating func lowerToLLVM(_ module: Module.ID) throws -> PhaseResult {
+  public mutating func compileToLLVM(_ module: Module.ID) throws -> PhaseResult {
     precondition(
       llvmModules[module] == nil, "LLVM IR already generated for '\(moduleName(module))'")
 
--- a/Sources/Driver/Driver.swift
+++ b/Sources/Driver/Driver.swift
@@ -45,9 +45,7 @@ public struct Driver {
   /// The program being compiled by the driver.
   public var program: Program
 
-  /// The corresponding LLVM module for each Hylo module.
-  ///
-  /// Populated by `lowerToLLVM(_:)`.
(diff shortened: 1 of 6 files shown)


tothambrus11:
I encountered a runtime failure on Windows during testing, I'm trying to diagnose it now. It's most likely unrelated to this PR.

tothambrus11:
I found the issue. `url(for: .itemReplacementDirectory ...)` likely has some permission-related bug on Windows. I'll report and push a more reliable workaround in a later PR.

tothambrus11 (approved):
The changes look good to me, thanks for fixing this! It's nice the clang can drive the other linkers too, this makes our job much simpler for now. I tested this branch on Windows and WSL.

I agree that the reordering of declarations are useful in this PR. I also have some changes to `CompilerTests.swift` and `Driver.swift` in the llvm-lowering branch, so in the upcoming near future let's try to avoid reordering declarations, so we keep rebasing easier.

Sources/Driver/Driver.swift
```diff
@@ -377,3 +369,14 @@ public struct Driver {
   }
 
 }
+
+extension SwiftyLLVM.Module {
+
+  /// Verifies the IR is `self` if this function has been compiled in debug mode.
```
```suggestion
  /// Verifies the IR in `self` iff this function has been compiled in debug mode.
```

kyouko-taiga:
I am all for avoiding reordering but we should also avoid mixing structs, static properties, and stored properties in seemingly arbitrary order. I don't think we need strict rules about it and there can be many exceptions but in general let's try to observe this order:

1. nested types and aliases
2. static properties
3. member properties
4. initializers
5. computed properties
6. methods
7. static methods

That's how most code has been written. Again, not a strict rule but a general guideline.

tothambrus11:
We can merge this if ready.

I think the problem I had was a build system related issue, because without any further changes, after a rebuild from scratch, the problem went away (I had a build crash in my first build due to disk being fully, then the subsequent build might might have been incorrect). 

The root of the issue anyways is that the same function is implemented both in Foundation and in FoundationEssentials. The one in FoundationEssentials is misbehaving: https://github.com/swiftlang/swift-foundation/issues/991

Issue #22, opened 2026-01-25
Allow `return` statement within if branches even if it's the only top-level expression within the function


tothambrus11:
Currrently, both of these cause a compilation error: `no coercion from Void to Int`. 
```swift
fun peek() -> Int {
  if self {
     return 1
  } else {
     return 0
  }
}
```
```swift
fun peek() -> Int {
  if self {
    let a = 20
    return a + 0
  } else {
    let b = 20
    return b + 0
  }
}
```

The compiler should notice that the if construct is being used as a statement if it contains at least one return statement. This is also Swift's behavior, so it probably makes sense to allow this.

PR #163, opened 2026-05-22
Add test for returning in if type-checking

tothambrus11:
Verifies that https://github.com/hylo-lang/hylo-new/issues/22 is fixed.


Tests/CompilerTests/positive/return-in-if.hylo
```diff
@@ -0,0 +1,19 @@
+//! stage:typing
```
kyouko-taiga, line 1: It is often a good idea to make these sorts of tests independent from the standard library so that we can test with `no-std`, which is significantly faster.

Issue #69, opened 2026-03-30
Properly determine when synthesized tuple conformances are trivial


kyouko-taiga:
The compiler won't currently conclude that the conformance of `{Int, Int}` to `Deinitializable` (or any other core trait) is trivial because it relies on source-level definitions. Specifically, the conformance is defined from the following definition, which is in the standard library:

```hylo
/// A witness that a tuple is `Deinitializable` when its elements are too.
public given <T is Deinitializable, U is Deinitializable> {T, ...U} is Deinitializable {}
```

We should add special rules to let the compiler determine that if a conformance is obtained by applying the above definition on trivial witnesses, then the resulting witness is also trivial. Doing that would then let the compiler conclude that a struct or enum trivially conforms to `Deinitializable` when all its stored properties do too.

kyouko-taiga:
> This will be useful for lowering infix+'s definition.

Not really. `AdditiveArithmetic` is not a trait whose conformance can be synthesized.

tothambrus11:
This seems to be working. The generated IR for a trivially deinitializable object doesn't include a call to deinit.

See https://github.com/hylo-lang/hylo-new/pull/164:

```hylo
//! stage:lowering

struct X is Deinitializable & Movable {
  public init() {}
}

struct Y is Deinitializable {
  let a: { X, X }
  init() {
    &self.a = (X(), X())
  }
}

fun x() {
  let y = Y()
}
```

Results in the following IR:
```hylo
fun x(set %p0: Void) {
%b0:
  %r0 = alloca Y, #preferred
  %r1 = alloca Void, #preferred
  %r2 = access [set] %r0
  %r3 = access [set] %r1
  %r4 = apply Y.init(%r2) => %r3
  %r5 = access [let] %r0
  %r6 = access [set] %p0
  %r7 = assume_state %r6 initialized
  %r8 = end %r6
  %r9 = return
}
```

<details>

<summary>full Main module IR</summary>
```
fun X.$<ConformanceDeclaration at inline-trait-implementation:3.13>() -> let Deinitializable<X> {
%b0:
  %r0 = alloca Deinitializable<X>, #preferred
  %r1 = witnesstable {Deinitializable.deinit<Self: X>} as Deinitializable<X>
  %r2 = access [set] %r0
  %r3 = store %r1 to %r2
  %r4 = end %r2
  %r5 = access [let] %r0
  %r6 = yield %r5
  %r7 = end %r5
  %r8 = access [sink] %r0
[...]

PR #164, opened 2026-05-23
Add trivially-deinitializable.hylo test

tothambrus11:
Verifies that https://github.com/hylo-lang/hylo-new/issues/69 works. (Well, we have to look at IR manually to see that it works correctly)


kyouko-taiga:
That is not the right test to make. We don't see any call to `deinit` because the value being destroyed is a struct having a trivial conformance to `Deinitializable`.

You will see the calls to the givens if you write this:
```halo
public fun main() {
  let x = (1, 2, 3)
}
```

Issue #95, opened 2026-04-14
Resolution failure in extension


kyouko-taiga:
The following code should type check:

```hylo
trait P {}
extension <T is P> T {
  fun f() { self.f() }
}
```

But it seems typer is unable to conclude that `T` conforms to `P` in the scope of `f`.

```
test.hylo:7.18-19: error: type 'T' has no member 'f'
  fun f() { self.f() }
                 ^
```

tothambrus11:
This seems to work already. I added a test cases in #166

PR #166, opened 2026-05-23
Add resolution-in-extension.hylo test

tothambrus11:
Verifies that #95 works.

### DIFF
--- a/Tests/CompilerTests/positive/resolution-in-extension.hylo
+++ b/Tests/CompilerTests/positive/resolution-in-extension.hylo
@@ -1,6 +1,7 @@
 //! stage:typing no-std
 
 trait P {}
+
 extension <T is P> T {
   fun f() { self.f() }
 }
--- a/Tests/CompilerTests/positive/resolution-in-extension.hylo
+++ b/Tests/CompilerTests/positive/resolution-in-extension.hylo
@@ -1,0 +1,6 @@
+//! stage:typing no-std
+
+trait P {}
+extension <T is P> T {
+  fun f() { self.f() }
+}


Tests/CompilerTests/positive/resolution-in-extension.hylo
```diff
@@ -0,0 +1,6 @@
+//! stage:typing no-std
+
+trait P {}
+extension <T is P> T {
```
kyouko-taiga, line 5: nitpick:
```suggestion

extension <T is P> T {
```

kyouko-taiga (approved):
Thanks! The test looks good to me.

I think I would test more complex situations in the same test case, because this one looks a bit too trivial. Maybe we want to do that with a subsequent PR, though, especially if the compiler fails (which would made me only half surprised 😅).

Anyway, here are a few things I would test:

``` hylo
trait P { type X }

struct A<T> {}

// A more complex conditional extension.
extension <T is P> A<T> {
  public fun f() {}
}

// Something involving an associated type.
extension <T is P where T.X is P> A<T> {
  public fun g() {}
}

public main f(a: A<T>) {
  given Void is P { public type X = Void }
  a.f() // by the first extension
  a.g() // by the second extension
}
```

tothambrus11:
This test case failed:
```
//! stage:typing no-std
trait P { type X }

struct A<T> {}

// A more complex conditional extension.
extension <T is P> A<T> {
  public fun f() {}
}

// Something involving an associated type.
extension <T is P where T.X is P> A<T> {
  public fun g() {}
}

public fun main(a: A<Void>) {
  given Void is P { public type X = Void }
  a.f() // by the first extension
  a.g() // by the second extension // error: type 'A<Void>' has no member 'g'
}
```

PR #176, opened 2026-05-25
Stand up LLVM code generation

kyouko-taiga:


Sources/BackEnd/ConcreteLayout.swift
```diff
@@ -0,0 +1,24 @@
+import SwiftyLLVM
+
+/// A description of the layout of a Hylo for a particular host.
```
tothambrus11:
```suggestion
/// A description of the layout of a Hylo type for a particular target.
```

Docs/SubscriptDecomposition.md
```diff
@@ -0,0 +1,61 @@
+# Subscript Decomposition
+
+Intuitively, a subscript can be understood as a pair of functions.
+The first, called "ramp", computes a value to project.
+The second, called "slide" undoes the setup made to compute that value and potentially applies changes made to the projected value back to the whole of which it is a part.
```
tothambrus11:
```suggestion
The second, called "slide" undoes the setup made to compute that value and potentially applies changes made to the projected value back to the whole which it is a part of.
```

Sources/FrontEnd/IR/Mangling/DemangledSymbol.swift
```diff
@@ -134,6 +134,12 @@ internal indirect enum DemangledEntity: Hashable, Sendable {
   /// An IRFunction with `name == .existentialized`.
   case existentialized(DemangledEntity)
 
+  /// An IRFunction with `name == .slide`.
+   case slide(DemangledEntity, Int)
+
+  /// An IRFunction with `name == .plateau`.
+   case plateau(DemangledEntity, Int)
```
tothambrus11:
```suggestion
  /// An IRFunction with `name == .slide`.
  case slide(DemangledEntity, Int)

  /// An IRFunction with `name == .plateau`.
  case plateau(DemangledEntity, Int)
```

Sources/FrontEnd/IR/Instructions/IRAlloca.swift
```diff
@@ -1,17 +1,17 @@
...
-/// The result of the instruction is the address of stack-allocated storage capable of holding one
-/// instance of `storageType`. The storage is uninitialized and deallocated automatically when the
-/// function returns, at which point it must be deinitialized.
+/// The instruction defines a place capable of storing an instance of `storage`, allocated on the
+/// stack. The place is uninitialized after its creation and it must be deinitialized before its
+/// deallocation, which occurs automatically when the function returns.
 ///
 /// Unlike LLVM's alloca, this instruction cannot be used to allocate dynamically sized buffers. It
 /// is nonetheless possible to allocate storage for a fixed number of contiguous instances using a
 /// tuple (e.g., `Int[8]` in surface syntax).
```
tothambrus11, line 11: This use case seems to be missing an instruction now that we deleted `allocx`. We can add it later though.

Docs/SubscriptDecomposition.md
kyouko-taiga: I think my sentence is correct and has the advantage of not leaving a trailing preposition, which I believe is not grammatically correct.

Sources/FrontEnd/IR/Instructions/IRAlloca.swift
kyouko-taiga: I did not follow. What case are you referring to?

tothambrus11: Stack-allocating a temporary buffer with dynamic size. Technically, we could construct a new type witness with custom size, but could be cleaner to have a dedicated instruction or constructor variant of alloca.

kyouko-taiga: I don't think we need a dedicated instruction for that. At least I don't think that's a priority atm.

Sources/BackEnd/ConcreteLayout.swift
```diff
@@ -0,0 +1,24 @@
...
+/// A description of the layout of a Hylo type for a particular target.
+internal struct ConcreteLayout {
+
+  /// The LLVM types of the fields constituting the parts of an instance, if any.
+  ///
+  /// This property may be empty if the type described by `self` has no stored property, either
+  /// because it is opaque or because it is a machine type.
+  internal let fields: [SwiftyLLVM.AnyType.UnsafeReference]
+
+  /// A map from the index of a property to the index of its corresponding field, or `-1` if that
```
tothambrus11:
```suggestion
  /// A map from the index of a property to the index of its corresponding LLVM field, or `-1` if that
```

Sources/BackEnd/FunctionGenerationContext.swift
```diff
@@ -0,0 +1,69 @@
...
+  internal let result: FunctionMetadata
+
+  /// A map from Hylo IR basic block to its LLVM counterpart.
+  internal var block: [FrontEnd.IRBlock.ID: SwiftyLLVM.BasicBlock.UnsafeReference]
+
+  /// A map from Hylo IR register to its LLVM counterpart, unless it has been erased.
+  internal var value: [FrontEnd.IRValue: SwiftyLLVM.AnyValue.UnsafeReference]
+
+  /// The set of basic blocks that have been factored out into a plateau.
+  internal var factoredOut: IRBlockSet
```
tothambrus11, line 26:
```suggestion
  internal var plateau: IRBlockSet
```

Sources/BackEnd/FunctionMetadata.swift
```diff
@@ -0,0 +1,56 @@
...
+    internal let convention: PassingConvention
+
+  }
+
+  /// A handle to the LLVM instance representing the compiled function.
+  internal let llvm: SwiftyLLVM.Function.UnsafeReference
+
+  /// A table mapping each term parameter of the Hylo function to its passing convention.
+  ///
+  /// This property describes how each parameter of an Hylo IR function are represented (or not) in
```
tothambrus11:
```suggestion
  /// This property describes how each parameter of a Hylo IR function are represented (or not) in
```

Sources/BackEnd/Program+LLVM.swift
```diff
@@ -0,0 +1,771 @@
...
+    if f.output!.type.layout.size > 0 {
+      let v = ctx.llvm.insertExtractValue(from: r, at: 0, at: p)
+      ctx.llvm.insertReturn(v, at: p)
+    } else {
+      ctx.llvm.insertReturn(i32.unsafe[].zero, at: p)
+    }
+  }
+
+  /// Returns the representation of `v` in LLVM IR.
+  private mutating func codegen(
```
tothambrus11, line 920:
```suggestion
  private mutating func llvmValue(
```

Sources/BackEnd/Program+LLVM.swift
```diff
@@ -0,0 +1,771 @@
...
+      return codegen(integer: n, instanceOf: t, in: &ctx.module)
+    case .register:
+      return ctx.value[v]!
+    default:
+      fatalError("no LLVM representation of type '\(show(v))'")
+    }
+  }
+
+  /// Returns a LLVM IR constant equivalent to `n`, which is an instance of an integer type `t`.
+  private mutating func codegen(
```
tothambrus11, line 936:
```suggestion
  private mutating func llvmValue(
```

Sources/BackEnd/ConcreteLayout.swift
```diff
@@ -0,0 +1,24 @@
+import SwiftyLLVM
+
+/// A description of the layout of a Hylo type for a particular target.
+internal struct ConcreteLayout {
+
+  /// The LLVM types of the fields constituting the parts of an instance, if any.
```
tothambrus11, line 50: How should we lower code that takes the address of a zero-sized field? Or if that's impossible, could the frontend prevent that reliably?

kyouko-taiga: We can determine that all zero-sized objects get some implementation-defined address. The specification of Hylo does not guarantee that (sub-)objects have a unique address.

Sources/BackEnd/FunctionGenerationContext.swift
kyouko-taiga: I don't think the suggestion makes it clearer what the property is about at call sites. At the margin it should be `factoredOutInPlateau` but the call site does not really care _why_ the blocks have been factored out.

Sources/BackEnd/ConcreteLayout.swift
tothambrus11: Ah yes, that's great! I used this trick to create an SSA value in Ferlium to avoid extra alloca instructions for `Void` return slots.

Sources/BackEnd/Prototype.swift
```diff
@@ -0,0 +1,60 @@
+import SwiftyLLVM
+
+/// The signature of a LLVM IR function compiled form a Hylo IR function.
```
tothambrus11, line 3:
```suggestion
/// The signature of a LLVM IR function compiled from a Hylo IR function.
```

Sources/BackEnd/Program+LLVM.swift
```diff
@@ -0,0 +1,1120 @@
...
+
+    let iptr = ctx.llvm.iptr
+    let payloadSize = s.utf8.count
+    let pointerSize = ctx.llvm.layout.pointerSize
+
+    // Do the contents fit inline storage?
+    if (payloadSize < pointerSize) && (pointerSize <= 8) {
+      var units = UInt64(truncatingIfNeeded: payloadSize) << 2
+      for (i, u) in s.utf8.enumerated() {
+        units |= UInt64(u) << (i + 1) * 8
```
tothambrus11, line 779:
```suggestion
        units |= UInt64(u) << ((i + 1) * 8)
```
Was this the intended operator grouping?

Sources/BackEnd/ConcreteLayout.swift
```diff
@@ -0,0 +1,68 @@
...
+    }
+
+    /// Returns `true` iff `l` does not denote a fixed size of `r` bytes.
+    internal static func != (l: Self, r: Int) -> Bool {
+      (l.raw != r) || (r < 0)
+    }
+
+    /// Returns `true` iff `l` denotes a fixed size of less than or exactly `r` bytes.
+    internal static func <= (l: Self, r: Int) -> Bool {
+      UInt(bitPattern: l.raw) < r
```
tothambrus11:
```suggestion
      UInt(bitPattern: l.raw) <= r
```

Sources/BackEnd/Program+LLVM.swift
```diff
@@ -0,0 +1,1120 @@
...
+        let padding = next - size
+        fields.append(ctx.llvm.arrayType(padding, ctx.llvm.i8).asAnyType)
+      }
+
+      rowToField[p] = fields.count
+      fields.append(rs[p].llvm)
+      size = next + fieldSize
+    }
+
+    let a = rs.last?.layout.alignment ?? 1
```
tothambrus11, line 1139:
```suggestion
    let a = if let max = ps.last { rs[max].layout.alignment } else { 1 }
```
`rs` is not reordered, only `ps`.

Sources/FrontEnd/IR/Mangling/ManglingEncoding.swift
```diff
@@ -570,7 +580,15 @@ internal struct ManglingEncoding: Sendable {
       append(typeArguments: a, to: &output)
     case .existentialized(let s):
       output.add(operator: .existentializedDeclaration)
-      append(function: s, of: m, to: &output)
+      append(function: s, to: &output)
+    case .slide(let s, let n):
```
tothambrus11, line 570: Slide and plateau don't seem to have a demangling implementation yet.

Sources/BackEnd/Program+LLVM.swift
```diff
@@ -0,0 +1,1120 @@
...
+        let t = ctx.llvm.functionType(from: inputs, to: o.llvm)
+        return .init(
+          signature: t,
+          mapping: .init(inputs: inputMap, output: .init(type: o, convention: .byValue(-1))))
+      }
+
+      // The result is returned by output parameter.
+      else {
+        let n = inputs.count
+        let t = ctx.llvm.functionType(from: Array(inputs, terminatedBy: o.llvm), to: o.llvm)
```
tothambrus11, line 636: Should this have `to: ctx.llvm.void` and an extra `ptr` input?

kyouko-taiga: Good catch. The extra parameter is added by `Array(inputs, terminatedBy: o.llvm)` but the output type is incorrect.

tothambrus11:
A Claude review of my PR accidentally reviewed this one, and found some bugs in the subscript lowering:

### Bug 1 — a projection inside a slide crashes the compiler (nil unwrap)

If the code after a yield (the "slide") itself contains a projection — i.e. another subscript call, or anything that lowers to a project instruction such as .copy() — the compiler crashes.

Reproducer:
```hylo
subscript id(x: inout Int32) inout -> Int32 { yield &x }

subscript proj(x: inout Int32, y: inout Int32) inout -> Int32 {
  yield &y
  inout w = &id[&x]   // <-- a `project` lives in the slide
  &w = 7
}

fun main() -> Int32 {
  var a: Int32 = 0
  var b: Int32 = 0
  inout p = &proj[&a, &b]
  &p = 5
  return a                // expected: 7
}
```
Result: Fatal error: Unexpectedly found nil at Program+LLVM.swift:194, via setupDominatingDefinitions → incorporateContentsOfPlateau → definePlateau.

Root cause (confirmed by instrumenting the unwrap): the nil source is parameter(1). When a slide body contains a project, definePlateau builds the nested plateau's prologue with ctx.prologue(dominating:), and prologue collects the captured parameters from result.prototype.mapping.inputs (SubscriptDecomposition.swift:71). But result is the slide's FunctionMetadata, whose mapping inputs are empty (defineSlide sets Prototype.Mapping(inputs: [], ...), line 197; definePlateau likewise, line 259). So the original subscript's parameters are never captured for the inner plateau, and the redefine loop at SubscriptDecomposition.swift:486 dereferences a parameter that was never mapped. In short: nested projections (a projection within another projection's slide/plateau region) take the parameter set from the wrong function.



### Bug 2 — a conditional yield produces invalid LLVM IR

A subscript that yields on more than one control-flow path fails LLVM verification.

Reproducer:
```hylo
subscript pick(c: Bool, x: inout Int32, y: inout Int32) inout -> Int32 {
  if c { yield &x } else { yield &y }
}

fun main() -> Int32 {
  var a: Int32 = 0
  var b: Int32 = 0
  inout p = &pick[true, &a, &b]
  &p = 7
  return a                // expected: 7
}
```
Result: LLVMError: Function return type does not match operand type of return inst! ret void / i32.

The two yields sit in separate blocks that both branch to the return block, producing two slides. A function whose LLVM return type is i32 (a ramp or plateau) is getting a ret void — the multi-slide / multi-successor return-control wiring in incorporate(IRReturn …) / insertReturn emits the wrong return-instruction type for one of the generated functions. (Single-yield subscripts, like the existing subscript-decomposition.hylo test, work fine, so this is specific to yields on multiple paths.)

kyouko-taiga:
> A Claude review of my PR accidentally reviewed this one, and found some bugs in the subscript lowering:

Did you confirm this yourself? Am I reading your own words? If so, please open issues so we can properly keep track of these.

PR #178, opened 2026-05-25
Port conditional compilation support from old compiler.

lucteo:
Add support for conditional compilation support at statement level. Add parser tests.
Add tests for conditional compilation.

Known issues: parsing of semantic versions requires spaces around dot (to not take floating literals).

Limitation: in this patch, we don't allow the user to configure the compilation factors.

### DIFF
--- a/Sources/FrontEnd/Module.swift
+++ b/Sources/FrontEnd/Module.swift
@@ -212,6 +212,9 @@ public struct Module: Sendable {
   /// The position of `self` in the containing program.
   public let identity: Module.ID
 
+  /// Conditions for selecting conditional compilation branches.
+  private let compilationConditions: ConditionalCompilationFactors
+
   /// The dependencies of the module.
   public private(set) var dependencies: [Module.Name] = []
 
@@ -221,10 +224,11 @@ public struct Module: Sendable {
   /// The IR functions in the module.
   public internal(set) var ir: IR = .init()
 
-  /// Creates an empty module with the given name and identity.
-  public init(name: Name, identity: Module.ID) {
+  /// Creates an empty module with the given name, identity, and compilation conditions.
+  public init(name: Name, identity: Module.ID, compilationConditions: ConditionalCompilationFactors) {
     self.name = name
     self.identity = identity
+    self.compilationConditions = compilationConditions
   }
 
   /// `true` iff `self` is Hylo's standard library.
@@ -276,7 +280,7 @@ public struct Module: Sendable {
     if let f = sources.index(forKey: s.name) {
       return (inserted: false, identity: .init(module: identity, offset: f))
     } else {
-      let p = Parser(s)
+      let p = Parser(s, compilationConditions: compilationConditions)
       var f = Module.SourceContainer(
         identity: .init(module: identity, offset: sources.count), source: s)
       p.parseTopLevelDeclarations(in: &f)
@@ -484,6 +488,9 @@ extension Module: Archivable {
     /// A mapping from file name to its contents.
     internal var sources = OrderedDictionary<FileName, SourceFile>()
 
+    /// Conditions for selecting conditional compilation branches.
+    internal var compilationConditions: ConditionalCompilationFactors
+
     /// Returns the result of calling `action` on `self` wrapped in `Any`.
     internal mutating func withWrapped<T>(_ action: (inout Any) throws -> T) rethrows -> T {
       try withUnsafeMutablePointer(to: &self) { (this) in
@@ -523,7 +530,8 @@ extension Module: Archivable {
         ctx.modules[Module.ID(i + 1)] = ctx.identities[d]
       }
 
-      var me = Self(name: name, identity: ctx.identities[name]!)
+      var me = Self(name: name, identity: ctx.identities[name]!,
+        compilationConditions: ctx.compilationConditions)
 
       // <source count> <identity> <contents> ...
       let count = try Int(archive.readUnsignedLEB128())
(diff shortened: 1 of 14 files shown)


Sources/FrontEnd/Parser/Parser.swift
```diff
@@ -2201,6 +2208,287 @@ public struct Parser {
...
+  ///       '#else' '{' statement-list '}' `#endif`
+  ///       `#endif`
+  ///     conditional-compilation-condition ::=
+  ///       `true`
+  ///       `false`
+  ///       `operatingSystem` `(` identifier `)`
+  ///       `architecture` `(` identifier `)`
+  ///       `feature` `(` identifier `)`
+  ///       `compiler` `(` identifier `)`
+  ///       `compilerVersion` `(` comparison `)`
```
tothambrus11, line 2228: What is the difference between these two? Does the first prepare for the case when we have multiple compilers?

lucteo: the first one conditions on the compiler name (e.g., hyloc, ghylo, etc.)
the second one conditions on the compiler version. e.g., `compiler_version(>= 1.0.0)`

Sources/FrontEnd/Syntax/Statements/ConditionalCompilation.swift
```diff
@@ -0,0 +1,213 @@
+import Archivist
+
+/// A conditional-compilation statement.
+@Archivable
+public struct ConditionalCompilation: Equatable, Statement {
```
kyouko-taiga, line 5: Swift calls that a compiler [control statement](https://docs.swift.org/swift-book/documentation/the-swift-programming-language/statements/#Compiler-Control-Statements).

IMO this names makes more sense. The statement itself isn't conditional compilation; that's the name of the feature.

Sources/FrontEnd/Parser/Parser.swift
```diff
@@ -45,10 +45,14 @@ public struct Parser {
   /// The context in which the parser is being used.
   private var context: Context = .default
 
+  /// Conditions for selecting conditional compilation branches.
+  private let compilationConditions: ConditionalCompilationFactors
+
   /// Creates an instance parsing `source`.
-  public init(_ source: SourceFile) {
+  public init(_ source: SourceFile, compilationConditions: ConditionalCompilationFactors) {
```
kyouko-taiga, line 52:
```suggestion
  public init(_ source: SourceFile, settings: ConditionalCompilationFactors) {
```
Those are not conditions.

Sources/FrontEnd/Parser/Parser.swift
```diff
@@ -2201,6 +2208,287 @@ public struct Parser {
...
+  }
+
+  /// Parses a conditional compilation body.
+  ///
+  /// If `skipParsing` is `true`, all the tokens are skipped until the next `#elseif`, `#else`, or
+  /// `#endif` that matches the current nesting level.
+  private mutating func parseConditionalCompilationStatements(
+    skipParsing: Bool,
+    in file: inout Module.SourceContainer
+  ) throws -> [StatementIdentity] {
```
kyouko-taiga, line 2271:
```suggestion
  private mutating func parseConditionalCompilationBody(
    interpretingToken tokensAreInterpreted: Bool,
    in file: inout Module.SourceContainer
  ) throws -> [StatementIdentity] {
```

Sources/FrontEnd/Parser/Parser.swift
```diff
@@ -2201,6 +2208,287 @@ public struct Parser {
...
+  ///       `or` `(` conditional-compilation-condition `,` conditional-compilation-condition `)`
+  ///     comparison ::=
+  ///       `>=` semantic-version
+  ///       `<` semantic-version
+  ///     semantic-version ::=
+  ///       decimal-number '.' decimal-number '.' decimal-number
+  ///
+  private mutating func parseConditionalCompilation(
+    in file: inout Module.SourceContainer
+  ) throws -> ConditionalCompilation.ID {
```
kyouko-taiga, line 2241: If you simply skip the branch that's not satisfied by the conditional complication factors, is there a point creating a new tree?

Sources/FrontEnd/Parser/Parser.swift
```diff
@@ -2201,6 +2208,287 @@ public struct Parser {
...
+    if skipParsing {
+      try skipConditionalCompilationBranch(stoppingAtElse: false)
+      guard take(.poundEndif) != nil else { throw expected("'#endif'") }
+      return []
+    }
+
+    if let head = take(.poundElseif) {
+      return [.init(try parseConditionalCompilationTail(head: head, canBeActive: active, in: &file))]
+    }
+    else if take(.poundElse) != nil {
```
kyouko-taiga, line 2329:
```suggestion
    } else if take(.poundElse) != nil {
```

Sources/FrontEnd/Parser/Parser.swift
```diff
@@ -2201,6 +2208,287 @@ public struct Parser {
...
+      let b = try parseConditionalCompilationStatements(skipParsing: false, in: &file)
+      guard take(.poundEndif) != nil else { throw expected("'#endif'") }
+      return b
+    }
+    else if take(.poundEndif) != nil {
```
kyouko-taiga, line 2334:
```suggestion
    } else if take(.poundEndif) != nil {
```

Sources/FrontEnd/Module.swift
kyouko-taiga, line 577: The compilation condition factors should probably be serialized and deserialized. I'm surprised to see no change in this function.

kyouko-taiga (commented):
Nice work.

I would like to discuss the option of reusing more of the existing syntax to express conditional compilation. I think we could get away with a construct like `'compiletime' 'if' expression '{' statements '}' ...` and use normal expressions to express conditions. We can then have special casing in typer to deal with your conditions.

The advantage would be that we don't have to introduce new syntax, with all that implies in terms of language surface and parser complexity.

PR #179, opened 2026-05-25
Store IDs rather than names in 'IRValue.function'

kyouko-taiga:
Although a name uniquely identifies a function, IDs are more convenient during code generation.


Sources/FrontEnd/IR/IREmitter.swift
```diff
@@ -2579,10 +2579,10 @@ internal struct IREmitter {
 
   /// Returns a reference to the given lowered function.
   internal mutating func functionReference(to f: IRFunction.ID) -> IRValue {
-    let n = program[module].ir[f].name
+    // let n = program[module].ir[f].name
```
tothambrus11, line 2582: This can be probably removed.

PR #182, opened 2026-05-26
Fix the lowering of conditional statements

### LINKED ISSUE #120: Handling returned control flow in IREmitter
I got a bit confused by the lowering of the `return` statement, specifically about the comment at the end:

```swift
  /// Generates the IR of `s`.
  private mutating func lower(_ s: Return.ID) -> ControlFlow {
    let r = currentFunction.returnRegister!

    // Store the return value into the return register.
    if let e = program[s].value {
      lower(store: e, to: r)
    } else if currentFunction.result(of: r)?.type == .void {
      lowering(s, { $0._assume_state(r, initialized: true) })
    }

    // The return instruction is emitted by the caller handling this control-flow effect.
    return .return(s)
  }
```

It's used in `private mutating func lower(_ s: StatementIdentity) -> ControlFlow {`, which in turn is used in `private mutating func lowerAsStatement(_ s: If.ID) -> ControlFlow {`.

The latter function only inspects whether the control flow equals `next`, otherwise it discards the returned control flow. This may lead to some bugs where an inner statement wants to return, but the conditional doesn't apply this intention. Am I missing something, or is this a bug? How could we make this easier to reason about?

kyouko-taiga:
This patch fixes #120. Specifically:

- It fixes a bug that would leave instructions in the function after a block has been removed.
- It fixes a bug that would leave unreachable blocks after dead code elimination.
- It fixes a bug that would emit invalid IR for conditional statements featuring early returns.

### DIFF
--- a/Sources/FrontEnd/IR/IREmitter.swift
+++ b/Sources/FrontEnd/IR/IREmitter.swift
@@ -526,26 +526,32 @@ internal struct IREmitter {
   @discardableResult
   private mutating func lowerAsStatement(_ s: If.ID) -> ControlFlow {
     let onFailure = insertionContext.function!.addBlock()
-    let tail = insertionContext.function!.addBlock()
+    let after = insertionContext.function!.addBlock()
 
     // Lower the conditions and the success branch.
     for c in program[s].conditions {
       insertionContext.point = .end(of: lowerCondition(c, onFailure: onFailure))
     }
 
-    if lower(program[s].success) == .next {
-      lowering(after: program[s].success, { $0._br(tail) })
+    switch lower(program[s].success) {
+    case .next:
+      lowering(after: program[s].success, { $0._br(after) })
+    case .return(let r):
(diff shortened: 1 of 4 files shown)


kyouko-taiga:
Thanks @LouisMLettry @floriandesmons.

tothambrus11:
Thanks, I tested this now with a couple of examples, it works nicely! It would be nice to add some higher confidence test cases here that assert some properties about the IR, or just add IR verification assertions in general (e.g. all blocks have a single terminator at the end). Maybe behavioral tests (when LLVM lowering is done) would be enough, with full branch coverage of a Hylo function. If it's not so easy to add a test, we can also just make an issue and add it later.

kyouko-taiga:
I agree that we need more checks. However it's a whole task to write them. There are some foundations but I expect it will take at least one day of work to write the tests for all invariants and as much time to review it.

PR #187, opened 2026-06-01
Interpreter: Make empty program working

RishabhRD:
This is an initial work for porting interpreter from OFE to NFE. Created `interpreter` branch from current `main` branch to merge changes of interpreter. Once we would have good work, we can merge `interpreter` to `main`.

The PR contains minimal changes needed to run an empty main function on interpreter.


Sources/Interpreter/Interpreter.swift
```diff
@@ -0,0 +1,268 @@
+import FrontEnd
+import Utilities
+
+/// The position of an instruction in the program.
+private struct CodePointer {
```
dabrahams: `InstructionPointer` or `InstructionAddress` could be slightly better.  A code pointer points to "some code" which might have any granularity.

Sources/Interpreter/Interpreter.swift
```diff
@@ -0,0 +1,268 @@
...
+
+  /// The position relative to `functionInModule` indicated by `self`.
+  var instructionInFunction: AnyInstructionIdentity
+
+}
+
+/// A value manipulated by the IR.
+private struct Value {
+  /// The underlying type-erased representation of value.
+  public var payload: Any
```
dabrahams: My instinct says

```suggestion
  public var storage: Any
```

“Payload” connotes that the thing carrying it has some substantial general purpose other than storing the payload (usually, delivering it somewhere).

Sources/Interpreter/Interpreter.swift
```diff
@@ -0,0 +1,268 @@
...
+private enum InstructionResult {
+
+  /// Produces a value.
+  ///
+  /// Execution continues at the next instruction in sequence.
+  case value(Value)
+
+  /// Transfer control to specific instruction.
+  case jump(CodePointer)
+
```
dabrahams: The comment on `value` is written as though it were a function, which makes it clash with the description of the enum.

```suggestion
/// The part of one instruction's execution that follows any memory and I/O effects.
///
/// Each instruction ends by either initializing a constant
/// register associated with the instruction's address 
/// and current stack frame, or transfers control to another instruction.
private enum InstructionEpilogue {

  /// Store
  case initializeRegister(to: Value)

  /// Control is transferred to the given instruction.
  case jump(to: AnyInstructionIdentity)

```

Sources/Interpreter/Interpreter.swift
```diff
@@ -0,0 +1,268 @@
...
+}
+
+/// The local variables, parameters, and return address for a function
+/// call.
+private struct StackFrame {
+
+  // TODO: add local variables and parameters, need to introduce `Memory` for the same.
```
```suggestion
  // TODO: add local variables and parameters, which require `Memory`.
```

Try to avoid “the same.”  It gets hard to understand what “same” refers to.  And it's obscure.  Approximately nobody talks like that.  Usually “it” is a good substitute, but not here.

Sources/Interpreter/Interpreter.swift
```diff
@@ -0,0 +1,268 @@
...
+  /// Execution continues at the next instruction in sequence.
+  case value(Value)
+
+  /// Transfer control to specific instruction.
+  case jump(CodePointer)
+
+}
+
+/// The local variables, parameters, and return address for a function
+/// call.
```
dabrahams, line 50: I keep thinking this could be easier to describe as “The ephemeral (or non-`Memory`) execution state of a function call.”  To do that would require storing the program counter for the function in the stack frame, rather than the return address.  That would allow the other storage of the program counter to disappear.

Sources/Interpreter/Interpreter.swift
```diff
@@ -0,0 +1,268 @@
...
+      precondition(!isEmpty)
+      yield frames[frames.count - 1]
+    }
+    _modify {
+      precondition(!isEmpty)
+      yield &frames[frames.count - 1]
+    }
+  }
+
+  /// Boolean indicating whether stack contains atleast 1 stack frame.
```
dabrahams: Omit needless words, especially type information.

```suggestion
  /// Whether there is at least 1 stack frame.
```

or

```suggestion
  /// `true` iff there is at least 1 stack frame.
```

We should be consistent with which form we use.  I don't know what the rest of the code does.  Ask @kyouko-taiga what's done in hylo-new

Sources/Interpreter/Interpreter.swift
dabrahams: It would probably be most effective to investigate this angle after this PR is landed

Sources/Interpreter/Interpreter.swift
```diff
@@ -0,0 +1,268 @@
...
+      }
+      return
+    case nil:
+      try advanceProgramCounter()
+    }
+
+  }
+
+  /// Executes a single instruction without recording its result.
+  private mutating func stepResult() throws -> InstructionResult? {
```
```suggestion
  /// Applies the `Memory` and I/O effects of the current instruction and returns its epilogue.
  private mutating func applyCurrentInstruction() throws -> InstructionEpilogue? {
```

But what's this optional doing here?  It's not explained by the doc comment.  Can't we just return a `Void` stored in the `Any` rather than returning `nil`?

Sources/Interpreter/Interpreter.swift
```diff
@@ -0,0 +1,268 @@
...
+    switch currentInstruction {
+    case is IRAccess:
+      // TODO: fake implementation to make empty program work.
```
dabrahams: As phrased the comment says, “I'll need to create a fake implementation to make empty program work.”


```suggestion
      // TODO: add a real implementation
```

It would be nice to see an explanation of what a real implementation would do, because it's not obvious what effect an `IRAccess` instruction should have.

Sources/Interpreter/Interpreter.swift
```diff
@@ -0,0 +1,268 @@
...
+    case let x as IRTypeWitness:
+      _ = x
+    case let x as IRUnreachable:
+      _ = x
+    case let x as IRWitnessTable:
+      _ = x
+    case let x as IRYield:
+      _ = x
+    default:
+      fatalError("Interpreter: unimplemented instruction")
```
dabrahams, line 238: It's a little weird that we have this fatal error here for all unimplemented instructions except for a select few that are stubbed to make a test case work.  It would be good to eliminate the stubs ASAP after the PR lands because the fatal error makes it look like we are being more complete than we actually are.

Sources/Interpreter/Interpreter.swift
```diff
@@ -0,0 +1,268 @@
...
+
+  }
+
+  /// Removes topmost stack frame and return code pointer to next instruction of any
+  /// previous stack frame.
+  ///
+  /// - Precondition: the program is running.
+  private mutating func popStackFrame() -> CodePointer {
+    // precondition(topOfStack.allocations.isEmpty,
+    //     "All local variables allocations for function must be deallocated before returning.")
```
```suggestion
    //     "Function returns before deallocating all local variable storage.")
```

Sources/Interpreter/Interpreter.swift
```diff
@@ -0,0 +1,268 @@
...
+        .functions[programCounter.functionInModule]
+        .instruction(after: programCounter.instructionInFunction)
+    else { throw IRError() }
+    programCounter.instructionInFunction = i
+  }
+}
+
+extension Program {
+  /// Entry module and entry function.
+  fileprivate var entry: (module: Module.ID, function: IRFunction.ID) {
```
dabrahams: Don't we have an abstraction like `GlobalFunctionID` that bundles the module ID with the function ID?  If not, probably should make one.  Maybe even call it `Program.EntryPoint`

```suggestion
  /// The function whose invocation executes the whole program.
```

Tests/InterpreterTests/InterpreterTests.swift
```diff
@@ -0,0 +1,13 @@
+import XCTest
+
+final class InterpreterRunTests: XCTestCase {
+
+  func testEmptyProgram() async throws {
+    let s =
+      """
+        public fun main() { }
+      """.sourceFile
+    try await runOnInterpreter(s)
```
dabrahams: If we're going to de-verbosify that name so much, we may as well use `interpret`.  


```suggestion
    try await interpret(s)
```

Also, that should be an extension on `String`.  

```suggestion
    try await s.interpret()
```

I feel like a lot of work I did to make testing of simple source files embedded in strings work in the OFE and report good diagnostics is being thrown out here.  

Perhaps @kyouko-taiga decided it was too complex to be worth the code required to support it. If so, you shouldn't use this form and should instead create a `.hylo` file and follow some other pattern.  Personally I find it difficult to keep track of all the different things I want to test if I need to put each tiny “file” in a separate file, but realistically, what I find easy doesn't matter.

Tests/InterpreterTests/Utilities.swift
```diff
@@ -0,0 +1,59 @@
+import FrontEnd
+import Interpreter
+
+extension String {
+  /// Interpret `self` as `SourceFile`.
```
dabrahams: The word “Interpret” is misleading here.

Tests/InterpreterTests/Utilities.swift
```diff
@@ -0,0 +1,59 @@
...
+  if p[m].containsError {
+    printDiagnostics(p[m])
+    throw ProgramError.loweringFailed
+  }
+
+  var logger = TreePrinter(program: p)
+
+  var executor = Interpreter(p)
+  while executor.isRunning {
+    print(executor.currentInstruction.show(using: &logger))
```
dabrahams, line 11: If this is temporary, a comment should say so.  If not, the function's doc comment should describe the effect.

Tests/InterpreterTests/Utilities.swift
```diff
@@ -0,0 +1,59 @@
...
+func runOnInterpreter(_ s: SourceFile) async throws {
+  var p = Program()
+  let m = p.demandModule("Main")
+  p[m].addDependency(Module.standardLibraryName)
+  _ = p[m].addSource(s)
+
+  await p.assignScopes(m)
+  p.assignTypes(m, loggingInferenceWhere: { _, _ in false })
+  if p[m].containsError {
+    printDiagnostics(p[m])
```
dabrahams: If this is temporary, a comment should say so. If not, the function's doc comment should describe the effect.

Tests/InterpreterTests/Utilities.swift
```diff
@@ -0,0 +1,59 @@
...
+  /// Interpret `self` as `SourceFile`.
+  var sourceFile: SourceFile {
+    // TODO: Needed `asSourceFile` from OFE. However, `SourceFile` needs some
+    // amount of work for the same.
+    SourceFile(contents: self)
+  }
+}
+
+/// Runs program in `s` on `Interpreter`.
+func runOnInterpreter(_ s: SourceFile) async throws {
```
dabrahams: This function has several phases that should be broken out into separate functions each with a descriptive name.

Tests/InterpreterTests/Utilities.swift
```diff
@@ -0,0 +1,59 @@
...
+/// Print diagnostics in `m`.
+private func printDiagnostics(_ m: Module) {
+  for d in m.diagnostics {
+    var r = ""
+    d.render(into: &r, style: .unstyled)
+    print(r, terminator: "")
+  }
+}
+
+/// Error generated by `Program` while processing.
```
dabrahams: It is not generated by the program, but by the compiler.

Tests/InterpreterTests/Utilities.swift
```diff
@@ -0,0 +1,59 @@
...
+private enum ProgramError: Error {
```
dabrahams: I don't understand why this thing isn't just `Diagnostics`.

The word `Program` in `ProgramError` is not very descriptive.  Everything in this compiler has do to with programs and in fact a “program error” would be a likely concept in any program.

dabrahams (approved):
I am approving so you can make progress.  I expect you to make judgement calls about which suggestions to apply now, which to defer to a later PR, and which to discuss with me.

Sources/Interpreter/Interpreter.swift
tothambrus11: From what I saw, these are the conventions:
- For properties: `true` iff ...
- For functions: Returns `true` iff ...

dabrahams: Yes please.

RishabhRD:
Merged main branch to this branch to make CI passing and interpreter aligned with IR instructions on main.

RishabhRD:
NOTE: the additional commits being shown here are due to merging main branch as interpreter branch didn't had those changes and there were CI failures that needed updated changes from main.

I should have updated interpreter branch first and merged interpreter branch to this, then this would not be confusing. Will take care of that going forward.

PR #195, opened 2026-06-08
Adds fat arrow operators to given conformance declarations with context parameters

### LINKED ISSUE #194: Syntax of given conformance declarations with context parameters
In https://github.com/hylo-lang/hylo-new/blob/main/Docs/State-of-Generics.md, given conformance declarations with context parameters have the syntax:
```
given <{context parameters}> => {conformance declarations}
```
while in `Parser.swift` and in the standard library, given conformance declarations with context parameters have the syntax:
```
given <{context parameters}> {conformance declarations}
```

I am guessing that the version with the fat arrow in the docs is just out of date, but I wanted to check, because if the fat arrow is required then (at least) `Parser.swift` and the standard library need to be updated.

simonvarey:
Adds fat arrow operators to given conformance declarations with context parameters. Given conformance declarations with context parameters now have the syntax:
```
'given' '<' generic-parameters where-clause? '>' '=>' expression
```

This PR adds '=>' as an operator token type. Closes #194

### DIFF
--- a/Sources/FrontEnd/Parser/Lexer.swift
+++ b/Sources/FrontEnd/Parser/Lexer.swift
@@ -222,7 +222,7 @@ public struct Lexer: IteratorProtocol, Sequence {
     case "&": tag = .ampersand
     case "=": tag = .assign
     case "->": tag = .arrow
-    case "=>": tag = .fatarrow
+    case "=>": tag = .thickArrow
     case "==": tag = .equal
     default: tag = .operator
     }
--- a/Sources/FrontEnd/Parser/Lexer.swift
+++ b/Sources/FrontEnd/Parser/Lexer.swift
@@ -222,8 +222,8 @@ public struct Lexer: IteratorProtocol, Sequence {
     case "&": tag = .ampersand
     case "=": tag = .assign
     case "->": tag = .arrow
-    case "==": tag = .equal
     case "=>": tag = .fatarrow
+    case "==": tag = .equal
     default: tag = .operator
     }
     return .init(tag: tag, site: span(start ..< position))
--- a/Sources/FrontEnd/Parser/Lexer.swift
+++ b/Sources/FrontEnd/Parser/Lexer.swift
@@ -223,6 +223,7 @@ public struct Lexer: IteratorProtocol, Sequence {
     case "=": tag = .assign
     case "->": tag = .arrow
     case "==": tag = .equal
+    case "=>": tag = .fatarrow
     default: tag = .operator
     }
     return .init(tag: tag, site: span(start ..< position))
--- a/Sources/FrontEnd/Parser/Token.swift
+++ b/Sources/FrontEnd/Parser/Token.swift
@@ -58,6 +58,7 @@ public struct Token: Hashable, Sendable {
     case assign
     case conversion
     case equal
+    case fatarrow
     case star
     case `operator`
 
--- a/Sources/FrontEnd/Parser/Token.swift
+++ b/Sources/FrontEnd/Parser/Token.swift
@@ -58,7 +58,7 @@ public struct Token: Hashable, Sendable {
     case assign
     case conversion
     case equal
-    case fatarrow
+    case thickArrow
     case star
     case `operator`
 
--- a/Tests/CompilerTests/negative/generic-requirement.diagnostics.expected
+++ b/Tests/CompilerTests/negative/generic-requirement.diagnostics.expected
@@ -1,3 +1,3 @@
 negative/generic-requirement.hylo:13.1-6: error: no implementation of 'f(x:)'
-given<T> S<T> is P {
+given<T> => S<T> is P {
 ~~~~~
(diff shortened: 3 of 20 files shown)


Sources/FrontEnd/Parser/Lexer.swift
```diff
@@ -225,6 +225,7 @@ public struct Lexer: IteratorProtocol, Sequence {
     case "=": tag = .assign
     case "->": tag = .arrow
     case "==": tag = .equal
+    case "=>": tag = .fatarrow
```
kyouko-taiga:
```suggestion
    case "->": tag = .arrow
    case "=>": tag = .fatarrow
    case "==": tag = .equal
```

Sources/FrontEnd/Parser/Token.swift
```diff
@@ -60,6 +60,7 @@ public struct Token: Hashable, Sendable {
     case assign
     case conversion
     case equal
+    case fatarrow
```
kyouko-taiga:
```suggestion
    case thickArrow
```

Tests/CompilerTests/negative/generic-requirement.diagnostics.expected
```diff
@@ -1,3 +1,3 @@
 negative/generic-requirement.hylo:13.1-6: error: no implementation of 'f(x:)'
-given<T> S<T> is P {
+given<T> => S<T> is P {
```
kyouko-taiga:
```suggestion
given <T> => S<T> is P {
```

Tests/CompilerTests/negative/generic-requirement.hylo
```diff
@@ -10,7 +10,7 @@ given Void is P {
 }
 
 struct S<T> {}
-given<T> S<T> is P {
+given<T> => S<T> is P {
```
kyouko-taiga:
```suggestion
given <T> => S<T> is P {
```

kyouko-taiga (approved):
Very nice, thanks.

I conclude from this patch that you prefer the think arrow @simonvarey? I personally think it looks good in the files you have modified, as it helps distinguishing what's context and what's being declared.

I'll try to bring up this change to the next core meeting and unless there are strong objections we'll merge your work.

simonvarey:
> I conclude from this patch that you prefer the think arrow @simonvarey? I personally think it looks good in the files you have modified, as it helps distinguishing what's context and what's being declared.
> 
I'm not sure I've grasped given declarations well enough to have a valuable opinion here, but I agree that the arrow makes it clearer that the declared conformance is conditional. My concern would be that `given <T is A> => T is B` might be read as 'given T is A then T is B' when (if I understand correctly) it should be read as 'it is given that if T is A then T is B'

kyouko-taiga:
> My concern would be that given <T is A> => T is B might be read as 'given T is A then T is B' when (if I understand correctly) it should be read as 'it is given that if T is A then T is B'

It sounds to me that the second formulation is simply a more verbose way of saying the first. Am I missing something?

tothambrus11:
I think the difference may be about how `given` is interpreted. `given (conditional-conformance)` vs `given A, then B`. The first interpretation is what's actually happening, we add a conditional conformance as a "given'. The second interpretation happens to makes perfect English grammatical sense, but it suggests that `given A` is some special syntax for "assuming A".

simonvarey:
@tothambrus11 that is what I was thinking. The issue might be more with the term `given` than the thick arrow itself. (This might be too much bikeshedding, but I would normally read "given ..." as "supposing ..." rather than "this gives ...".)

kyouko-taiga:
This thread may not the best place to re-open a debate about something that's been litigated too much already in my opinion but here is how one can read `given` in a way that makes grammatical and mathematical sense.

In a formal setting, one can write "given some variable `x` of type `T`, the expression `e(x)` is blah". Another way to say the same thing would be "let `x` be a variable of type `T`, the expression `e(x)` is blah."

It is fine if one wants to read either "let" or "given" as "supposing". In both cases, we want introduce some symbols having particular properties in the rest of a sentence that, typically, only makes sense in the **context** of these symbols. So I don't see any issue reading `given T is P ; f()`  as "supposing `T is P` then `f()` is blah".

A conformance is just a function. The conditions of a conformance are just its parameters. When we declare `T is P` we're actually declaring a function (really, a subscript) `() -> P<T>` and when we say `T is Q => T is P` we're actually declaring a function `Q<T> -> P<T>`. That is why an arrow makes sense. These functions are not exactly like ordinary function because we let the compiler infer their arguments rather than requiring them to be passed explicitly. That is why we use a thick arrow: it marks the distinction and reminds us the observation that [functions are implications](https://en.wikipedia.org/wiki/Curry–Howard_correspondence).

In Hylo, `given` introduce values in the implicit context. We can introduce any value we want, not only conformances. For example:

```hylo
fun f<where let w: Int>() { print(w) }
given Int = 42
f() // prints 42
```

We have, perhaps too prematurely, sprinkled *a lot* of sugar and special syntax around conformance declarations. But at the end of the day we are simply declaring (context) functions in the implicit context. In other words, this

```
given T is Q => T is P { ... }
rest()
```

means roughly the following, even though it is not expressible in exactly that way in surface syntax (although it's not currently recognized by the parser but should eventually)

```
given Q<T> => P<T> = ...
rest()
```

Both programs can be read as "suppose there exist a function from `Q<T>` to `P<T>`, then `rest` is blah".

tothambrus11:
@simonvarey Thanks for the work, we can push this now (we were waiting for LLVM lowering to get in).

PR #199, opened 2026-06-11
Add execution testing with standard library lowering

tothambrus11:
New testing options:
- `//! stage:run assert-exit-code:42 assert-abort:true` (defaults: exit code 0 and no abort is expected)
- `test-case.stdout.expected` - when present, contents is compared with the observed stdout (with line endings normalized to avoid CRLF issues on Windows)
- compiling the standard library unless `no-std`
- saving stdlib compilation artifacts on failure
- saving executable on failure (you can then re-invoke it from a debugger manually)


tothambrus11:
@kyouko-taiga You might want to merge this to the LLVM branch. The CI fails should be fixed by a more complete LLVM lowering + fixing of the incorrect diagnostics during the IR transformation of the standard library.

Sources/Utilities/StringProtocol+Extensions.swift
```diff
@@ -29,6 +29,11 @@ extension StringProtocol {
     .init(filter({ (a) in a != c }))
   }
 
+  /// Returns self with unix-style line endings.
```
kyouko-taiga:
```suggestion
  /// Returns `self` with unix-style line endings.
```

Tests/BackEndTests/SimpleFunctionEmitterTest.swift
```diff
@@ -56,6 +56,106 @@ final class SimpleFunctionEmitterTest: XCTestCase {
     XCTAssertEqual(output.trimming(while: \.isWhitespace), "")
   }
 
+  func testLoweringStdlib() async throws {
+
```
kyouko-taiga, line 60:
```suggestion
```

Tests/BackEndTests/SimpleFunctionEmitterTest.swift
```diff
@@ -56,6 +56,106 @@ final class SimpleFunctionEmitterTest: XCTestCase {
...
+    let m0 = d.program.demandModule("M0")
+
+    d.program[m0].addDependency(Module.standardLibraryName)
+
+    _ = d.program[m0].addSource(
+      """
+      fun create() -> Int32 {
+        Int32()
+      }
+      """)
```
kyouko-taiga, line 73:
```suggestion
    _ = d.program[m0].addSource("fun create() -> Int32 { 0 }"))
```

Why is the function called `create`?

Tests/BackEndTests/SimpleFunctionEmitterTest.swift
```diff
@@ -56,6 +56,106 @@ final class SimpleFunctionEmitterTest: XCTestCase {
...
+
+    await d.program.assignScopes(m0)
+    try assertNoDiagnostics(in: d.program)
+
+    d.program.assignTypes(m0, loggingInferenceWhere: { _, _ in false })
```
kyouko-taiga:
```suggestion
    d.program.assignTypes(m0, loggingInferenceWhere: { (_, _) in false })
```

Tests/BackEndTests/SimpleFunctionEmitterTest.swift
```diff
@@ -56,6 +56,106 @@ final class SimpleFunctionEmitterTest: XCTestCase {
...
+    try assertNoDiagnostics(in: d.program)
+
+    d.program.lower(m0)
+    d.program.lower(d.program.modules[Module.standardLibraryName]!.identity)
+
+    let stdlib = d.program.modules[Module.standardLibraryName]!.identity
+    print("STD LIB DIAGNOSTICS AFTER LOWERING: \(Array(d.program[stdlib].diagnostics).count)")
```
kyouko-taiga, line 85: I'm surprised to see a print in all execution paths.

Tests/BackEndTests/SimpleFunctionEmitterTest.swift
```diff
@@ -56,6 +56,106 @@ final class SimpleFunctionEmitterTest: XCTestCase {
...
+    for diag in d.program[stdlib].diagnostics {
+      print("  - \(diag)")
+    }
+    print("ALL DIAGNOSTICS AFTER LOWERING: \(Array(d.program.diagnostics).count)")
+
+    var p1 = TreePrinter(program: d.program)
+    XCTAssertEqual(
```
kyouko-taiga, line 92: Why do we have to test the output of IR here? I don't understand what this test case is about.

Presumably we'll do IR testing using the regular testing architecture so it seems this function is adding a lot of code that is destined to be removed.

Tests/BackEndTests/SimpleFunctionEmitterTest.swift
```diff
@@ -56,6 +56,106 @@ final class SimpleFunctionEmitterTest: XCTestCase {
...
+    d.program.applyTransformationPasses(m0)
+    try assertNoDiagnostics(in: d.program)
+    
+    d.program.applyTransformationPasses(d.program.modules[Module.standardLibraryName]!.identity)
+    try assertNoDiagnostics(in: d.program)
+
+    _ = try d.compileToLLVM(m0)
+    let llvmIR = try XCTUnwrap(d.llvmIR(of: m0))
+
+    XCTAssertEqual(
```
kyouko-taiga, line 117: Same remark as the one I made about Hylo IR.

Tests/BackEndTests/SimpleFunctionEmitterTest.swift
```diff
@@ -56,6 +56,106 @@ final class SimpleFunctionEmitterTest: XCTestCase {
...
+      declare void @R0U7Int32S10F06initlT01tR516selfsTK1tR5(ptr)
+      """)
+  }
+
+}
+
+/// Asserts that `program` has no diagnostics.
+func assertNoDiagnostics(in program: Program, file: StaticString = #filePath, line: UInt = #line)
+  throws
+{
```
kyouko-taiga:
```suggestion
func assertNoDiagnostics(
  in program: Program, file: StaticString = #filePath, line: UInt = #line
) throws {
```

Tests/CompilerTests/CompilerTests.swift
```diff
@@ -23,10 +26,21 @@ final class CompilerTests: XCTestCase {
     /// The manifest of the test.
     let manifest: Manifest
 
+    /// What the program is expected to print to stdout when `self` is a run-stage test.
```
kyouko-taiga, line 29:
```suggestion
    /// The expected contents of `stdout` when `self` is a run-stage test.
```
What does `self` is a run-stage test" means?

Tests/CompilerTests/CompilerTests.swift
```diff
@@ -92,25 +106,49 @@ final class CompilerTests: XCTestCase {
...
     var llvmIR: String?
 
+    /// The lowered IR of the standard library, if any.
+    var stdlibRawIR: String?
+
+    /// The transformed IR of the standard library, if any.
+    var stdlibTransformedIR: String?
+
+    /// The compiled IR artifact of the standard library, if any.
+    var stdlibLLVMIR: String?
```
kyouko-taiga, line 116: Why are we keeping this around? It looks like YAGNI.

Tests/CompilerTests/CompilerTests.swift
```diff
@@ -92,25 +106,49 @@ final class CompilerTests: XCTestCase {
...
+    /// The standard output of the executed program, if any.
+    var standardOutput: String?
+
     /// Saves the artifacts into test-case-level observation files of `test`.
     func save(into test: TestDescription) throws {
       if let rawIR {
         try test.saveTestCaseLevelObservation(rawIR, tag: "raw-ir")
       }
+      if let stdlibRawIR {
+        try test.saveTestCaseLevelObservation(stdlibRawIR, tag: "stdlib.raw-ir")
```
kyouko-taiga, line 132: I don't like the additional code paths made for standard library. It creates too much code for very few test cases. If we want to test the compilation artifacts of a different module we should make it more general. If we want to special case the standard library we can create a different test driver.

Tests/CompilerTests/CompilerTests.swift
```diff
@@ -136,19 +174,19 @@ final class CompilerTests: XCTestCase {
     case missingExecutableOutput
 
     /// The test failed because of a compilation error.
-    case compilationError(String)
+    case compilationError(message: String)
```
kyouko-taiga, line 177: The label doesn't add much IMO.

kyouko-taiga (changes requested):
If you want I can cherry pick the parts that I like instead of iteratively modify this PR until it gets merged. Otherwise, I would mostly like to reduce the amount of changes made specifically to test the standard library.

Tests/BackEndTests/SimpleFunctionEmitterTest.swift
tothambrus11: No particular reason, it just exercises the initializer of a struct.

Tests/CompilerTests/CompilerTests.swift
tothambrus11: Ah, it's collapsed, so it may be confusing for the review. This is part of `TestDescription`, which represents a test case at a given location. Its manifest contains its stage information, which can be "run".

Tests/CompilerTests/CompilerTests.swift
tothambrus11: This is useful during the development of LLVM compilation, so that you can easily dump the resulting LLVM IR from a test case. You could alternatively use the CLI, but in my experience that's more cumbersome (requires setting up CLI arguments manually for each different we try).

This, as any other artifact gets saves either on a failure of a stage (not applicable to LLVM lowering, crashes are excluded), or when `artifactsAreSavedOnSuccess` is true (I usually keep that on during development).

Tests/CompilerTests/CompilerTests.swift
tothambrus11: Yeah, we could make it more general at some pont, but I think we shouldn't try to make a build system until we have a working compiler. Making a general solution is much more complicated.

Being able to print the IR of the standard library should be useful for development purposes until we get all langauge features lowered to LLVM. I'm fine with removing it later.

Tests/CompilerTests/CompilerTests.swift
tothambrus11: In `compilationError`, I agree that `message` doesn't make much of a difference.
I wanted to make a distinction in `case .invalidTestDescription(let message):` between whether I'm passing the test description that was invalid or the error message that describes why it was invalid. I considered updating its description but adding a name seemed simpler. Then I updated the rest of the cases for consistency.

Tests/BackEndTests/SimpleFunctionEmitterTest.swift
tothambrus11: Yeah, exactly. 

This test is not a proper test (it doesn't pass). It's a triage that shows the miscompilation I mentioned on Slack in PM. The tests in this file are minimalized version of running our compiler pipeline without all the abstractions and magic of the file-based testing framework. This aims to make it easier to debug ongoing compilation issues where we may want to be more flexible with how we exactly set up the driver and be sure the bug is not in the testing framework setup.

The current problem: The standard library didn't succeed during IR transformation, it resulted in the following diagnostics:

```
/home/ambrus/hylo-new-2/Tests/CompilerTests/CompilerTests.swift:259: error: CompilerTests.test_positive_assert_exit_code_42 : failed - Compilation failure:
program contains one or more errors:
../../StandardLibrary/Sources/Core/Numbers/AdditiveArithmetic.hylo:13.12-25: error: no given instance of 'Movable<Self>' in this scope
    &self = self + other
           ~~~~~~~~~~~~~
../../StandardLibrary/Sources/Core/Numbers/AdditiveArithmetic.hylo:13.25: error: no given instance of 'Deinitializable<Self>' in this scope
    &self = self + other
                        ^
../../StandardLibrary/Sources/Core/Numbers/AdditiveArithmetic.hylo:25.12-25: error: no given instance of 'Movable<Self>' in this scope
    &self = self - other
           ~~~~~~~~~~~~~
../../StandardLibrary/Sources/Core/Numbers/AdditiveArithmetic.hylo:25.25: error: no given instance of 'Deinitializable<Self>' in this scope
    &self = self - other
                        ^
../../StandardLibrary/Sources/Core/Numbers/Int.hylo:68.7-33: error: no given instance of 'Deinitializable<{word, i1}>' in this scope
      return .new(value: result)
      ~~~~~~~~~~~~~~~~~~~~~~~~~~
../../StandardLibrary/Sources/Core/Numbers/Int.hylo:77.7-33: error: no given instance of 'Deinitializable<{word, i1}>' in this scope
      return .new(value: result)
      ~~~~~~~~~~~~~~~~~~~~~~~~~~
../../StandardLibrary/Sources/Core/Numbers/Numeric.hylo:13.12-25: error: no given instance of 'Movable<Self>' in this scope
    &self = self * other
           ~~~~~~~~~~~~~
../../StandardLibrary/Sources/Core/Numbers/Numeric.hylo:13.25: error: no given instance of 'Deinitializable<Self>' in this scope
    &self = self * other
```

Tests/CompilerTests/CompilerTests.swift
kyouko-taiga: I understand but I don't like that your PR is doing too much stuff. Rushing a hacky solution that's not indispensable in anticipation for a harder one that we'll have to make down the line is a bad bet. It's just code churn.

tothambrus11: Feel free to take whatever parts are useful for you.

PR #201, opened 2026-06-15
refactor: cache standard library type classification in Typer

### LINKED ISSUE #118: Refactor checking of standard library types, caching
In `isStandardLibraryIntegerType`:
```
Program.StandardLibraryEntity.allIntegerTypes.contains { standardLibraryType($0) == d }
```
> This has a cost of about 10 hasmap lookups. We could add a cached version of the list of standard library integer type declarations in Program. It could be worth the extra top-level stored property in Program.

@kyouko-taiga wrote:

What if we tried something like this:

```swift
private mutating func isStandardLibraryFloatingPointType(_ t: AnyTypeIdentity) -> Bool {
  isStandardLibraryType(t, in: [.float32, .float64])
}

private mutating func isStandardLibraryType(
  _ t: AnyTypeIdentity, in ts: [Program.StandardLibraryEntity]
) -> Bool {
  guard program.containsStandardLibrary else { return false }
  let u = program.types.dealiased(t)
  return ts.contains(where: { (e) in standardLibraryType(e) == u  })
}
```

I didn't check but I strongly suspect the loop will get unrolled and inlined. The array will get statically allocated, if not completely inlined after unrolling.

IIRC you mentioned that this method is expensive elsewhere and I agree. I think we can cache the result of `standardLibraryType(_:)` in the typer. But maybe you'd like to do that as a separate patch.

_Originally posted by @kyouko-taiga in https://github.com/hylo-lang/hylo-new/pull/116#discussion_r3148267735_

mvanhorn:
## Summary

Unifies the standard library type classification checks in `Typer` behind a single memoized helper, as proposed in #118.

- Adds a `standardLibraryEntityToType` cache to `Typer.Memos`, keyed by `Program.StandardLibraryEntity`, alongside the existing lazy caches.
- Memoizes `standardLibraryType(_:)` against that cache so repeated resolutions of the same entity are O(1) instead of re-resolving the declaration to a type on every call.
- Introduces a private `isStandardLibraryType(_:in:)` helper that guards on `program.containsStandardLibrary`, dealiases the type once, and checks membership using the now-cached lookups.
- Reimplements `isStandardLibraryIntegerType` and `isStandardLibraryFloatingPointType` as thin wrappers over the new helper.

## Why this matters

As noted in #118, classifying a type as a standard library integer or floating point type re-ran a per-call membership scan, with each `standardLibraryType(_:)` call resolving the entity to a type from scratch (roughly ten hashmap lookups per check). The maintainer's proposed approach was to unify these into a single helper backed by a memoized cache that works like the other cached methods on `Typer.Memos`, mirroring the lazy-cache pattern already used throughout the typer. This change follows that approach: the cache turns repeated entity resolutions into O(1) lookups while keeping the classification logic identical.

## Testing

The change is behavior-preserving: the same entities (`[.int, .int32, .int64]` for integers, `[.float32, .float64]` for floats) are checked against the same dealiased type, and the `program.containsStandardLibrary` guard is preserved so partial-standard-library programs still return `false` without crashing. `swiftc -parse` on the modified file passes; a full `swift build` requires the LLVM toolchain headers, which are not available in this environment.

Fixes #118

### DIFF
--- a/Sources/FrontEnd/Typer/Typer.swift
+++ b/Sources/FrontEnd/Typer/Typer.swift
@@ -4639,7 +4639,7 @@ public struct Typer {
     guard program.containsStandardLibrary else { return false }
 
     let u = program.types.dealiased(t)
-    return ts.contains { u == standardLibraryType($0) }
+    return ts.contains(where: { (v) in standardLibraryType(v) == u })
   }
 
   /// Returns the type of the given standard library entity.
--- a/Sources/FrontEnd/Typer/Typer.swift
+++ b/Sources/FrontEnd/Typer/Typer.swift
@@ -102,6 +102,9 @@ public struct Typer {
     /// The cache of `Typer.declaredType(of:)` for predefined givens.
     fileprivate var predefinedGivens: [Given: AnyTypeIdentity]
 
+    /// The cache of `Typer.standardLibraryType(_:)`.
+    fileprivate var standardLibraryEntityToType: [Program.StandardLibraryEntity: AnyTypeIdentity]
+
     /// The cache of `Typer.canDeriveCoercion(_:_:applying:)`.
     fileprivate var canDeriveCoercion: [Given: [TypePair: (Bool, Bool)]]
 
@@ -121,6 +124,7 @@ public struct Typer {
       self.witnessToAliases = [:]
       self.declarationToTentativeType = [:]
       self.predefinedGivens = [:]
+      self.standardLibraryEntityToType = [:]
       self.canDeriveCoercion = [:]
     }
 
@@ -4616,15 +4620,7 @@ public struct Typer {
   /// The module containing the standard library must have been loaded in the `self.program`, or
   /// `self.module` is the standard library.
   private mutating func isStandardLibraryIntegerType(_ t: AnyTypeIdentity) -> Bool {
-    guard program.containsStandardLibrary else { return false }
-    switch program.types.dealiased(t) {
-    case standardLibraryType(.int),
-        standardLibraryType(.int32),
-        standardLibraryType(.int64):
-      return true
-    default:
-      return false
-    }
+    isStandardLibraryType(t, in: [.int, .int32, .int64])
   }
 
   /// Returns `true` iff `t` is a standard library floating point type (e.g., `Hylo.Float32`).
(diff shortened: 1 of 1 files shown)


Sources/FrontEnd/Typer/Typer.swift
```diff
@@ -4616,30 +4620,29 @@ public struct Typer {
...
-        standardLibraryType(.float64):
-      return true
-    default:
-      return false
+
+    let u = program.types.dealiased(t)
+    for e in ts {
+      if u == standardLibraryType(e) { return true }
     }
+    return false
```
```suggestion
    return ts.contains(u)
```
Does that work or are we using an overloaded form of `==`?

kyouko-taiga (changes requested):
Thanks a lot for this change!

I left a small comment about the loop in `isStandardLibraryType` which I think can be improved a little.

Sources/FrontEnd/Typer/Typer.swift
mvanhorn: Good call on simplifying this. The literal `return ts.contains(u)` doesn't type-check, though - `ts` is `[Program.StandardLibraryEntity]` while `u` is an `AnyTypeIdentity`, so each entity still has to go through `standardLibraryType(_:)` before the comparison (and that method is `mutating`, since it populates the cache, so the types can't be precomputed cheaply).

On the `==`: it's the synthesized one. `AnyTypeIdentity` is `Hashable, Sendable` with no custom `==`, so this is plain value equality, not an overload.

I kept your intent by switching the loop to `contains(where:)`, which preserves the early-out and uses that same equality:
```swift
return ts.contains { u == standardLibraryType($0) }
```
Pushed in efdebf5. The `FrontEnd` target builds with the change.

Sources/FrontEnd/Typer/Typer.swift
```diff
@@ -4616,30 +4620,26 @@ public struct Typer {
...
 
   /// Returns `true` iff `t` is a standard library floating point type (e.g., `Hylo.Float32`).
   ///
   /// The module containing the standard library must have been loaded in the `self.program`, or
   /// `self.module` is the standard library.
   private mutating func isStandardLibraryFloatingPointType(_ t: AnyTypeIdentity) -> Bool {
+    isStandardLibraryType(t, in: [.float32, .float64])
+  }
+
+  /// Returns `true` iff `t` is one of the given standard library types.
```
```suggestion
  /// Returns `true` iff `t` is one of the standard library types `ts`.
```

tothambrus11:
Thanks for the PR, looks good to me!

It appears to me that you may be using AI to phrase your PR descriptions and comments. If that's true, please avoid doing so in the future to preserve accuracy & intent and avoid unnecessary text.

Sources/FrontEnd/Typer/Typer.swift
```diff
@@ -4616,30 +4620,26 @@ public struct Typer {
...
-    switch program.types.dealiased(t) {
-    case standardLibraryType(.float32),
-        standardLibraryType(.float64):
-      return true
-    default:
-      return false
-    }
+
+    let u = program.types.dealiased(t)
+    return ts.contains { u == standardLibraryType($0) }
```
```suggestion
    return ts.contains(where: { (v) in standardLibraryType(v) == u })
```

kyouko-taiga (approved):
LGTM

mvanhorn:
Appreciate you taking this, @kyouko-taiga. Memoizing the standard library type classification in Typer behind one helper reads much cleaner.

mvanhorn:
Appreciate the merge, @kyouko-taiga. Memoizing the standard library type classification behind one helper should keep those Typer checks tidy and fast.

PR #203, opened 2026-06-15
Fix Windows CI by upgrading Swift version to 6.3.2; Fix CI not running on external PRs

tothambrus11:


tothambrus11:
It seems the Windows CI breakage is fixed in Swift 6.3.

PR #208, opened 2026-06-24
Use platform's default relocation model in driver

tothambrus11:
Previously, the platform's default relocation model was only set by the command line, so the tests didn't apply it. Changed it so that the driver already applies this default.


kyouko-taiga (approved):
LGTM

PR #210, opened 2026-06-25
Partial artifact matching based test suite

tothambrus11:
You can add a `test-case.<artifact-tag>.expected` file besides your `test-case.hylo` to assert the contents of a compilation artifact. Valid artifacts are:
- `*.raw-ir.expected`
- `*.transformed-ir.expected`
- `*.llvm-ir.expected`

An artifact is composed of a set of contiguous sections, delimited by empty lines. E.g. each function in Hylo IR and LLVM IR are their own sections.

It is sufficient to specify a subset of sections of the actual artifact. The expected section is matched with the observed section having the closest first line. Then they are compared for equality.

### DIFF
--- a/Sources/Utilities/Sequence+Extensions.swift
+++ b/Sources/Utilities/Sequence+Extensions.swift
@@ -65,4 +65,9 @@ extension Sequence {
     joinedString(separator: separator) { String(describing: $0) }
   }
 
+  /// Returns the element of `self` that minimizes `f`, or `nil` if `self` is empty.
+  public func min<R: Comparable>(by keyPath: (Element) -> R) -> Element? {
+    self.min(by: { (a, b) in keyPath(a) < keyPath(b) })
+  }
+
 }
--- a/Sources/Utilities/Sequence+Extensions.swift
+++ b/Sources/Utilities/Sequence+Extensions.swift
@@ -65,9 +65,13 @@ extension Sequence {
     joinedString(separator: separator) { String(describing: $0) }
   }
 
-  /// Returns the element of `self` that minimizes `f`, or `nil` if `self` is empty.
-  public func min<R: Comparable>(by keyPath: (Element) -> R) -> Element? {
-    self.min(by: { (a, b) in keyPath(a) < keyPath(b) })
+  /// Returns the element in `self` with the smallest value measured by `p`, if any.
+  /// 
+  /// If there is more than one smallest element, it is unspecified which one is returned.
+  ///
+  /// - Complexity: O(n)
+  public func min<R: Comparable>(measuredBy p: (Element) -> R) -> Element? {
+    self.min(by: { (a, b) in p(a) < p(b) })
   }
 
 }
--- a/Sources/Utilities/StringProtocol+Extensions.swift
+++ b/Sources/Utilities/StringProtocol+Extensions.swift
@@ -29,6 +29,11 @@ extension StringProtocol {
     .init(filter({ (a) in a != c }))
   }
 
+  /// Returns `self` with unix-style line endings.
+  public func normalizedLineEndings() -> String {
+    replacingOccurrences(of: "\r\n", with: "\n")  // Windows
+  }
+
 }
 
 extension StringProtocol where SubSequence == Substring {
--- a/Tests/CompilerTests/CompilerTests.swift
+++ b/Tests/CompilerTests/CompilerTests.swift
@@ -3,9 +3,10 @@ import Foundation
 import FrontEnd
 import StandardLibrary
 import Utilities
+import StableCollections
 import XCTest
 
-/// `true` iff intermediate compilation artifacts must be saved for successful tests.
+/// `true` iff intermediate compilation artifacts shall be saved for successful tests.
 private let alwaysSaveArtifacts: Bool = false
 
 /// The driver for generated compiler tests.
@@ -63,10 +64,11 @@ final class CompilerTests: XCTestCase {
     /// Returns where to save the generated executable upon failure.
     func executableDestination() -> URL {
       let suffix = Host.nativeExecutableSuffix
+      let tag = ArtifactTag.executable
       if isPackage {
-        return root.appending(component: ".executable\(suffix)")
+        return root.appending(component: ".\(tag).observed\(suffix)")
       } else {
-        return root.deletingPathExtension().appendingPathExtension("executable\(suffix)")
+        return root.deletingPathExtension().appendingPathExtension("\(tag).observed\(suffix)")
       }
     }
 
@@ -80,19 +82,63 @@ final class CompilerTests: XCTestCase {
 
   }
 
-  /// The intermediate artifacts of a module's compilation.
+  /// The type of a compilation artifact.
+  enum ArtifactTag: String, CustomStringConvertible, Comparable {
+
+    /// Hylo IR immediately after lowering, without any processing.
+    case rawIR = "raw-ir"
+
+    /// Hylo IR after mandatory transformation passes.
+    case refinedIR = "refined-ir"
+
+    /// LLVM IR.
+    case llvmIR = "llvm-ir"
+
+    /// The executable generated from the compiled program.
+    case executable = "executable"
+
+    /// All textual artifacts of the compilation and testing pipeline.
+    static var allTextual: [ArtifactTag] {
+      [.rawIR, .refinedIR, .llvmIR]
+    }
+
+    /// The minimum stage of compilation to produce this artifact.
+    var minimumStage: Manifest.Stage {
+      switch self {
+      case .rawIR: .lowering
+      case .refinedIR: .lowering
+      case .llvmIR: .llvm
+      case .executable: .execution
+      }
+    }
+
+    /// The textual description of `self`.
+    var description: String { rawValue }
+
+    /// The order of `self` among all artifact tags, which is used to sort them.
+    var order: Int {
+      switch self {
+      case .rawIR: 0
+      case .refinedIR: 1
+      case .llvmIR: 2
+      case .executable: 3
+      }
+    }
+
+    /// Returns true iff `l` is ordered before `r`.
+    static func < (l: CompilerTests.ArtifactTag, r: CompilerTests.ArtifactTag) -> Bool {
+      l.order < r.order
+    }
+
+  }
+
+  /// The intermediate artifacts of a module's compilation and testing.
   ///
   /// Members shall be set after the corresponding stage is completed.
   private struct Artifacts {
 
-    /// The lowered IR of the compiled module, if any.
-    var rawIR: String?
-
-    /// The transformed IR of the compiled module, if any.
-    var transformedIR: String?
-
-    /// The compiled IR artifact of the tested module, if any.
-    var llvmIR: String?
+    /// The textual artifacts of the compilation and testing pipeline, keyed by their tag.
+    private(set) var textual = SortedDictionary<ArtifactTag, String>()
 
     /// The URL of the generated executable residing in a temporary directory, if any.
     ///
@@ -102,18 +148,19 @@ final class CompilerTests: XCTestCase {
 
     /// Saves the artifacts into test-case-level observation files of `test`.
     func save(into test: TestDescription) throws {
-      if let a = rawIR {
-        try test.saveTestCaseLevelObservation(a, tag: "raw-ir")
+      for (tag, artifact) in textual {
+        try test.saveTestCaseLevelObservation(artifact, tag: tag.rawValue)
       }
-      if let a = transformedIR {
-        try test.saveTestCaseLevelObservation(a, tag: "transformed-ir")
-      }
-      if let a = llvmIR {
-        try test.saveTestCaseLevelObservation(a, tag: "ll")
-      }
-      if let a = executable {
+
+      if let a: URL = executable {
+        try? FileManager.default.removeItem(at: test.executableDestination())
         try FileManager.default.copyItem(at: a, to: test.executableDestination())
       }
+    }
+
+    /// Records the textual `artifact` as the artifact tagged `tag`.
+    mutating func record(_ artifact: String, for tag: ArtifactTag) {
+      textual[tag] = artifact
     }
 
   }
@@ -182,29 +229,47 @@ final class CompilerTests: XCTestCase {
 
   /// Compiles and runs `input` expecting no compilation error.
   func positive(_ input: TestDescription) async throws {
-    // Compile the input up until the specified stage.
-    let o = input.manifest.optimizations != .none
-    let r = try await compile(input, withOptimizations: o)
-    try assertSansError(r.driver.program)
+    await loggingSourceOnFailure(root: input.root) {
+      // Compile the input up until the specified stage.
+      let o = input.manifest.optimizations != .none
+      let r = try await compile(input, withOptimizations: o)
+      try assertSansError(r.driver.program)
 
-    // Should an executable be tested?
-    if input.manifest.stage == .execution {
-      let e = try XCTUnwrap(r.artifacts.executable)
-      let x = try Process.execute(e)
-      assertExitStatus(x, describedBy: input)
+      // Should an executable be tested?
+      if input.manifest.stage == .execution {
+        let e = try XCTUnwrap(r.artifacts.executable)
+        let x = try Process.execute(e)
+        assertExitStatus(x, describedBy: input)
+      }
+
+      if testFailed || alwaysSaveArtifacts { try r.artifacts.save(into: input) }
     }
-
-    if testFailed || alwaysSaveArtifacts { try r.artifacts.save(into: input) }
   }
 
   /// Compiles `input` expecting at least one compilation error.
   func negative(_ input: TestDescription) async throws {
-    let r = try await compile(input, withOptimizations: false)
-    let m = "program compiled but an error was expected.\nSource: \(input.root.path)\n"
-    XCTAssert(r.driver.program.containsError, m)
-    assertExpectations(r.expectedDiagnostics, r.driver.program.diagnostics)
+    await loggingSourceOnFailure(root: input.root) {
+      let r = try await compile(input, withOptimizations: false)
+      let m = "program compiled but an error was expected.\nSource: \(input.root.path)\n"
+      XCTAssert(r.driver.program.containsError, m)
+      assertExpectations(r.expectedDiagnostics, r.driver.program.diagnostics)
 
-    if testFailed || alwaysSaveArtifacts { try r.artifacts.save(into: input) }
+      if testFailed || alwaysSaveArtifacts { try r.artifacts.save(into: input) }
+    }
+  }
+
+  /// Executes `action`, and logs the test case's `root` on test failure for easier diagnostics.
+  ///
+  /// Also logs and catches any error thrown by `action`, failing the test case.
+  private func loggingSourceOnFailure<T>(root: URL, _ action: () async throws -> T) async {
+    do {
+      _ = try await action()
+      if testFailed {
+        XCTFail("\nSource: \(root.path)\n")
+      }
+    } catch let e {
+      XCTFail("\(e)\nSource: \(root.path)\n")
+    }
   }
 
   /// Compiles `input` and returns the resulting artifacts.
@@ -236,8 +301,8 @@ final class CompilerTests: XCTestCase {
     var artifacts = Artifacts()
     do {
       try await compile(
-        m, until: input.manifest.stage, using: &driver,
-        accumulatingArtifactsInto: &artifacts)
+        m, until: input.manifest.stage, using: &driver, accumulatingArtifactsInto: &artifacts,
+        expectedArtifacts: input.manifest.artifactExpectations)
     } catch let e {
       try artifacts.save(into: input)
       throw e
@@ -255,7 +320,8 @@ final class CompilerTests: XCTestCase {
   /// adding compilation artifacts to `artifacts`.
   private func compile(
     _ m: Module.ID, until stage: Manifest.Stage, using driver: inout Driver,
-    accumulatingArtifactsInto artifacts: inout Artifacts
+    accumulatingArtifactsInto artifacts: inout Artifacts,
+    expectedArtifacts: SortedDictionary<ArtifactTag, String>
   ) async throws {
     // Exit if there are parsing errors or if the stage is set to `parsing`.
     if driver.program[m].containsError || (stage == .parsing) { return }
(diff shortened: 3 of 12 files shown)


Sources/Utilities/StringProtocol+Extensions.swift
```diff
@@ -29,6 +29,12 @@ extension StringProtocol {
     .init(filter({ (a) in a != c }))
   }
 
+  /// Returns `self` with unix-style line endings.
+  public func normalizedLineEndings() -> String {
+    replacingOccurrences(of: "\r\n", with: "\n")  // Windows
+      .replacingOccurrences(of: "\r", with: "\n")  // Old Mac
```
kyouko-taiga:
```suggestion
    self
      .replacingOccurrences(of: "\r\n", with: "\n")  // Windows
      .replacingOccurrences(of: "\r", with: "\n")  // Old Mac
```

kyouko-taiga: How old are we talking for macOS? We have pretty high OS requirements in Package.swift anyway.

Tests/CompilerTests/CompilerTests.swift
```diff
@@ -80,19 +80,34 @@ final class CompilerTests: XCTestCase {
 
   }
 
-  /// The intermediate artifacts of a module's compilation.
+  /// The type of a textual compilation artifact.
```
kyouko-taiga: Why is that restricted to textual artifacts?

Tests/CompilerTests/CompilerTests.swift
```diff
@@ -80,19 +80,34 @@ final class CompilerTests: XCTestCase {
...
+  enum ArtifactTag: String, CustomStringConvertible {
+
+    case rawIR = "raw-ir"
```
kyouko-taiga, line 89:
```suggestion
    /// Hylo IR immediately after lowering, without any processing.
    case rawIR = "raw-ir"
```

Tests/CompilerTests/CompilerTests.swift
```diff
@@ -80,19 +80,34 @@ final class CompilerTests: XCTestCase {
...
+    case transformedIR = "transformed-ir"
```
kyouko-taiga:
```suggestion

        /// Hylo IR after mandatory transformation passes.
        case refinedIR = "refined-ir"
```

Tests/CompilerTests/CompilerTests.swift
```diff
@@ -80,19 +80,34 @@ final class CompilerTests: XCTestCase {
...
+    case llvmIR = "llvm-ir"
```
kyouko-taiga, line 95:
```suggestion

        /// LLVM IR.
        case llvmIR = "llvm-ir"
```

Tests/CompilerTests/CompilerTests.swift
kyouko-taiga: It was named like that in old front-end and I think it is more descriptive than "transformed".

Tests/CompilerTests/CompilerTests.swift
```diff
@@ -80,19 +80,34 @@ final class CompilerTests: XCTestCase {
...
+    /// The IR language of the artifact.
+    var language: IRMatching.Language {
+      switch self {
+      case .rawIR: .hyloIR
+      case .transformedIR: .hyloIR
+      case .llvmIR: .llvmIR
+      }
+    }
+
+    /// The textual tag of `self`.
```
kyouko-taiga:
```suggestion
    /// The textual description of `self`.
```

Tests/CompilerTests/CompilerTests.swift
```diff
@@ -80,19 +80,34 @@ final class CompilerTests: XCTestCase {
...
-    /// The lowered IR of the compiled module, if any.
-    var rawIR: String?
-
-    /// The transformed IR of the compiled module, if any.
-    var transformedIR: String?
-
-    /// The compiled IR artifact of the tested module, if any.
-    var llvmIR: String?
+    /// The textual artifacts of the compilation and testing pipeline, keyed by their tag.
+    private(set) var textual: [ArtifactTag: String] = [:]
```
kyouko-taiga: Small dictionaries are much more expensive than arrays. I think the `Utilities` module has a `SortedDictionary`, which will be faster and have a smaller memory footprint.

Tests/CompilerTests/CompilerTests.swift
```diff
@@ -192,20 +203,30 @@ final class CompilerTests: XCTestCase {
...
+  ///
+  /// Also logs and catches any error thrown by `action`, failing the test case.
+  private func loggingSourceOnFailure<T>(root: URL, _ action: () async throws -> T) async {
+    do {
+      _ = try await action()
+      if testFailed {
+        XCTFail("\nSource: \(root.path)\n")
+      }
+    } catch let e {
+      XCTFail("\(e.localizedDescription)\nSource: \(root.path)\n")
```
kyouko-taiga: I guess you added `Source \(root.path):` so that one can copy/paste a filepath from a test report, but for the record I really do not like this change (I know it doesn't come from this PR). The name of the test case is already encoded in the test case itself, and the location is obvious from that name. Any competent IDE should be able to locate a source file from its base name.

Tests/CompilerTests/IRMatching.swift
```diff
@@ -0,0 +1,90 @@
...
+  static func comparisons(
+    expected: String, observed: String, language: Language
+  ) -> [Comparison] {
+    let expected = sections(of: expected.normalizedLineEndings())
+    let observed = sections(of: observed.normalizedLineEndings())
+
+    return expected.map { (e) in
+      let head = e[0]
+
+      // Pair with the observed section with same key or the one whose first line is most similar.
```
kyouko-taiga: I don't understand this comment, grammatically.

Tests/CompilerTests/IRMatching.swift
```diff
@@ -0,0 +1,90 @@
...
+
+/// Returns the smallest number of additions and deletions that transform `b` into `a`.
+private func distance(_ a: Substring, _ b: Substring) -> Int {
+  a.difference(from: b).count
+}
+
+extension Array {
+
+  /// Returns the element of `self` that minimizes `f`, or `nil` if `self` is empty.
+  func min(by keyPath: (Element) -> some Comparable) -> Element? {
```
kyouko-taiga:
```suggestion
  func min<T: Comparable>(by keyPath: (Element) -> T) -> Element? {
```
I will never get behind these opaque types.

kyouko-taiga: Also I think there is a similar functionality for sorting in `Utilities`. I would consider moving this function there.

Tests/CompilerTests/IRMatching.swift
```diff
@@ -0,0 +1,90 @@
...
+    self.min(by: { keyPath($0) < keyPath($1) })
```
kyouko-taiga:
```suggestion
    self.min(by: { (a, b) in keyPath(a) < keyPath(b) })
```

Tests/CompilerTests/Manifest.swift
```diff
@@ -22,6 +22,74 @@ struct Manifest {
...
 
+    /// The position of `self` in the pipeline.
+    var order: Int {
+      switch self {
+      case .parsing: return 0
+      case .typing: return 1
+      case .lowering: return 2
+      case .llvm: return 3
+      case .execution: return 4
+      }
```
kyouko-taiga, line 33:
```suggestion
      switch self {
      case .parsing: 0
      case .typing: 1
      case .lowering: 2
      case .llvm: 3
      case .execution: 4
      }
```

Just a suggestion. No strong opinion.

Tests/CompilerTests/Manifest.swift
```diff
@@ -22,6 +22,74 @@ struct Manifest {
...
+
+    /// True iff `l` is earlier in the pipeline than `r`.
+    static func < (l: Stage, r: Stage) -> Bool {
+      l.order < r.order
+    }
+
+  }
+
+  /// Expectations for artifacts of the compilation and testing pipeline.
+  struct ArtifactExpectations {
```
kyouko-taiga: Is there a way to represent expected artifacts and produced artifacts with the same data structure?

Tests/CompilerTests/README.md
```diff
@@ -7,12 +7,40 @@ A use case is either a single Hylo source file or a directory representing a pac
 A single-file test is compiled to a binary executable, just as if it was passed as an argument to `hc`.
 A package test is built according to the configuration specified by its manifest.
 
+
+## Testing without the standard library
+
+You can add `//! no-std` on the first line of a single-file test to disable the loading of the standard library. We encourage using `no-std` when possible, as it makes the test significantly faster. 
+
+## Stopping after a compilation stage
+Add the `//! stage:stage_name` manifest attribute to stop after a specific compilation stage. Valid stages are:
```
kyouko-taiga:
```suggestion

Add `//! stage:stage_name` to stop the driver after a specific compilation stage. Valid stages are:
```

Tests/CompilerTests/README.md
```diff
@@ -7,12 +7,40 @@ A use case is either a single Hylo source file or a directory representing a pac
...
+- `stage:parsing`
+- `stage:typing`
+- `stage:lowering`
+- `stage:llvm`
+- `stage:execution` (default); applicable only in positive tests.
+
+## Expectation types
+
+### Exit status
+Specify an expected exit code by `exit-status:42` manifest attribute. `0` is expected by default.
```
kyouko-taiga, line 28:
```suggestion

Specify an expected exit code by `exit-status:42` manifest attribute. `0` is expected by default.
```

Tests/CompilerTests/README.md
```diff
@@ -7,12 +7,40 @@ A use case is either a single Hylo source file or a directory representing a pac
...
+Asserted when `stage` is set to `execution` (default). 
+
+### Textual Artifact Expectations
+You can add a `test-case.<artifact-tag>.expected` file besides your `test-case.hylo` to assert the contents of a compilation artifact. Valid artifacts are:
```
kyouko-taiga:
```suggestion

Add a `test-case.<artifact-tag>.expected` file besides your `test-case.hylo` to assert the contents of a compilation artifact. Valid artifacts are:
```

Tests/CompilerTests/README.md
```diff
@@ -7,12 +7,40 @@ A use case is either a single Hylo source file or a directory representing a pac
...
+- `*.raw-ir.expected`
+- `*.transformed-ir.expected`
+- `*.llvm-ir.expected`
+
+An artifact is composed of a set of contiguous sections, delimited by empty lines. E.g. each function in Hylo IR and LLVM IR are their own sections.
+
+It is sufficient to specify a subset of sections of the actual artifact. The expected section is matched with the observed section having the closest first line. Then they are compared for equality.
+
+## Inspecting Intermediate Artifacts
+The intermediate compilation artifacts are saved on test failure as `*.observed` files besides your test case. To save these even when a test passes, you can temporarily set `artifactsAreSavedOnSuccess` to `true` in `CompilerTests.swift`.
```
kyouko-taiga:
```suggestion

Intermediate compilation artifacts are saved on test failure as `*.observed` files besides the corresponding test case. Aou can temporarily set `artifactsAreSavedOnSuccess` to `true` in `CompilerTests.swift` to save these even when a test passes.
```

I think this option should be controlled by an environment variable.

kyouko-taiga (changes requested):
Nice work overall.

Tests/CompilerTests/CompilerTests.swift
tothambrus11: While the location is obvious for humans, it's not obvious for an IDE. The test case name has `test_positive` prefix and the file names are transformed from dashes to underscores, so an IDE lookup requires careful selection of a substring in the test case name + manual replacement of `-`->`_`. With this change, it becomes a single Cmd + click, so I think it's worth the extra printing effort.

Tests/CompilerTests/IRMatching.swift
tothambrus11: My last update simplified the logic, now we don't extract keys, just always match based on the most similar first line. Updated to:
```
// Pair with the observed section having the most similar first line.
```

Tests/CompilerTests/Manifest.swift
tothambrus11: Thanks! I also prefer the latter option when possible.

Tests/CompilerTests/README.md
tothambrus11: I considered that, but unfortunately environment variables can't be changed easily in IDEs without restarting them. I change this variable whenever I'm debugging something, so there needs to be a way to flip this flag easily. I don't think anyone would want to set this flag to true forever in their environment, as that can significantly slow down usual testing times due to saving files.

Tests/CompilerTests/CompilerTests.swift
tothambrus11: When each test case is reading from files, this seems like a negligible performance win, while requiring the key to become Comparable. The change is already done, so I'm fine with keeping SortedDictionary.

Tests/CompilerTests/CompilerTests.swift
tothambrus11: I added `executable` case and a list of textual artifacts.

Sources/Utilities/StringProtocol+Extensions.swift
tothambrus11: Mac before OSX. I was thinking to decide whether this is premature optimization or whether preparing for the case is premature worrying. I'll remove the second line, we can fix it if it ever becomes an issue.

Tests/CompilerTests/CompilerTests.swift
```diff
@@ -193,20 +234,30 @@ final class CompilerTests: XCTestCase {
...
+  /// Executes `action`, and logs the test case's `root` on test failure for easier diagnostics.
+  ///
+  /// Also logs and catches any error thrown by `action`, failing the test case.
+  private func loggingSourceOnFailure<T>(root: URL, _ action: () async throws -> T) async {
+    do {
+      _ = try await action()
+      if testFailed {
+        XCTFail("\nSource: \(root.path)\n")
+      }
+    } catch let e {
```
kyouko-taiga, line 270: In fact, I realized this is a poor strategy because it is possible an exception will be thrown because artifacts could not be saved. I adopted another setup on my branch that I think addresses this issue.

tothambrus11: Good catch, I applied the fix after rebasing.

Sources/Utilities/Sequence+Extensions.swift
```diff
@@ -65,4 +65,9 @@ extension Sequence {
     joinedString(separator: separator) { String(describing: $0) }
   }
 
-}
+  /// Returns the element of `self` that minimizes `f`, or `nil` if `self` is empty.
```
kyouko-taiga: I think the docs is wrong. My suggestion:

```suggestion
  /// Returns the element in `self` with the smallest value at `keyPath`, if any.
```

Also, I think it's worth saying a few words about what happens if there is more than one element with the same value at `keyPath`.

Tests/CompilerTests/IRMatching.swift
```diff
@@ -0,0 +1,63 @@
...
+
+  /// Returns the sections of `s`, in order of appearance.
+  ///
+  /// A section is a maximal run of consecutive non-blank lines, where a line is blank if it's empty
+  /// or contains only whitespace.
+  static func sections(of s: String) -> [Section] {
+    s.split(separator: "\n", omittingEmptySubsequences: false)
+      .split(whereSeparator: { $0.allSatisfy(\.isWhitespace) })
+      .map(Array.init)
+  }
```
kyouko-taiga, line 87: This whole thing seems expensive.

`sections(of:)` does two linear passes over essentially the entire input. I think we can implement a simple forward iterator that does the same thing with a linear pass and without allocating an array of arrays.

The loop in `comparions(expected:observed)` is O(n^2) if I'm not mistaken. On top of that, it is building another array in the loop at line 44, inside the very hot inner loop.

Here's how I would implement it:

```swift
extension Sequence {

  public func min<R: Comparable>(measuredBy p: (Element) -> R) -> Element? {
      self.min(by: { (a, b) in p(a) < p(b) })
    }

}

func comparisons(expected: String, observed: String) -> [Comparison] {
  var observed = Array(sections(of: observed))
  var end = observed.count - 1

  return sections(of: expected).map { (e) in
    let head = e.firstLine
    if let m = (0 ..< end).min(measuredBy: { (i) in head.difference(from: observed[i]).count }) {
      observed.swapAt(m, end)
      end -= 1
      return Comparison(expected: e, observed: observed[m])
    } else {
      return Comparison(expected: e, observed: nil)
    }
  }
}

func sections(of s: String) -> AnyIterator<Substring> {
  var i = s.startIndex

  return AnyIterator { () -> Substring? in
    // Skip leading whitespaces.
    while (i != s.endIndex) && s[i].isWhitespace { i = s.index(after: i) }
    // Did we reach the end of the string?
    if i == s.endIndex { return nil }

    // Find the next blank line.
    var rest = s[i...]
    while let e = rest.firstIndex(where: \.isNewline) {
      rest = s[s.index(after: e)...]
      let head = rest.firstLine

      if head.allSatisfy(\.isWhitespace) {
        defer { i = s.index(after: head.endIndex) }
        return s[i ..< e]
      } else {
        rest = s[s.index(after: e)...]
      }
    }

    // No more blank lines.
    defer { i = s.endIndex }
    return s[i...]
  }

}
```

I am not convinced that it is worth doing anything more complicated than an exact match on the first line of two sections. Do you have a use case in mind?

Sources/Utilities/Sequence+Extensions.swift
```diff
@@ -65,4 +65,9 @@ extension Sequence {
...
+  public func min<R: Comparable>(by keyPath: (Element) -> R) -> Element? {
+    self.min(by: { (a, b) in keyPath(a) < keyPath(b) })
+  }
```
kyouko-taiga: It occurred to me while working on a suggestion that this overload is ambiguous with `Sequence.min(by:)`, which is not great. The following solves the issue, I think, including in cases where trailing closure syntax is used:

```suggestion
  public func min<R: Comparable>(measuredBy p: (Element) -> R) -> Element? {
    self.min(by: { (a, b) in p(a) < p(b) })
  }
```

Also, `p` is not a key path, which is a specific feature in Swift. We should not use this term for an arbitrary function.

kyouko-taiga (changes requested):
I am concerned about the complexity (both algorithmic and cognitive) of the matching logic. Can we optimize and/or simplify.

Tests/CompilerTests/IRMatching.swift
tothambrus11: If the generated function signature changes slightly, we would likely match and display a correct assertion. Otherwise, the user is left to read the whole `.observed` file and do the same manually, so I think allowing a bit of lenience is very useful for UX.

Thanks for the iterator based approach. While we are most likely IO-bound with the 4+ file reads during a test case, I can imagine that we will generate a lot of IR at some point, so this can help. We can also try matching by first line in the happy path, and only do the sorting based approach when there was a failure.

Sources/Utilities/Sequence+Extensions.swift
tothambrus11: Swift doesn't specify it, so I just added that it's unspecified.

Tests/CompilerTests/IRMatching.swift
kyouko-taiga: Do you have a real use case to present ? Because as it stands I don't buy that there will be cases where just the first line changes a little and I'm calling YAGNI.

Besides, if a failure occurs, we have to inspect the observed file anyway and at that point we can use any third party tool to get a nice diff. 

I don't want to do complex work in the failure path either. I want to reduce complexity.

tothambrus11: As any other part of the IR, the first line (our section matching key) may change in the future for various reasons:
- change of IR syntax
- change of lowered function calling convention
- change of name mangling (when matching LLVM IR)

The reason for why it's important to be somewhat resilient to these changes is to be able to still produce the right set of assertions in case of a test failure. In the future, this will be also be necessary for automatic blessing.

While there is a chance that we might never need this in an ideal world, I think leaving out the handling of this case is a complexity optimization not worth taking, as this feature allows us to make changes like the above more freely, knowing that we won't have to inspect the whole IR of each test case, just see the correctly matched breakages per section.

kyouko-taiga: I still don't buy it, really, but I also don't want to stand in the way if you feel strongly about it. Let's simply let the record show that I'm convinced this feature is YAGNI as it stands.

Tests/CompilerTests/IRMatching.swift
```diff
@@ -0,0 +1,91 @@
...
+  /// Each *section* of `expected` is matched independently against a section of `observed`, so the
+  /// result has one comparison per expected section. Sections of `observed` with no counterpart in
+  /// `expected` are ignored, making `expected` a partial specification of `observed`.
+  ///
+  /// A section is a maximal run of consecutive non-blank lines. An expected section is matched
+  /// against the observed section having the most similar first line.
+  static func comparisons(expected: String, observed: String) -> [Comparison] {
+    let observed = Array(sections(of: observed.normalizedLineEndings()))
+
+    // Index the observed sections by their first line so the common case (an exact match) is just a lookup.
```
kyouko-taiga:
```suggestion
    // Index the observed sections by their first line so the common case (an exact match) is just
    // a lookup.
```

Tests/CompilerTests/IRMatching.swift
```diff
@@ -0,0 +1,91 @@
...
+    // been consumed. As in `split(omittingEmptySubsequences: false)`, a trailing newline yields a
+    // final empty line.
+    func nextLine() -> Substring? {
+      if done { return nil }
+      let start = i
+      while (i != s.endIndex) && !s[i].isNewline { i = s.index(after: i) }
+      let line = s[start ..< i]
+      if i == s.endIndex { done = true } else { i = s.index(after: i) }
+      return line
+    }
```
kyouko-taiga: I believe it is better to declare this function inside the closure of your `AnyIterator` constructor for performance, so as to create single captures for variables `i` and `done`.

Tests/CompilerTests/IRMatchingTests.swift
```diff
@@ -0,0 +1,209 @@
...
+    XCTAssertNotEqual(cs[0].expected, cs[0].observed)
+  }
+
+  func testHyloPartialFallsBackToClosestFirstLine() {
+    let expected = """
+      fun main(set %p0: Int32) {
+        %r0 = return
+      }
+      """
+    // No exact first-line match: the body changes the signature slightly. The closest section wins.
```
kyouko-taiga:
```suggestion
    // No exact first-line match: the body changes the signature slightly. Closest section wins.
```

Tests/CompilerTests/IRMatching.swift
```diff
@@ -0,0 +1,91 @@
...
+        byFirstLine[head] ?? observed.min(measuredBy: { (s) in distance(head, s.firstLine) })
+
+      return Comparison(expected: e, observed: matched)
+    }
+  }
+
+  /// Returns an iterator over the sections of `s`, in order of appearance.
+  ///
+  /// A section is a maximal run of consecutive non-blank lines, where a line is blank if it's empty
+  /// or contains only whitespace. Sections are yielded as slices of `s` in a single forward pass.
```
kyouko-taiga:
```suggestion
  /// A section is a maximal run of consecutive non-blank lines, where a line is blank if it is
  /// empty or contains only whitespace. Sections are yielded as slices of `s` computed with a
  // single forward pass over `s`.
```

Tests/CompilerTests/IRMatching.swift
```diff
@@ -0,0 +1,91 @@
...
+  static func sections(of s: String) -> AnyIterator<Substring> {
+    var i = s.startIndex
+    var done = false
+
+    // Returns the next line of `s`—excluding its terminating newline—or `nil` once every line has
+    // been consumed. As in `split(omittingEmptySubsequences: false)`, a trailing newline yields a
+    // final empty line.
```
kyouko-taiga:
```suggestion
    /// Returns the next line of `s`—excluding its terminating newline—or `nil` once every line has
    /// been consumed. As in `split(omittingEmptySubsequences: false)`, a trailing newline yields a
    /// final empty line.
```

kyouko-taiga (approved):
LGTM modulo minor inline comments. Some of your lines are too long.

We can't write on the 100th column.

Issue #149, opened 2026-05-08
Parser crash on incomplete subscript call expression


tothambrus11:
_Disclaimer: while this may be a small fix, it requires getting familiar with our parsing library and understand how subscripts (square brackets) are parsed. The code to be modified lives in https://github.com/hylo-lang/hylo-new/tree/main/Sources/FrontEnd/Parser. You will likely need to edit Parser.swift, but Lexer.swift might also be nice to look at._

The following code results in crash in the parser. We should instead recover from this with a diagnostic.
```swift
subscript a() -> Int { 
  yield 1 + 2
}

fun main() -> Int {
  return a[
}
```

```

  0                  0x00007e8726ce2d76 _assertionFailure(_:_:file:line:flags:) + 438 in libswiftCore.so
  1 [ra]             0x00007e8726c06a61 _StringGuts.validateCharacterIndex(_:) + 160 in libswiftCore.so
  2 [ra]             0x00007e8726cea532 String.index(after:) + 17 in libswiftCore.so
  3 [ra]             0x0000588c92989c34 SourceFile.index(after:) + 67 in HyloPackageTests.xctest at /home/ambrus/hylo-new/Sources/FrontEnd/Files/SourceFile.swift:181:53
  4 [ra]             0x0000588c92923f4e Parser.newlinesBeforeNextToken() + 861 in HyloPackageTests.xctest at /home/ambrus/hylo-new/Sources/FrontEnd/Parser/Parser.swift:2478:28
  5 [ra]             0x0000588c929239f7 Parser.statementDelimiterBeforeNextToken() + 38 in HyloPackageTests.xctest at /home/ambrus/hylo-new/Sources/FrontEnd/Parser/Parser.swift:2484:6
  6 [ra]             0x0000588c92934da3 Parser.ensureStatementDelimiter() + 50 in HyloPackageTests.xctest at /home/ambrus/hylo-new/Sources/FrontEnd/Parser/Parser.swift:2705:9
  7 [ra]             0x0000588c9292202b $defer #1 () in Parser.parseStatement(in:) + 26 in HyloPackageTests.xctest at /home/ambrus/hylo-new/Sources/FrontEnd/Parser/Parser.swift:2188:13
  8 [ra] [system]    0x0000588c92933a9c Parser.parseStatement(in:) + 571 in HyloPackageTests.xctest at /home/ambrus/hylo-new/Sources/FrontEnd/Parser/Parser.swift
[...]

PR #214, opened 2026-06-26
Fix parser crash on unterminated calls

tothambrus11:
`position` refers to the next unconsumed token's position. If that's endIndex, it is illegal to get the subsequent index. This caused #149. The crash happened both on incomplete subscript and function calls.


tothambrus11:
@kyouko-taiga Could you take a quick look?

kyouko-taiga (approved):
I wonder why I tested `index(after:)` but sadly I forgot and I didn't leave any comment 😔 .

If all tests pass, LGTM.

Issue #24, opened 2026-01-26
Compiler crash at incomplete `type` declaration


tothambrus11:
The compiler crashes when there is an incomplete typealias declaration:

```hylo
type        // crashes
```
```hylo
type X      // crashes
```
```hylo
type X =       // works, produces diagnostic correctly
```
```hylo
type X = {Int, Int}    // works, no diagnostic
```

Reproduced using the [current main branch](https://github.com/hylo-lang/hylo-new/commit/5fa99adf68e6e4c5e583a78c59382fd504f9987e) and the `ir-lowering` branch.

The compiler crash prduces the following stack trace:
```
FrontEnd/Typer.swift:1046: Fatal error: Unexpectedly found nil while unwrapping an Optional value

💣 Program crashed: Illegal instruction at 0x000076d439eb79d6

Platform: x86_64 Linux (Ubuntu 24.04.3 LTS)

Thread 1 crashed:

  0 0x000076d439eb79d6 _assertionFailure(_:_:file:line:flags:) + 438 in libswiftCore.so
  1 Typer.declaredType(of:) + 906 in hc at /workspaces/hylo-language-server/hylo-new/Sources/FrontEnd/Typer/Typer.swift:1046:69

  1044│     assert(d.module == module, "dependency is not typed")
  1045│ 
  1046│     let c = program.parent(containing: d, as: TraitDeclaration.self)!                                                                                               
      │                                                                    ▲
  1047│     let t = typeOfTraitSelf(in: c)
  1048│     let w = WitnessExpression(value: .abstract, type: t)

  2 Typer.check(_:) + 67 in hc at /workspaces/hylo-language-server/hylo-new/Sources/FrontEnd/Typer/Typer.swift:333:9

   331│   /// Type checks `d`.
   332│   private mutating func check(_ d: AssociatedTypeDeclaration.ID) {
   333│     _ = declaredType(of: d)                                                                                                                                         
      │         ▲
   334│     checkUniqueDeclaration(d, of: program[d].identifier.value)
   335│   }

  3 Typer.check(_:) + 396 in hc at /workspaces/hylo-language-server/hylo-new/Sources/FrontEnd/Typer/Typer.swift:299:7

   297│     switch program.tag(of: d) {
   298│     case AssociatedTypeDeclaration.self:
   299│       check(castUnchecked(d, to: AssociatedTypeDeclaration.self))                                                                                                   
      │       ▲
   300│     case BindingDeclaration.self:
[...]

PR #215, opened 2026-06-26
Fix compiler crash on incomplete type declaration

tothambrus11:
See https://github.com/hylo-lang/hylo-new/issues/24

A type declaration is an associated type only inside a trait body; everywhere else it's expected to be a type alias, and thus a `=` or `<` is required after the identifier. After this fix, all the following cases get reported without a crash:

```
// Incomplete type alias declaration
type Y

// Incomplete type declaration
type

trait P {
  // Incomplete associated type declaration
  type
}
```

### DIFF
--- a/Sources/FrontEnd/Parser/Parser.swift
+++ b/Sources/FrontEnd/Parser/Parser.swift
@@ -28,6 +28,11 @@ public struct Parser {
       if case .typeBody = self { return true } else { return false }
     }
 
+    /// Returns `true` iff `self` denotes the body of a trait declaration.
+    var isTraitBody: Bool {
+      self == .typeBody(.init(TraitDeclaration.self))
+    }
+
   }
 
   /// The tokens in the input.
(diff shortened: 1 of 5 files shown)


Sources/FrontEnd/Parser/Parser.swift
```diff
@@ -28,6 +28,15 @@ public struct Parser {
       if case .typeBody = self { return true } else { return false }
     }
 
+    /// Returns `true` iff `self` denotes the body of a trait declaration.
+    var isTraitBody: Bool {
+      if case .typeBody(let t) = self { 
+        t == TraitDeclaration.self
+      } else { 
+        false
+      }
```
kyouko-taiga:
```suggestion
      self == .typeBody(TraitDeclaration.self)
```
Would that work?

Sources/FrontEnd/Parser/Parser.swift
```diff
@@ -1185,8 +1194,23 @@ public struct Parser {
     let introducer = try take(.type) ?? expected("'type'")
     let identifier = parseSimpleIdentifier()
 
-    // If the next token is `<` or `=`, commit to a type alias declaration.
-    if next(is: .leftAngle) || next(is: .assign) {
+    // In a trait body, a 'type' declaration that is not followed by `<` or `=` introduces an
+    // associated type. Everywhere else, 'type' always introduces a type alias.
+    if context.isTraitBody && !(next(is: .leftAngle) || next(is: .assign)) {
```
kyouko-taiga:
```suggestion
    if next(is: .assign)) || (context.isTraitBody && !(next(is: .leftAngle)) {
```
I think (without evidence) that type aliases are more common than associated type declarations.

kyouko-taiga (approved):
LGTM. I left minor suggestions.

Sources/FrontEnd/Parser/Parser.swift
tothambrus11: There are at least as many trait implementations as traits (unless we have dead code or do fancy stuff), so I think your intuition should be correct.

Sources/FrontEnd/Parser/Parser.swift
tothambrus11: Oh nice, yes!

`self == .typeBody(.init(TraitDeclaration.self))`

PR #217, opened 2026-06-27
Update SwiftyLLVM for simplified API

tothambrus11:


kyouko-taiga (approved):
LGTM. What are your thoughts @tothambrus11 ?

tothambrus11:
I found the phrases like `v.v` and `x.v` a little meaningless, but apart from that I think it's an improvement. This is something one needs to get used to when working on LLVM lowering.

PR #218, opened 2026-06-27
Fix typer crash on bodyless local functions

tothambrus11:
`Typer.implicitCaptures(of:)` force-unwraps `program[d].body`, assuming every local function has a body.

A local function written without a body is a already diagnosed, but without an early exit. This PR updates `impicitCaptures` to be lenient about a missing `body`, returning a sensible default (0 captures) when body is not present.

### Reproducer
```hylo
//! no-std stage:typing

fun outer() -> Void {
  fun inner()
}
```

### DIFF
--- a/Sources/FrontEnd/Typer/Typer.swift
+++ b/Sources/FrontEnd/Typer/Typer.swift
@@ -274,8 +274,11 @@ public struct Typer {
       }
     }
 
+    /// Nothing to do if the function has no body; missing definitions are diagnosed elsewhere.
+    guard let body = program[d].body else { return captures }
+
     // Visit the syntax tree to collect all captures.
-    var work = program[d].body!.reversed().map({ (s) in (s.erased, false) })
+    var work = body.reversed().map({ (s) in (s.erased, false) })
     while let (n, isMarkedForMutation) = work.popLast() {
       switch program.tag(of: n) {
       case InoutExpression.self:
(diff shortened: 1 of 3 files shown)


Tests/CompilerTests/negative/body-less-local-function.hylo
```diff
@@ -0,0 +1,7 @@
+//! no-std stage:typing
+
+// A local function without a body must be diagnosed rather than crashing the typer while it
+// collects implicit captures (see `Typer.implicitCaptures(of:)`).
```
tothambrus11:
```suggestion
```

Tests/CompilerTests/negative/body-less-local-function.diagnostics.expected
```diff
@@ -0,0 +1,3 @@
+negative/body-less-local-function.hylo:6.7-12: error: declaration requires a body
```
tothambrus11:
```suggestion
negative/body-less-local-function.hylo:4.7-12: error: declaration requires a body
```

Sources/FrontEnd/Typer/Typer.swift
```diff
@@ -274,8 +274,11 @@ public struct Typer {
       }
     }
 
+    /// Implicit captures require body. Missing body is diagnosed without early exit.
```
```suggestion
    /// Nothing to do if the function has no body; missing definitions are diagnosed elsewhere.
```

Tests/CompilerTests/negative/bodyless-local-function.hylo
```diff
@@ -0,0 +1,5 @@
+//! no-std stage:typing
```
kyouko-taiga, line 1: File should be named `bodyless-local-function` since "bodyless" is a word.

kyouko-taiga (changes requested):
Very nice! I left inline comments.

PR #219, opened 2026-06-27
List members for LSP support

tothambrus11:
I added support for listing type members visible from a given scope. This works similarly as name resolution except we don't filter the results but collect all of them.

There was a funky member showing in the member list when local `given`s were in scope for a type, `$e` and `$f`. These are synthesized during parsing. I extended the `VariableDeclaration` to save whether the variable has a synthesized name or not. Then I filter declarations during member enumeration based on this.

I believe this enumeration can be quite expensive as we add more declarations, so in the future we should find a more gradual approach where we first list native members, then stream results e.g. level by level asynchronously. Maybe we can also make some tree-shaped cache where assembling a final name list may be cheaper than doing it from scratch.

### DIFF
--- a/Sources/FrontEnd/Program.swift
+++ b/Sources/FrontEnd/Program.swift
@@ -1946,4 +1946,49 @@ extension Program {
     })
   }
 
+  /// Returns every member reachable on a value of type `t` from `scopeOfUse`.
+  ///
+  /// The result is the union of `t`'s native members, the members of the extensions of `t` visible
+  /// from `scopeOfUse`, and the requirements of every visible trait to which `t` conforms in the
+  /// implicit context at `scopeOfUse`.
+  ///
+  /// Pass `selectionIsStatic` to choose which members are returned: static members and initializers
+  /// reached through a type or instance members reached through a value.
+  ///
+  /// - Requires: `m` has been type checked.
+  public mutating func members(
+    of t: AnyTypeIdentity, in m: Module.ID, visibleFrom scopeOfUse: ScopeIdentity,
+    static selectionIsStatic: Bool
+  ) -> [MemberCandidate] {
+    withTyper(typing: m) { (typer) in
+      typer.members(of: t, visibleFrom: scopeOfUse, static: selectionIsStatic)
+        .compactMap { (c) in
+          c.reference.target.map { (d) in
+            MemberCandidate(declaration: d, reference: c.reference, type: c.type)
+          }
+        }
+    }
+  }
+
+}
+
+/// A member reachable on a receiver, together with its type and how it is referred to.
+public struct MemberCandidate: Sendable {
+
+  /// The declaration of the member.
+  public let declaration: DeclarationIdentity
+
+  /// How the member is referred to (direct, bound member, or inherited via extension/conformance).
+  public let reference: DeclarationReference
+
+  /// The type of the member as selected on the receiver, with conformance substitutions applied.
+  public let type: AnyTypeIdentity
+
+  /// Creates an instance with the given properties.
+  public init(declaration: DeclarationIdentity, reference: DeclarationReference, type: AnyTypeIdentity) {
+    self.declaration = declaration
+    self.reference = reference
+    self.type = type
+  }
+
 }
--- a/Sources/FrontEnd/Program.swift
+++ b/Sources/FrontEnd/Program.swift
@@ -1968,36 +1968,28 @@ extension Program {
   /// - Requires: `m` has been type checked.
   public mutating func members(
     of t: AnyTypeIdentity, in m: Module.ID, visibleFrom scopeOfUse: ScopeIdentity,
-    static selectionIsStatic: Bool
+    resolvedStatically selectionIsStatic: Bool
   ) -> [MemberCandidate] {
     withTyper(typing: m) { (typer) in
-      typer.members(of: t, visibleFrom: scopeOfUse, static: selectionIsStatic)
-        .compactMap { (c) in
-          c.reference.target.map { (d) in
-            MemberCandidate(declaration: d, reference: c.reference, type: c.type)
-          }
-        }
+      typer.members(of: t, visibleFrom: scopeOfUse, resolvedStatically: selectionIsStatic)
+        .map { (c) in MemberCandidate(declaration: c.reference, type: c.type) }
     }
   }
 
 }
 
-/// A member reachable on a receiver, together with its type and how it is referred to.
+/// A member declaration and its type.
 public struct MemberCandidate: Sendable {
 
-  /// The declaration of the member.
-  public let declaration: DeclarationIdentity
-
   /// How the member is referred to (direct, bound member, or inherited via extension/conformance).
-  public let reference: DeclarationReference
+  public let declaration: DeclarationReference
 
   /// The type of the member as selected on the receiver, with conformance substitutions applied.
   public let type: AnyTypeIdentity
 
   /// Creates an instance with the given properties.
-  public init(declaration: DeclarationIdentity, reference: DeclarationReference, type: AnyTypeIdentity) {
+  public init(declaration: DeclarationReference, type: AnyTypeIdentity) {
     self.declaration = declaration
-    self.reference = reference
     self.type = type
   }
 
--- a/Sources/FrontEnd/Program.swift
+++ b/Sources/FrontEnd/Program.swift
@@ -1946,6 +1946,16 @@ extension Program {
     })
   }
 
+  /// Returns the type denoted by "Self" in `scopeOfUse`, or `nil` if "Self"
+  /// is not legal there.
+  ///
+  /// - Requires: The module containing `scopeOfUse` has been type checked.
+  public mutating func typeOfSelf(in scopeOfUse: ScopeIdentity) -> AnyTypeIdentity? {
+    withTyper(typing: scopeOfUse.file.module) { (typer) in
+      typer.typeOfSelf(in: scopeOfUse)
+    }
+  }
+
   /// Returns every member reachable on a value of type `t` from `scopeOfUse`.
   ///
   /// The result is the union of `t`'s native members, the members of the extensions of `t` visible
--- a/Sources/FrontEnd/Syntax/Declarations/VariableDeclaration.swift
+++ b/Sources/FrontEnd/Syntax/Declarations/VariableDeclaration.swift
@@ -14,6 +14,9 @@ public struct VariableDeclaration: Declaration, Pattern {
   /// The site from which `self` was parsed.
   public var site: SourceSpan { identifier.site }
 
+  /// `true` iff `identifier` was synthesized rather than user-written.
+  public var synthesized: Bool
+
 }
 
 extension VariableDeclaration: Showable {
@@ -29,10 +32,12 @@ extension VariableDeclaration: Archivable {
 
   public init<T>(from archive: inout ReadableArchive<T>, in context: inout Any) throws {
     self.identifier = try archive.read(Parsed<String>.self, in: &context)
+    self.synthesized = try archive.read(Bool.self, in: &context)
   }
 
   public func write<T>(to archive: inout WriteableArchive<T>, in context: inout Any) throws {
     try archive.write(identifier, in: &context)
+    try archive.write(synthesized, in: &context)
   }
 
 }
--- a/Sources/FrontEnd/Syntax/Declarations/VariableDeclaration.swift
+++ b/Sources/FrontEnd/Syntax/Declarations/VariableDeclaration.swift
@@ -15,7 +15,7 @@ public struct VariableDeclaration: Declaration, Pattern {
   public var site: SourceSpan { identifier.site }
 
   /// `true` iff `identifier` was synthesized rather than user-written.
-  public var synthesized: Bool
+  public var isSynthesized: Bool
 
 }
 
@@ -32,12 +32,12 @@ extension VariableDeclaration: Archivable {
 
   public init<T>(from archive: inout ReadableArchive<T>, in context: inout Any) throws {
     self.identifier = try archive.read(Parsed<String>.self, in: &context)
-    self.synthesized = try archive.read(Bool.self, in: &context)
+    self.isSynthesized = try archive.read(Bool.self, in: &context)
   }
 
   public func write<T>(to archive: inout WriteableArchive<T>, in context: inout Any) throws {
     try archive.write(identifier, in: &context)
-    try archive.write(synthesized, in: &context)
+    try archive.write(isSynthesized, in: &context)
   }
 
 }
--- a/Sources/FrontEnd/Typer/Typer.swift
+++ b/Sources/FrontEnd/Typer/Typer.swift
@@ -4601,9 +4601,9 @@ public struct Typer {
 
   /// Returns `true` iff `m` is part of the declaration shown when enumerating the members of a type.
   ///
-  /// In addition to `resolvableWithQualification`, this excludes the variable introduced by an
-  /// implicit (`using`/`given`) binding. Such a value is should be reached through implicit
-  /// resolution, not member selection.
+  /// In addition to `resolvableWithQualification`, this excludes the synthesized variable 
+  /// introduced by a given. Such a value should be reached through implicit resolution, not
+  /// member selection.
   private func isEnumerableMember(_ m: DeclarationIdentity) -> Bool {
     guard resolvableWithQualification(m) else { return false }
     if let v = program.cast(m, to: VariableDeclaration.self) {
--- a/Sources/FrontEnd/Typer/Typer.swift
+++ b/Sources/FrontEnd/Typer/Typer.swift
@@ -3099,7 +3099,7 @@ public struct Typer {
 
   /// Returns the type of an instance of `Self` in `s`, or `nil` if `s` isn't notionally in the
   /// scope of a type declaration.
-  private mutating func typeOfSelf(in s: ScopeIdentity) -> AnyTypeIdentity? {
+  internal mutating func typeOfSelf(in s: ScopeIdentity) -> AnyTypeIdentity? {
     if let memoized = cache.scopeToTypeOfSelf[s] { return memoized }
 
     guard let n = s.node else { return nil }
--- a/Sources/FrontEnd/Typer/Typer.swift
+++ b/Sources/FrontEnd/Typer/Typer.swift
@@ -4243,15 +4243,15 @@ public struct Typer {
   ) -> [NameResolutionCandidate] {
     if q.isVariable { return [] }
 
-    // `qualificationForSelection` normalizes the receiver type (a `let`/`inout` projection unwraps
-    // to its projectee, a metatype to its inhabitant). The static/instance choice is the caller's,
-    // not derived from `q`, so the two cannot silently disagree.
     let r = qualificationForSelection(on: q).type
 
     var candidates: [NameResolutionCandidate] = []
-    candidates.append(contentsOf: nativeMembers(of: r, resolvedStatically: s))
-    candidates.append(contentsOf: extensionMembers(of: r, visibleFrom: scopeOfUse, resolvedStatically: s))
-    candidates.append(contentsOf: inheritedMembers(of: r, visibleFrom: scopeOfUse, resolvedStatically: s))
+    candidates.append(contentsOf:
+      nativeMembers(of: r, resolvedStatically: s))
+    candidates.append(contentsOf:
+      extensionMembers(of: r, visibleFrom: scopeOfUse, resolvedStatically: s))
+    candidates.append(contentsOf:
+      inheritedMembers(of: r, visibleFrom: scopeOfUse, resolvedStatically: s))
 
     return Self.uniqueCandidates(candidates)
   }
@@ -4271,7 +4271,7 @@ public struct Typer {
 
   /// Returns the natives members of `q`.
   /// 
-  /// Static members are returned iff `selectionIsStatic` is true.
+  /// Static members are included iff `selectionIsStatic` is true.
   private mutating func nativeMembers(
     of q: AnyTypeIdentity, resolvedStatically selectionIsStatic: Bool
   ) -> [NameResolutionCandidate] {
@@ -4294,7 +4294,7 @@ public struct Typer {
 
   /// Returns the members of `q` that are inherited by extension and visible from `scopeOfUse`.
   /// 
-  /// Static members are returned iff `selectionIsStatic` is true.
+  /// Static members are included iff `selectionIsStatic` is true.
   private mutating func extensionMembers(
     of q: AnyTypeIdentity, visibleFrom scopeOfUse: ScopeIdentity, 
     resolvedStatically selectionIsStatic: Bool
@@ -4322,8 +4322,8 @@ public struct Typer {
 
   /// Returns a candidate for every trait requirement reachable on `q` through conformances visible
   /// from `scopeOfUse`.
-  /// 
-  /// Static members are returned iff `selectionIsStatic` is true.
+  ///
+  /// Static members are included iff `selectionIsStatic` is true.
   private mutating func inheritedMembers(
     of q: AnyTypeIdentity, visibleFrom scopeOfUse: ScopeIdentity, resolvedStatically selectionIsStatic: Bool
   ) -> [NameResolutionCandidate] {
--- a/Sources/FrontEnd/Typer/Typer.swift
+++ b/Sources/FrontEnd/Typer/Typer.swift
@@ -4230,17 +4230,16 @@ public struct Typer {
     return candidates
   }
 
-  /// Returns a candidate for every member of `q` reachable in `scopeOfUse`.
+  /// This method is similar to `resolve(_:memberOf:visibleFrom:)`, except that it collects all
+  /// members (declared together with the type or inherited by extension/conformance) that are
+  /// visible from `scopeOfUse`, rather than resolving a single name.
   ///
-  /// This is the enumerating counterpart of `resolve(_:memberOf:visibleFrom:)`: rather than
-  /// resolving a single name, it collects all native members, all members of the applicable
-  /// extensions visible from `scopeOfUse`, and all requirements of every visible trait to which `q`
-  /// conforms (decided by `summon`, exactly as name resolution would). It is the building block for
-  /// IDE features such as member completion.
+  /// This method is intended for tooling and should not be used to perform name resolution. Its
+  /// implementation does not uphold coherence.
   ///
   /// - Requires: `q` is not a unification variable and the enclosing module has been type checked.
   internal mutating func members(
-    of q: AnyTypeIdentity, visibleFrom scopeOfUse: ScopeIdentity, static s: Bool
+    of q: AnyTypeIdentity, visibleFrom scopeOfUse: ScopeIdentity, resolvedStatically s: Bool
   ) -> [NameResolutionCandidate] {
     if q.isVariable { return [] }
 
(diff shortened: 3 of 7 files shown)


Sources/FrontEnd/Syntax/Declarations/VariableDeclaration.swift
```diff
@@ -14,6 +14,9 @@ public struct VariableDeclaration: Declaration, Pattern {
   /// The site from which `self` was parsed.
   public var site: SourceSpan { identifier.site }
 
+  /// `true` iff `identifier` was synthesized rather than user-written.
+  public var synthesized: Bool
```
kyouko-taiga:
```suggestion
  public var isSynthesized: Bool
```

I know this convention is controversial but it has been used uniformly so far.

Sources/FrontEnd/Typer/Typer.swift
```diff
@@ -4165,6 +4165,126 @@ public struct Typer {
     return candidates
   }
 
+  /// Returns a candidate for every member of `q` reachable in `scopeOfUse`.
+  ///
+  /// This is the enumerating counterpart of `resolve(_:memberOf:visibleFrom:)`: rather than
+  /// resolving a single name, it collects all native members, all members of the applicable
+  /// extensions visible from `scopeOfUse`, and all requirements of every visible trait to which `q`
+  /// conforms (decided by `summon`, exactly as name resolution would). It is the building block for
+  /// IDE features such as member completion.
```
```suggestion
  /// This method is similar to `resolve(_:memberOf:visibleFrom:)`, except that it collects all
  /// members (declared together with the type or inherited by extension/conformance) that are
  // visible from `scopeOfUse`, rather than resolving a single name.
  ///
  /// This method is intended for tooling and should not be used to perform name resolution. Its
  /// implementation does not uphold coherence.
```

In general, you should avoid using "this" as the subject of a sentence.

Sources/FrontEnd/Typer/Typer.swift
```diff
@@ -4165,6 +4165,126 @@ public struct Typer {
...
+  ///
+  /// - Requires: `q` is not a unification variable and the enclosing module has been type checked.
+  internal mutating func members(
+    of q: AnyTypeIdentity, visibleFrom scopeOfUse: ScopeIdentity, static s: Bool
+  ) -> [NameResolutionCandidate] {
+    if q.isVariable { return [] }
+
+    // `qualificationForSelection` normalizes the receiver type (a `let`/`inout` projection unwraps
+    // to its projectee, a metatype to its inhabitant). The static/instance choice is the caller's,
+    // not derived from `q`, so the two cannot silently disagree.
```
kyouko-taiga: Why do you need such a long comment? The docs of the method is already explaining how it works. If you think that deserves clarification, then please modify the docs at definition site rather than here.

Sources/FrontEnd/Typer/Typer.swift
```diff
@@ -4165,6 +4165,126 @@ public struct Typer {
...
+    let r = qualificationForSelection(on: q).type
+
+    var candidates: [NameResolutionCandidate] = []
+    candidates.append(contentsOf: nativeMembers(of: r, statically: s))
+    candidates.append(contentsOf: extensionMembers(of: r, visibleFrom: scopeOfUse, statically: s))
+    candidates.append(contentsOf: inheritedMembers(of: r, visibleFrom: scopeOfUse, statically: s))
+
+    // Dedup by declaration identity (e.g. the same trait requirement reached through several
+    // witnesses) while keeping same-named members of distinct declarations (overloads, or members
+    // offered by more than one in-scope trait).
```
```suggestion
    // Remove duplicate resolution candidates while keeping same-named members of distinct
    // declarations (e.g., overloads or members offered by more than one in-scope trait).
```

I think this block deserves to be implemented in its own abstraction.

Sources/FrontEnd/Typer/Typer.swift
```diff
@@ -4165,6 +4165,126 @@ public struct Typer {
...
+      guard let d = c.reference.target else { return true }
+      return seen.insert(d).inserted
+    }
+  }
+
+  /// Returns a candidate for every native member of `q`.
+  /// 
+  /// Static members are returned iff `selectionIsStatic` is true.
+  private mutating func nativeMembers(
+    of q: AnyTypeIdentity, statically selectionIsStatic: Bool
```
```suggestion
    of q: AnyTypeIdentity, resolvedStatically selectionIsStatic: Bool
```

Sources/FrontEnd/Typer/Typer.swift
```diff
@@ -4165,6 +4165,126 @@ public struct Typer {
...
+    // witnesses) while keeping same-named members of distinct declarations (overloads, or members
+    // offered by more than one in-scope trait).
+    var seen: Set<DeclarationIdentity> = []
+    return candidates.filter { (c) in
+      guard let d = c.reference.target else { return true }
+      return seen.insert(d).inserted
+    }
+  }
+
+  /// Returns a candidate for every native member of `q`.
```
```suggestion
/// Returns the native members of `q`.
```
The term "candidate" makes sense in the context of name resolution, as results are candidates and not definitive answers. Here, you are returning a set of symbols that are definitely members, not candidate members.

Sources/FrontEnd/Typer/Typer.swift
```diff
@@ -4412,6 +4532,19 @@ public struct Typer {
     }
   }
 
+  /// Returns `true` iff `m` should be offered when enumerating the members of a type.
```
```suggestion
  /// Returns `true` iff `m` is part of the declaration shown when enumerating the members of a type.
```

Sources/FrontEnd/Typer/Typer.swift
```diff
@@ -4412,6 +4532,19 @@ public struct Typer {
...
+  ///
+  /// In addition to `resolvableWithQualification`, this excludes the variable introduced by an
+  /// implicit (`using`/`given`) binding. Such a value is should be reached through implicit
+  /// resolution, not member selection.
+  private func isEnumerableMember(_ m: DeclarationIdentity) -> Bool {
```
kyouko-taiga, line 4607: I am not a fan of this name. It means very little at call site without further qualification. All declarations are enumerable and there are many methods that enumerate members regardless of the result of this method.

Sources/FrontEnd/Typer/Typer.swift
```diff
@@ -4165,6 +4165,126 @@ public struct Typer {
...
+  /// Static members are returned iff `selectionIsStatic` is true.
+  private mutating func nativeMembers(
+    of q: AnyTypeIdentity, statically selectionIsStatic: Bool
+  ) -> [NameResolutionCandidate] {
+    //  Mirrors `resolve(_:nativeMemberOf:statically:)` but without filtering by name.
+    let (context, receiver) = program.types.contextAndHead(q)
+    var candidates: [NameResolutionCandidate] = []
+    for ds in declarations(nativeMembersOf: q).values {
+      for m in ds {
+        if !isEnumerableMember(m) { continue }
```
```suggestion
      for m in ds where isEnumerableMember(m) {
```

Sources/FrontEnd/Typer/Typer.swift
```diff
@@ -4165,6 +4165,126 @@ public struct Typer {
...
+        member = program.types.introduce(context, into: member)
+        candidates.append(
+          .init(reference: selectionIsStatic ? .direct(m) : .member(m), type: member))
+      }
+    }
+    return candidates
+  }
+
+  /// Returns a candidate for every member declared in an extension of `q` visible from 
+  /// `scopeOfUse`.
```
```suggestion
  /// Returns the members of `q` that are inherited by extension and visible from `scopeOfUse`.
```

Sources/FrontEnd/Typer/Typer.swift
```diff
@@ -4165,6 +4165,126 @@ public struct Typer {
...
+    of q: AnyTypeIdentity, visibleFrom scopeOfUse: ScopeIdentity, statically selectionIsStatic: Bool
+  ) -> [NameResolutionCandidate] {
+    // Mirrors `resolve(_:inheritedMemberOf:visibleFrom:statically:)` but enumerates every visible 
+    // trait and all of its requirements instead of resolving a single name.
+
+    var candidates: [NameResolutionCandidate] = []
+    for concept in traits(visibleFrom: scopeOfUse) {
+      let vs = program[concept].parameters.map({ _ in fresh().erased })
+      let model = typeOfModel(of: q, conformingTo: concept, with: vs)
+      for a in summon(model.erased, in: scopeOfUse) {
```
kyouko-taiga, line 4337: What's the point looping over all the possible candidates of summon?

Sources/FrontEnd/Typer/Typer.swift
```diff
@@ -4165,6 +4165,126 @@ public struct Typer {
...
+        for m in program[concept].members {
+          if !isEnumerableMember(m) { continue }
+          // Ignore the member if it is static but the qualification isn't.
+          if program.isStatic(m) && !selectionIsStatic { continue }
+          var member = typeOfImplementation(satisfying: m, in: w)
+          if !selectionIsStatic {
+            member = typeOfBoundMember(referringTo: m, withUnboundType: member)
+          }
+          candidates.append(
+            .init(reference: .inherited(w, m, statically: selectionIsStatic), type: member))
```
kyouko-taiga, line 4349: This code is duplicated in several place. It should be wrapped into an abstraction.

Sources/FrontEnd/Typer/Typer.swift
```diff
@@ -4412,6 +4532,19 @@ public struct Typer {
     }
   }
 
+  /// Returns `true` iff `m` should be offered when enumerating the members of a type.
+  ///
+  /// In addition to `resolvableWithQualification`, this excludes the variable introduced by an
+  /// implicit (`using`/`given`) binding. Such a value is should be reached through implicit
```
kyouko-taiga: What's a `using` binding? Did some AI think we're Scala 🤔 ?

Sources/FrontEnd/Program.swift
```diff
@@ -1823,4 +1823,59 @@ extension Program {
...
+          c.reference.target.map { (d) in
+            MemberCandidate(declaration: d, reference: c.reference, type: c.type)
+          }
+        }
+    }
+  }
+
+}
+
+/// A member reachable on a receiver, together with its type and how it is referred to.
```
kyouko-taiga:
```suggestion
/// A member declaration and its type.
```

Sources/FrontEnd/Program.swift
```diff
@@ -1823,4 +1823,59 @@ extension Program {
...
+public struct MemberCandidate: Sendable {
+
+  /// The declaration of the member.
+  public let declaration: DeclarationIdentity
+
+  /// How the member is referred to (direct, bound member, or inherited via extension/conformance).
+  public let reference: DeclarationReference
```
kyouko-taiga: A declaration reference includes a declaration for the cases that you have listed. One can even write `reference.target!`.

Also I think `reference` is the wrong name. I would pick `declaration` or `use`.

Tests/FrontEndTests/MemberListingTests.swift
```diff
@@ -0,0 +1,185 @@
+import FrontEnd
+import XCTest
+
+/// Tests for `Program.members(of:in:visibleFrom:static:)`, the member-listing query used by LSP.
+final class MemberListingTests: XCTestCase {
+
+  func testListsNativeConformanceAndExtensionMembers() async throws {
```
kyouko-taiga:
```suggestion
  func testEnumerateMembers() async throws {
```
We do not care about qualifying what sort of members the function is enumerating, since it is enumerating all of them and there is no test enumerating subsets.

Tests/FrontEndTests/MemberListingTests.swift
```diff
@@ -0,0 +1,185 @@
...
+    // An instance receiver should see its own members, the members of an applicable extension, and
```
kyouko-taiga: "instance receiver" doesn't mean much to me. Also I think this comment is not very useful. It is just repeating the rules of the language, without any particular caveats that is worth observing in this specific instance.

Tests/FrontEndTests/MemberListingTests.swift
```diff
@@ -0,0 +1,185 @@
...
+      extension A { fun h() {} }
+      """)
+    XCTAssert(names.contains("g"))  // native
+    XCTAssert(names.contains("f"))  // conformance
+    XCTAssert(names.contains("h"))  // extension
+  }
+
+  func testListsStaticMembersFromAllSources() async throws {
+    // A type receiver should see static members from the type itself, an extension, and a
+    // conformance.
```
kyouko-taiga: I think I understand how you understand the term "receiver". But I also think that is the wrong interpretation, or at least the wrong way to phrase it.

A receiver is an instance of a type, always. The type of a receiver may be either an "ordinary" type (e.g., `Int`) or a meta type (e.g., `Metatype<Int>`), which is a singleton type whose only instance is a type.

Tests/FrontEndTests/MemberListingTests.swift
```diff
@@ -0,0 +1,185 @@
...
+      }
+      given A is P { fun f() {} }
+      extension A { fun h() {} }
+      """)
+    XCTAssert(names.contains("g"))  // native
+    XCTAssert(names.contains("f"))  // conformance
+    XCTAssert(names.contains("h"))  // extension
+  }
+
+  func testListsStaticMembersFromAllSources() async throws {
```
kyouko-taiga:
```suggestion
  func testEnumerateStaticMembers() async throws {
```
Again, there is no need to specify that we're picking members "from all sources" since that is the default.

Tests/FrontEndTests/MemberListingTests.swift
```diff
@@ -0,0 +1,185 @@
...
+      extension A { public static fun g() {} }
+      """)
+    XCTAssert(names.contains("f"))  // native
+    XCTAssert(names.contains("g"))  // extension
+    XCTAssert(names.contains("h"))  // conformance
+  }
+
+  func testStaticConformanceRequirementRequiresStaticSelection() async throws {
+    // A static trait requirement is reachable only through a static selection (`A.s`), not on an
+    // instance, so the `static` flag must be honored in the conformance branch.
```
kyouko-taiga: Another dubious use of "instance". `Int` is an instance of `Metatype<Int>`.

Also, what's a conformance branch?

Tests/FrontEndTests/MemberListingTests.swift
```diff
@@ -0,0 +1,185 @@
...
+      }
+      given A is P { public static fun h() {} }
+      extension A { public static fun g() {} }
+      """)
+    XCTAssert(names.contains("f"))  // native
+    XCTAssert(names.contains("g"))  // extension
+    XCTAssert(names.contains("h"))  // conformance
+  }
+
+  func testStaticConformanceRequirementRequiresStaticSelection() async throws {
```
kyouko-taiga:
```suggestion
  func testEnumerateStaticInheritedMembers() async throws {
```

Tests/FrontEndTests/MemberListingTests.swift
```diff
@@ -0,0 +1,185 @@
...
+      struct A { public memberwise init }
+      given A is P { public static fun s() {} }
+      """
+    let statics = try await staticMembers(ofStruct: "A", in: source)
+    let instance = try await instanceMembers(ofStruct: "A", in: source)
+    XCTAssertTrue(statics.contains("s"))
+    XCTAssertFalse(instance.contains("s"))
+  }
+
+  func testDoesNotListSynthesizedConformanceWitnesses() async throws {
```
kyouko-taiga:
```suggestion
  func testIgnoreSynthesizedDeclarations() async throws {
```

The following is not a conformance witness.
```hylo
given let x: Int = 0
```

Tests/FrontEndTests/MemberEnumerationTests.swift
```diff
@@ -0,0 +1,185 @@
...
+      extension T : P { fun h() {} }
+      """)
+    XCTAssert(names.contains("r"))
+    XCTAssert(names.contains("h"))
+    XCTAssert(names.contains("x"))
+    XCTAssertFalse(names.contains(where: { $0.hasPrefix("$") }))
+  }
+
+  func testSubstitutesTypeArgumentsIntoMemberTypes() async throws {
+    // Listing the members of a generic type applied to concrete arguments must substitute those
```
kyouko-taiga, line 74:
```suggestion
  // Type parameters in the types of members qualified by an instance of an applied constructor should
  // be replaced with their corresponding arguments.
```

Also, "listing" sounds off. I would use enumerating, gathering, or collecting. The same comment applies to the name of the file.

Tests/FrontEndTests/MemberListingTests.swift
```diff
@@ -0,0 +1,185 @@
...
+    let v = try XCTUnwrap(p.select(.name(.init(identifier: "v"))).first)
+    let receiver = try XCTUnwrap(p.type(maybeAssignedTo: v))
+    let member = try XCTUnwrap(
+      p.members(of: receiver, in: m, visibleFrom: .init(file: f), static: false)
+        .first(where: { p.name(of: $0.declaration)?.identifier == "f" }))
+
+    XCTAssertEqual(p.show(member.type), "[let A<B>](let B) let -> Void")
+  }
+
+  func testTypeOfSelfIsTheEnclosingTypeWithinAStructScope() async throws {
```
kyouko-taiga: I really dislike this kind of test names. The sentence reads like a mangled symbol if I don't squint my eyes. Camel case is probably not the right way to write English.

Please pick something in line with the convention established by other tests. The name of the function has to be "about" the test, not the docs of the test.

Tests/FrontEndTests/MemberListingTests.swift
```diff
@@ -0,0 +1,185 @@
...
+      }
+      """)
+
+    XCTAssertNil(p.typeOfSelf(in: .init(file: f)))
+  }
+
+  // MARK: - Helpers
+
+  /// Returns the identifiers of the members `Program.members` reports for the struct named
+  /// `typeName` declared in `source`, selecting instance members.
```
kyouko-taiga: This sentence looks grammatically unsound.

Tests/FrontEndTests/MemberListingTests.swift
```diff
@@ -0,0 +1,185 @@
...
+  /// declared in `source`, selecting static members.
+  private func staticMembers(
+    ofStruct name: String, in source: SourceFile,
+    file: StaticString = #filePath, line: UInt = #line
+  ) async throws -> Set<String> {
+    try await memberNames(ofStruct: name, in: source, static: true, file: file, line: line)
+  }
+
+  /// Type-checks `source` as a single-file module and returns the identifiers of the members
+  /// reported for the struct named `typeName`, selecting static members iff `isStatic`.
```
kyouko-taiga:
```suggestion
  /// Type-checks `source` as a single-file module and returns the identifiers of the members
  /// of the struct named `typeName`, selecting static members iff `isStatic` is `true`.
```

Tests/FrontEndTests/MemberListingTests.swift
```diff
@@ -0,0 +1,185 @@
...
+  /// `typeName` declared in `source`, selecting instance members.
+  private func instanceMembers(
+    ofStruct typeName: String, in source: SourceFile,
+    file: StaticString = #filePath, line: UInt = #line
+  ) async throws -> Set<String> {
+    try await memberNames(ofStruct: typeName, in: source, static: false, file: file, line: line)
+  }
+
+  /// Returns the identifiers of the members `Program.members` reports for the struct named `name`
+  /// declared in `source`, selecting static members.
```
kyouko-taiga: This sentence looks grammatically unsound.

Tests/FrontEndTests/MemberListingTests.swift
```diff
@@ -0,0 +1,185 @@
...
+      """)
+
+    let d = try XCTUnwrap(
+      p.select(.and(.tag(StructDeclaration.self), .name(.init(identifier: "A")))).first,
+      "no struct named 'A'")
+    let selfType = try XCTUnwrap(p.typeOfSelf(in: .init(uncheckedFrom: d.erased)))
+    XCTAssertEqual(p.show(selfType), "A")
+  }
+
+  func testTypeOfSelfIsNilAtFileScope() async throws {
```
kyouko-taiga:
```suggestion
  func testTypeOfSelfAtFileScope() async throws {
```
I am not convinced we should start unit testing every single method of `Program` that way. And even if we did, this test does not belong here. `typeOfSelf` is not specific to the task of enumerating members.

Tests/FrontEndTests/MemberListingTests.swift
```diff
@@ -0,0 +1,185 @@
...
+
+    let d = try XCTUnwrap(
+      p.select(.and(.tag(StructDeclaration.self), .name(.init(identifier: typeName)))).first,
+      "no struct named '\(typeName)'", file: file, line: line)
+    // The type assigned to a struct declaration is its metatype; `members` normalizes that to the
+    // instance type and selects static or instance members per the flag.
+    let receiver = p.type(assignedTo: d)
+    return Set(
+      p.members(of: receiver, in: m, visibleFrom: .init(file: f), static: isStatic)
+        .compactMap({ p.name(of: $0.declaration)?.identifier }))
```
kyouko-taiga: That looks somewhat inefficient. You can write something much better with `reduce(into:)`

Tests/FrontEndTests/MemberListingTests.swift
```diff
@@ -0,0 +1,185 @@
...
+  private func staticMembers(
+    ofStruct name: String, in source: SourceFile,
+    file: StaticString = #filePath, line: UInt = #line
+  ) async throws -> Set<String> {
+    try await memberNames(ofStruct: name, in: source, static: true, file: file, line: line)
+  }
+
+  /// Type-checks `source` as a single-file module and returns the identifiers of the members
+  /// reported for the struct named `typeName`, selecting static members iff `isStatic`.
+  private func memberNames(
```
kyouko-taiga:
```suggestion
  private func members(
```
For consistency with the way other helpers are named.

kyouko-taiga (changes requested):
Well done. It is always difficult to navigate typer.

There are several things that I think could be improved, especially regarding the docs. I left inline comments.

Sources/FrontEnd/Typer/Typer.swift
tothambrus11: IIRC we talked about declaring regular function parameters with `using` to add them to the implicit context (it was a long time ago, so I'm likely wrong). Is there anything like this now / in plan?

Tests/FrontEndTests/MemberListingTests.swift
tothambrus11: Yes, we could, but I usually find a raw reduce harder to read than map/compactMap. This optimization would have negligible difference in performance compared to type-checking the program, given that the structs in test cases have at most a couple members.

Sources/FrontEnd/Typer/Typer.swift
```diff
@@ -4165,6 +4165,126 @@ public struct Typer {
     return candidates
   }
 
+  /// Returns a candidate for every member of `q` reachable in `scopeOfUse`.
```
RishabhRD: I see we don't use `s` in API summary. I am guessing it is there to include `static` members based on `true` or `false`. We should have that.

Side thought: We can have a more general API that takes predicate and includes only the members that satisfies the predicate and provide static predicate as a function that can be used on call site. IDK if that would be worthy refactor.

Sources/FrontEnd/Typer/Typer.swift
```diff
@@ -4165,6 +4165,126 @@ public struct Typer {
...
+    var seen: Set<DeclarationIdentity> = []
+    return candidates.filter { (c) in
+      guard let d = c.reference.target else { return true }
+      return seen.insert(d).inserted
+    }
+  }
+
+  /// Returns a candidate for every native member of `q`.
+  /// 
+  /// Static members are returned iff `selectionIsStatic` is true.
```
RishabhRD: IDK if `selectionIsStatic` is true, then the result only contains static members or includes static members with other members.
```suggestion
  /// Static members are included iff `selectionIsStatic` is true.
```

Sources/FrontEnd/Program.swift
```diff
@@ -1823,4 +1823,59 @@ extension Program {
...
+    }
+  }
+
+}
+
+/// A member reachable on a receiver, together with its type and how it is referred to.
+public struct MemberCandidate: Sendable {
+
+  /// The declaration of the member.
+  public let declaration: DeclarationIdentity
```
RishabhRD: I am worried that name and doc just repeats type name. Maybe `member` is better name?

Sources/FrontEnd/Program.swift
```diff
@@ -1823,4 +1823,59 @@ extension Program {
...
+
+  /// How the member is referred to (direct, bound member, or inherited via extension/conformance).
+  public let reference: DeclarationReference
+
+  /// The type of the member as selected on the receiver, with conformance substitutions applied.
+  public let type: AnyTypeIdentity
```
RishabhRD, line 1988: How about `signature` name? Intent is not repeating the type name.

Tests/FrontEndTests/MemberListingTests.swift
tothambrus11: Changed to `not on an instance of A`. Removed "must be honored in the conformance branch".

Tests/FrontEndTests/MemberListingTests.swift
kyouko-taiga: What about a for loop then? I don't think combining multiple functional transformations is simpler. `compactMap` is not helping readability in most cases, including this one IMO.

Sources/FrontEnd/Program.swift
tothambrus11: I refactored the struct to have 2 fields: a declaration and a type of the member. I would not name the field `member` because it would suggest that a member has a member, while in reality, a member has a type and a declaration.

Sources/FrontEnd/Typer/Typer.swift
tothambrus11: Should we mention that this method is specifically for user-facing member enumeration, and it only returns true if the member would be reachable through user-written names? Should we also rename the `members(of:...)` methods to reflect that it's only listing user-facing names?

Tests/FrontEndTests/MemberListingTests.swift
tothambrus11: Thanks for the explanation, that makes sense!

Tests/FrontEndTests/MemberListingTests.swift
tothambrus11: I find `compactMap` more readable than an iterative loop, I guess this is subjective. Anyway, I got rid of this altogether, as I found that it's sufficient to return a list of declaration references instead of saving the type, declaration and declaration reference in structs.

Sources/FrontEnd/Typer/Typer.swift
tothambrus11: I am not sure, I copied this from the corresponding `resolve` function: `resolve(_,inheritedMemberOf:visibleFrom:statically:)`. I only changed the control flow slightly, so that we do the collection exhaustively, rather than filtering by name.
It would be nice to abstract over the two somehow...

Sources/FrontEnd/Typer/Typer.swift
kyouko-taiga: I see. We have `where` for that:
```
fun is_big<where let threshold: Int>(x: Int) -> Bool { x > summon[Int] }
```

Sources/FrontEnd/Typer/Typer.swift
kyouko-taiga: `summon` gives a set of ways in which one can resolve an instance of a particular type. In your case, you're using it to find whether we can summon a witness of `q` conforming to some trait visible in `scopeOfUse`. If you find one such instance, then you can assume that all requirements of that trait are inherited members, and there is no need to look at the others.

For example, consider the following:
```hylo
trait P { fun m() -> Int }

given w0: Void is P { fun m() { 1 } }
given w1: Void is P { fun m() { 2 } }
```

Here, `summon` will give you both `w0` and `w1`, because neither is smaller than the other. Hence, the first iteration of the loop concludes that `m` is member of `Void` from the fact that `w0` witnesses `P<Void>`, and the second iteration concludes the same from the fact that `w1` also witnesses `P<Void>`.

In the normal resolution path, we have to enumerate the two candidates because we have to later let the typer emit a diagnostic complaining about the ambiguity. But presumably you're not interested to do the same. You _could_ remove a member from the return value if you know that resolution would fail to resolve it, but I suspect you don't. Is that correct?

Sources/FrontEnd/Typer/Typer.swift
tothambrus11: Thanks, the new name is much less ambiguous!

PR #223, opened 2026-06-28
Implement the serialization/deserialization of the IR

kyouko-taiga:


Sources/FrontEnd/Parser/Parser.swift
```diff
@@ -57,7 +57,15 @@ public struct Parser {
   internal consuming func parseTopLevelDeclarations(in file: inout Module.SourceContainer) {
     assert(file.roots.isEmpty)
     var roots: [DeclarationIdentity] = []
-    while peek() != nil {
+
+    while true {
+      // Ignore leading semicolons.
+      discard(while: { (t) in t.tag == .semicolon })
```
tothambrus11, line 63: It's a bit inconsistent that `; let a = 0` is allowed but `return ();` at the end of a block is not. I think we could be lenient about extra `;` on both sides and leave it to the formatter to get rid of unnecessary ones.

Here is a test case that we could add to the suite:

```
//! stage:typing no-std

; let o = () // works

fun f() -> {Void, Void} {
  // Extra semicolons are legal.
; let a = ();  
  let b = ();;
  return (a, b); // Except here, currently.
};

fun g() -> {Void, Void} {
  // Intermediate statements are not ignored.
  let v = (); let u = (); return (v, u)
};
```

StandardLibrary/StandardLibrary.swift
```diff
@@ -2,14 +2,16 @@ import Foundation
...
 
 /// The path to the bundled standard library's root folder.
 /// 
-/// This folder is meant to be used in distributable builds in order to bundle the standard library
-/// together with the executable. Set the flag `USE_BUNDLED_STANDARD_LIBRARY` to select it.
-public let bundledStandardLibrarySources = Bundle.module.url(forResource: "Sources", withExtension: nil)!
\ No newline at end of file
+/// This folder is intended to be used in distributable builds, in order to bundle the standard
+/// library together with the executable. The driver will pick this folder over the local one if
+/// the its sources are compiled with the flag `USE_BUNDLED_STANDARD_LIBRARY` set.
```
tothambrus11:
```suggestion
/// its sources are compiled with the flag `USE_BUNDLED_STANDARD_LIBRARY` set.
```

Sources/FrontEnd/TreePrinter.swift
```diff
@@ -90,6 +90,11 @@ public struct TreePrinter {
     return (i == 0) ? s : s + superscript(i)
   }
 
+  /// Clears the table ensuring the discrimination of different entities sharing the same name.
```
tothambrus11, line 93: Could you explain what you mean by this comment? How does  "ensuring the discrimination of different entities sharing the same name" happen when we actively remove some from our list? I would expect names to become less unique after this if something can now again take that identifier.

Tests/FrontEndTests/ProgramTests.swift
```diff
@@ -233,13 +235,11 @@ final class ProgramTests: XCTestCase {
     XCTAssertEqual(p.declaration(maybeReferredToBy: xName)?.target?.erased, xd)
   }
 
-  func testReturnsNilTypeIfCouldntTypeCheck() async throws {
-    let s: SourceFile = """
-      let (e, b) = ((), (), ())
-      """
-    var p = Program()
+  func testTypeMaybeAssignedNil() async throws {
```
tothambrus11, line 238: I find this name less suitable for a test case.

A test case name should tell, on a high level, what we are testing. 

For a reader, it should be clear: 
- what are the givens (the setup code for the test)
- what is the action being exercised
- what do we assert about the outcome

Some people annotate their tests with 3 comments (`// GIVEN:`, `// WHEN:`, `// THEN:`) to make this clear. This can be quite verbose for small tests like this, so a descriptive function name can  fulfill this role.

The new name only describes the outcome, but gives no context on why that should be nil.

Tests/FrontEndTests/ProgramTests.swift
```diff
@@ -84,124 +84,126 @@ final class ProgramTests: XCTestCase {
...
-
-    let w2 = try XCTUnwrap(
-      p.select(from: m, .and(.tag(ConformanceDeclaration.self), .name(.init(identifier: "w2"))))
-        .first)
-    XCTAssertEqual(given.erased, w2)
   }
 
-  func testQueryingSourceFileWithName() throws {
-    var p = Program()
+  func testQuerySourceFileWithName() throws {
```
tothambrus11, line 138: Many test names use "test" in a grammatically correct way, e.g., `testXDoesY`. I would keep them grammatically correct for consistency.

Sources/BackEnd/Program+LLVM.swift
```diff
@@ -1117,4 +1126,21 @@ extension Program {
     }
   }
 
+  /// Returns `v` iff it denotes a subscript in `program` known to have an empty slide.
```
tothambrus11: An empty slide doesn't imply an addressor. E.g. in the case when we yield a stack-allocated object. If we consider those having non-empty slide, it would be nice to describe here or in some project-level ABI doc.

Sources/BackEnd/Program+LLVM.swift
```diff
@@ -544,6 +544,15 @@ extension Program {
     return FunctionMetadata(prototype: p, value: v)
   }
 
+  /// Returns the result of `demandFunction(_:in:)` with the identity of `f`, which is declared in
+  /// the module being compiled.
+  private mutating func demandFunction(
```
tothambrus11, line 549: In general I would avoid specifying the behavior of one function in terms of another, unless we want to think of that function as if it was conceptually dependency injected, except we didn't bother to abstract it away as a closure parameter.

The current description ended up being longer than the function body, while telling the exact same information. If we only looked at the implementation, we would have the advantage of being able to hover over `demandFunction(_:in:)` to see its specification that we were referring to. 

In an earlier conversation on Slack, I was advocating for the utility of such specification dependencies, and I see the value behind them in terms of avoiding inlining the specs of all of our callees, which applies here quite well because the dependency has quite a lot of non-trivial docs that we shouldn't replicate.

So I am not sure what's the right move here, probably to do nothing. Anyway, food for thought...
 
[Reusing specification - Hylo.pdf](https://github.com/user-attachments/files/29480733/Reusing.specification.-.Hylo.pdf)

Sources/FrontEnd/Module.swift
```diff
@@ -179,6 +179,11 @@ public struct Module: Sendable {
       _modify { yield &functions.values[f] }
     }
 
+    /// Returns the identity of the function named `name`, if it exists in the module.
+    public func identity(function name: IRFunction.Name) -> Module.ID? {
```
tothambrus11:
```suggestion
    public func identity(function name: IRFunction.Name) -> IRFunction.ID? {
```

tothambrus11: Strongly typed indices could be useful, although this mistake alone doesn't justify them.

Sources/FrontEnd/IR/IRFunction.swift
```diff
@@ -911,3 +894,98 @@ extension IRBlock {
...
+          table[.register(i)] = .register(.init(address: .init(registers)))
+        }
+
+        // Write the instruction.
+        try archive.write(InstructionTag(t), in: &context)
+        try archive.write(c, in: &context)
+
+        registers += 1
+      }
+    }
```
tothambrus11: Credit to Claude `/code-review`; take it with salt & mustard.

`IRFunction.write(to:)` serializes block contents in the wrong order, so reloaded functions are corrupted.                                                                                                                 
                                                                                                                                                                                                                           
  The substitution table numbers blocks and registers in blocks.addresses order, and `init(from:)` reads instructions back in that same order. But the write loop emits block bodies in `DominatorTree` order (for b in         dominance). Whenever dominance order differs from address order — i.e. any function with branches — the reader assigns each block's instruction stream to the wrong block, and register operands resolve to the wrong    instruction.                                                                                                                                                                                                             
                                                                                                                                                                                                                           
  There's a second issue in the same loop: the header writes blocks.count, but the dominance walk skips blocks not reachable from entry. A function with a dead block writes fewer bodies than the header claims, so the     reader runs off the end and reads following bytes as instruction data, corrupting the rest of the archive.                                                                                                               
                                                                                                                                                                                                                           
  `testSerializationWithIR` doesn't catch either one because it only checks blocks.count and the total `instructions().count`, both of which are unchanged when bodies are shuffled between blocks.                            
                                                                                                                                                                                                                           
  Repro: I added `testSerializationPreservesBlockContents`, which compares the per-block opcode sequence before and after a round-trip (so it ignores the register/block renumbering, which is fine, and only flags blocks     whose contents actually moved). It fails on 4 stdlib functions, e.g. `Int.infix`- reloads with its return block and unreachable block swapped.                                                                             
                                                                                                                                                                                                                           
  Fix: drop the dominance walk and write in blocks.addresses order, matching the reader. The only wrinkle is that the table has to be fully built before writing, since an instruction can reference a register defined      later in address order — so it's two passes (build table, then write bodies) instead of building the table as you go. Writing in address order also writes every block, so the blocks.count mismatch goes away too.

## Reproducer
```swift
  func testSerializationPreservesBlockContents() async throws {
    var p = try await Program.withStandardLibrary()
    let m = p.identity(module: Module.standardLibraryName)!

    // Compile the standard library to IR and archive it.
    p.lower(m)
    p.applyTransformationPasses(m)
    let a = try p.archive(module: m)

    // Reload the standard library from the archive.
    var q = Program(forTesting: true)
    let (loaded, n) = try q.load(module: p[m].name, from: a)
    XCTAssert(loaded)

    // The test is only meaningful if some function has more than one basic block, since that is
    // what exercises the dominator-tree vs. block-address ordering in `IRFunction.write(to:)`.
    XCTAssert(
      p[m].ir.functions.values.contains(where: { $0.blocks.count > 1 }),
      "vacuous test: no multi-block function to exercise block ordering")

    // The instructions of each basic block must survive the round trip in the same block and in
    // the same order. We compare the per-block sequence of instruction tags rather than the total
    // instruction count (as `testSerializationWithIR` does), because the total is invariant under
    // bodies being assigned to the wrong blocks. We compare tags rather than the full textual IR
    // because serialization canonically renumbers registers and blocks, so the raw identifiers may
    // legitimately differ -- but the opcodes making up each block must not.
    XCTAssertEqual(p[m].ir.functions.count, q[n].ir.functions.count)
    for (f, g) in zip(p[m].ir.functions.values, q[n].ir.functions.values) {
      let original = f.blocks.addresses.map { (b) in
        f.instructions(in: b).map { InstructionTag(type(of: f.at($0))) }
      }
      let reloaded = g.blocks.addresses.map { (b) in
        g.instructions(in: b).map { InstructionTag(type(of: g.at($0))) }
      }
      XCTAssertEqual(
        original, reloaded,
        "block contents of '\(p.show(f.name))' were corrupted by serialization")
    }
  }
```

Docs/Serialization.md
```diff
@@ -0,0 +1,26 @@
...
+
+## Identities
+
+The representation of a program in the compiler involves three main recursive data structures, namely abstract syntax trees (AST), the type trees, and the static single assignment (SSA) intermediate representation (IR).
+Each of these three data structures is essentially an enumeration implemented similarly.
+Each case of the enumeration is represented by a struct whose instances are stored in some collection.
+The position of an instance in that collection serves as an identity, which may be used to implement composition or aggregation.
+
+These identities have some properties.
+1. They are _stable_ throughout the compilation process. That is, the identity of an instance is never invalidated unless that instance is destroyed.
```
tothambrus11, line 18: I would be curious to see answers to these questions possibly in a linked note:
- How do we handle invalidation? What happens on invalidated index access? Defined crash or possibly wrong read?
- What are examples of some actions that may destroy an instance? E.g. copying out the instance from program and subsequently dropping it probably doesn't count, and some parts of the program cannot ever be destroyed unless we destroy the whole program.

tothambrus11: See related issue: https://github.com/hylo-lang/hylo-new/issues/228
We can add this later.

tothambrus11 (changes requested):
Very nice work! I saw that IR lowering of standard library finally works, which is awesome!

I left some minor inline comments regarding naming and docs.
After my review, I also run Claude on the PR, which may have found a bug.

Sources/FrontEnd/Parser/Parser.swift
kyouko-taiga: Sure, I'll let you make another PR.

Tests/FrontEndTests/ProgramTests.swift
kyouko-taiga: We do not apply this convention.

Sources/FrontEnd/TreePrinter.swift
kyouko-taiga: If the printer is about to write `T` and it has already seen `T` for a different entity then it adds a superscript.

Tests/FrontEndTests/ProgramTests.swift
kyouko-taiga: The previous name was also not telling enough information. `testReturnsNilDeclarationIffNameIsNotResolved` does say what returns `nil` , only that something returns `nil` when some name is not resolved.

Almost all other test cases are named after the method that they test, with sometimes a suffix. I prefer we keep this convention.

Other testing framework allow for more descriptive descriptions in the test reports. That is not the case with XCTest. If you want more information about a specific test, feel free to add a comment.

Sources/BackEnd/Program+LLVM.swift
kyouko-taiga: It does at the codegen level. All Hylo IR instructions that may be in a slide (of a Hylo IR subscript) will be ghost instructions in an addressor, and therefore the slide of the compiled form will be empty.

Sources/FrontEnd/IR/IRFunction.swift
kyouko-taiga: Unfortunately Claude is not entirely correct. While it is true that the construction of the table is buggy, a correct fix is much simpler than the proposed solution. We simply have to construct the table using the same order as the subsequent iteration.

Visiting blocks in the order of addresses is wrong and will lead to a violation of SSA invariants during deserialization because we may encounter a use before its definition. Building the substitution table in two passes won't fix this problem.

The second reported issue is also inaccurate. There cannot be unreachable blocks at this point because serialization occurs after DCE, which is a mandatory pass. So it's even suspicious that Claude reports having "found" a bug showing that some unreachable block of `Int.infix-` has been improperly serialized. I supposed it got tricked by the fact that there is a call to the  `fatal_error`, which returns `Never`, in that basic block.

You know my stance about AI and this Claude comment only strengthen my belief that sending bug reports discovered by a robot without human intervention is a poor use of our time. I am convinced that you can figure out as well as me that half of this review comment is dubious at best. So moving forward, if you want to keep using this technology, I suggest you only use it as a way to point to places where extra scrutiny might be deserved.

Tests/FrontEndTests/ProgramTests.swift
```diff
@@ -233,13 +235,11 @@ final class ProgramTests: XCTestCase {
...
-  func testReturnsNilTypeIfCouldntTypeCheck() async throws {
-    let s: SourceFile = """
-      let (e, b) = ((), (), ())
-      """
-    var p = Program()
+  func testTypeMaybeAssignedNil() async throws {
+    var p = Program(forTesting: true)
     let m = p.demandModule("Main")
-    _ = p[m].addSource(s)
+    _ = p[m].addSource("let (e, b) = ((), (), ())")
```
tothambrus11:
```suggestion
    // When expression fails to type check:
    _ = p[m].addSource("let (e, b) = ((), (), ())")
```

Sources/FrontEnd/TreePrinter.swift
tothambrus11: Thanks! Ah, I was confused because I read the sentence as if there was a comma before "ensuring". It is already grammatically correct, but `that ensures` could help avoid the ambiguity. Or something like this:

```suggestion
  /// Resets the discriminator table so a subsequent rendering disambiguates
  /// same-named entities starting from scratch.
```

kyouko-taiga: I've added text in the detailed docs.

PR #224, opened 2026-06-28
Fix yield coherence checking BFS recursion criteria

tothambrus11:
The recursive step checked the wrong variable for visited, so some problems, like the one in the attached test case were never diagnosed.

### DIFF
--- a/Tests/CompilerTests/negative/illegal-nested-yield.hylo
+++ b/Tests/CompilerTests/negative/illegal-nested-yield.hylo
@@ -1,0 +1,11 @@
+//! stage:lowering
+
+subscript f(x: Bool, y: Bool) -> Bool {
+  yield false
+  if x {
+    if y {
+      // Second yield is nested in two basic blocks after the first yield.
+      yield true
+    }
+  }
+}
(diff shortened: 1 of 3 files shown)


Tests/CompilerTests/negative/illegal-nested-yield.hylo
```diff
@@ -0,0 +1,10 @@
+//! stage:lowering
+// A second yield nested two basic blocks past the first one.
+subscript f(x: Bool, y: Bool) -> Bool {
+  yield false
+  if x {
+    if y {
+      yield true
+    }
+  }
+}
```
kyouko-taiga:
```suggestion

subscript f(x: Bool, y: Bool) -> Bool {
  yield false
  if x {
    if y {
      // Second yield is nested in two basic blocks after the first yield.
      yield true
    }
  }
}
```

kyouko-taiga: Didn't understand your comment. Is that what you meant?

Also I think it's better to place the documentation inline here, but maybe that's only because I didn't understand the comment fully.

kyouko-taiga (changes requested):
Good catch.

I'm just having trouble understanding the comment in the test.

Tests/CompilerTests/negative/illegal-nested-yield.hylo
tothambrus11: Yes, exactly. I'm not sure if nesting is the right word to refer to when using basic blocks, but it's at least nested in source, and it's on a deeper level of the CFG traversal.

PR #229, opened 2026-07-01
Fix overflow for names with introducer.

### LINKED ISSUE #225: Mangler tries to ecnode `introducer` using tag that would overflow `Base64Digit`
Here we union the bits of tag with `0x80` if `introducer is non-nil`, trying to encode its presence as a new bit. However, even though a byte would fit this perfectly, this violates the base64 digit's precondition (`<64`), so this would crash.
```swift
  /// Writes the mangled representation of `name` to `output`.
  private mutating func append(name: Name, to output: inout ManglingContext) {
    // Only encode notation and introducer; labels are encoded in types.
    // Encode the presence of introducer in the tag.
    var tag = UInt8(name.notation.rawValue)
    if name.introducer != nil { tag = tag | 0x80 }
    output.add(base64Digit: tag)
    if let i = name.introducer {
      output.add(base64Digit: i)
    }
    output.add(string: name.identifier)
  }
```

Currently, this seems to be dead code - I didn't find any call sites that could have possibly have a non-nill introducer. How should we deal with this?

lucteo:
The base64 digit needs to have values < 64. The flag lead to overflow.

Fixes #225.

Unfortunately, creating a unit test for this isn't easy.

### DIFF
--- a/Sources/FrontEnd/IR/Mangling/ManglingEncoding.swift
+++ b/Sources/FrontEnd/IR/Mangling/ManglingEncoding.swift
@@ -1197,7 +1197,7 @@ internal struct ManglingEncoding: Sendable {
     // Only encode notation and introducer; labels are encoded in types.
     // Encode the presence of introducer in the tag.
     var tag = UInt8(name.notation.rawValue)
-    if name.introducer != nil { tag = tag | 0x80 }
+    if name.introducer != nil { tag = tag | 0x20 }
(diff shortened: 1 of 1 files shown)


tothambrus11 (approved):

tothambrus11:
Nice, OperatorNotation fits in 2 bits, so we have 4 more bits to use freely while staying in a base-64 digit. Thanks for the fix!

PR #232, opened 2026-07-02
Give a predefined identity to 'Hylo.Never'

kyouko-taiga:
This PR gives `Hylo.Never` a predefined identity so that one longer needs to mutate a type store to get an instance.

### DIFF
--- a/Sources/FrontEnd/Types/TypeStore.swift
+++ b/Sources/FrontEnd/Types/TypeStore.swift
@@ -24,10 +24,19 @@ public struct TypeStore: Sendable {
   /// The identifier of the next fresh variable.
   private var nextFreshIdentifier: Int
 
+  /// The identity of a type equal to `Hylo.Never`.
+  private var never: AnyTypeIdentity
+
   /// Creates an empty instance.
   public init() {
     self.types = []
     self.nextFreshIdentifier = 0
+    self.never = .init(uncheckedFrom: .error)
+
+    // Create `Hylo.Never`.
+    let p = demand(GenericParameter.nth(0, .proper))
+    let u = insert(UniversalType(parameters: [p], head: p.erased))
+    self.never = .init(uncheckedFrom: u)
   }
 
   /// Returns a number less than the number of types created with this store.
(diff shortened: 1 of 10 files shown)


tothambrus11 (approved):
Very nice change, this will speed up both the LSP in some cases and the mangling.

kyouko-taiga:
Unfortunately it seems I have triggered UB...

Sources/FrontEnd/Types/TypeStore.swift
```diff
@@ -677,6 +686,8 @@ public struct TypeStore: Sendable {
         yield ErrorType()
       case AnyTypeIdentity.void.offset:
         yield Tuple.empty
+      case AnyTypeIdentity.never.offset:
+        yield self[never]
```
tothambrus11:
```suggestion
        yield types[never.offset].wrapped
```
This seems to fix the UB. I'm investigating if this is a compiler bug.

kyouko-taiga: Nice fix!

I have implemented something else in parallel that seems to pass the tests locally. Let's see how the CI behaves. But I'm concerned about the UB.

tothambrus11: It is indeed a compiler bug.

```swift
struct Store {
  var a: Any { 42 }
  var b: Any { _read { yield a } }
  subscript(n: Int) -> Any {
    _read {
      if n == 0 { yield b } else { yield self[0] }
    }
  }
}
print(Store()[1] is Int)
```
This crashes similarly, only in release mode.

tothambrus11: https://github.com/swiftlang/swift/issues/90377

kyouko-taiga: We should consider removing some of these `_read` accessors. I think they are in fact slowing us down in particular cases.

I will write an issue. Investigating these tweaks is a not-so-complicated task some enthusiastic contributor could do.

kyouko-taiga: > I will write an issue. Investigating these tweaks is a not-so-complicated task some enthusiastic contributor could do.

#234

PR #236, opened 2026-07-03
Interpreter: Restructuring so that program counter becomes part of stack frame

### LINKED ISSUE #220: Consider interpreter restructuring
> I keep thinking this could be easier to describe as “The ephemeral (or non-`Memory`) execution state of a function call.”  To do that would require storing the program counter for the function in the stack frame, rather than the return address.  That would allow the other storage of the program counter to disappear. 

 _Originally posted by @dabrahams in [#187](https://github.com/hylo-lang/hylo-new/pull/187/changes/BASE..082af8a776edb461f76673d2a8c492c31b3b07f9#r3405199392)_

RishabhRD:
Fixes #220

### DIFF
--- a/Sources/Interpreter/Interpreter.swift
+++ b/Sources/Interpreter/Interpreter.swift
@@ -18,7 +18,7 @@ private struct InstructionPointer {
     step = i
   }
 
-  /// Creates an instance pointing to the first instruction of `f` defined in `p`.
+  /// Creates an instance pointing to the first instruction of `f`, which is defined in `p`.
   public init(interpreting f: GlobalFunctionIdentity, definedIn p: Program) {
     precondition(p[f.module].functions[f.function].isDefined)
     let i = p.entry(f)!
--- a/Sources/Interpreter/Interpreter.swift
+++ b/Sources/Interpreter/Interpreter.swift
@@ -41,6 +41,8 @@ private struct GlobalFunctionIdentity {
 extension Program {
 
   /// Returns first instruction of `f`.
+  ///
+  /// - Precondition: `self` is sufficiently lowered for interpretation.
   fileprivate func firstInstruction(_ f: GlobalFunctionIdentity) -> AnyInstructionIdentity {
     let fn = self[f.module].functions[f.function]
     return fn.blocks[fn.entry!].first!
--- a/Sources/Interpreter/Interpreter.swift
+++ b/Sources/Interpreter/Interpreter.swift
@@ -73,10 +73,10 @@ private struct Stack {
     frames.append(f)
   }
 
-  /// Removes the top frame and returns next instruction to be executed.
-  public mutating func pop() -> InstructionPointer {
+  /// Removes the top frame.
+  public mutating func pop() {
+    precondition(!isEmpty)
     frames.removeLast()
-    return frames.last!.programCounter
   }
 
   /// The top stack frame.
@@ -226,7 +226,8 @@ public struct Interpreter {
     case let x as IRProperty:
       _ = x
     case is IRReturn:
-      return .jump(to: popStackFrame())
+      callStack.pop()
+      return .jump(to: programCounter)
     case let x as IRStore:
       _ = x
     case let x as IRSubfield:
@@ -246,16 +247,6 @@ public struct Interpreter {
     }
     unreachable("Unimplemented processing of instruction")
 
-  }
-
-  /// Removes topmost stack frame and return code pointer to next instruction of any
-  /// previous stack frame.
-  ///
-  /// - Precondition: the program is running.
-  private mutating func popStackFrame() -> InstructionPointer {
-    // precondition(topOfStack.allocations.isEmpty,
-    //     "Function returns before deallocating all local variable storage")
-    return callStack.pop()
   }
 
   /// Moves the program counter to the next instruction.
--- a/Sources/Interpreter/Interpreter.swift
+++ b/Sources/Interpreter/Interpreter.swift
@@ -4,22 +4,22 @@ import Utilities
 /// The position of an instruction in the program.
 private struct InstructionPointer {
 
-  /// A defined function in program.
-  public let functionInProgram: GlobalFunctionID
+  /// The function containing the instruction.
+  public let container: GlobalFunctionIdentity
 
-  /// The position relative to `functionInModule` indicated by `self`.
-  var instructionInFunction: AnyInstructionIdentity
+  /// The instruction designated by `self`, relative to `container`.
+  var step: AnyInstructionIdentity
 
   /// Creates an instance pointing to `i` in `f`.
   ///
   /// - Precondition: `f` is defined.
-  public init(_ i: AnyInstructionIdentity, in f: GlobalFunctionID) {
-    functionInProgram = f
-    instructionInFunction = i
+  public init(_ i: AnyInstructionIdentity, in f: GlobalFunctionIdentity) {
+    container = f
+    step = i
   }
 
-  /// Creates an instance pointing to entry instruction of `f` defined in `p`.
-  public init(toEntryOf f: GlobalFunctionID, definedIn p: Program) {
+  /// Creates an instance pointing to the first instruction of `f` defined in `p`.
+  public init(interpreting f: GlobalFunctionIdentity, definedIn p: Program) {
     precondition(p[f.module].functions[f.function].isDefined)
     let i = p.entry(f)!
     self = .init(i, in: f)
@@ -28,7 +28,7 @@ private struct InstructionPointer {
 }
 
 /// A unique function in a `Program`.
-private struct GlobalFunctionID {
+private struct GlobalFunctionIdentity {
 
   /// The module containing `self`.
   public let module: Module.ID
@@ -41,7 +41,7 @@ private struct GlobalFunctionID {
 extension Program {
 
   /// Returns first instruction of `f`.
-  fileprivate func entry(_ f: GlobalFunctionID) -> AnyInstructionIdentity? {
+  fileprivate func entry(_ f: GlobalFunctionIdentity) -> AnyInstructionIdentity? {
     let fn = self[f.module].functions[f.function]
     guard let e = fn.entry else { return nil }
     return fn.blocks[e].first!
@@ -89,10 +89,10 @@ private struct Stack {
   /// Local variables, parameters, and return addresses.
   private var frames: [StackFrame] = []
 
-  /// Adds a frame for a call to the nullary `f` defined in `p`.
-  public mutating func enter(_ f: GlobalFunctionID, definedIn p: Program) {
+  /// Adds a frame for a call to the nullary `f`, a nullary function defined in `p`.
+  public mutating func enter(_ f: GlobalFunctionIdentity, definedIn p: Program) {
     // TODO: support parameters.
-    let s = InstructionPointer(toEntryOf: f, definedIn: p)
+    let s = InstructionPointer(interpreting: f, definedIn: p)
     let f = StackFrame(currentStep: s)
     frames.append(f)
   }
@@ -115,7 +115,7 @@ private struct Stack {
     }
   }
 
-  /// Number of frames in the call stack.
+  /// The depth of call stack.
   public var count: Int {
     frames.count
   }
@@ -130,10 +130,10 @@ private struct Stack {
 /// A virtual machine that executes Hylo's in-memory IR representation.
 public struct Interpreter {
 
-  /// The program to be executed.
+  /// The program being executed.
   private let program: Program
 
-  /// Next instruction to be executed.
+  /// The next instruction to execute.
   private var programCounter: InstructionPointer {
     _read {
       yield topOfStack.currentStep
@@ -164,9 +164,9 @@ public struct Interpreter {
   /// - Precondition: the program is running.
   public var currentInstruction: any Instruction {
     _read {
-      yield program[programCounter.functionInProgram.module]
-        .functions[programCounter.functionInProgram.function]
-        .at(programCounter.instructionInFunction)
+      yield program[programCounter.container.module]
+        .functions[programCounter.container.function]
+        .at(programCounter.step)
     }
   }
 
@@ -192,7 +192,7 @@ public struct Interpreter {
     if case .jump(let pc) = r {
       programCounter = pc
     } else if case .initializeRegister(let v) = r {
-      topOfStack.registers[programCounter.instructionInFunction] = v
+      topOfStack.registers[programCounter.step] = v
       try advanceProgramCounter()
     }
 
@@ -266,17 +266,17 @@ public struct Interpreter {
   /// Moves the program counter to the next instruction.
   private mutating func advanceProgramCounter() throws {
     guard
-      let i = program[programCounter.functionInProgram.module]
-        .functions[programCounter.functionInProgram.function]
-        .instruction(after: programCounter.instructionInFunction)
+      let i = program[programCounter.container.module]
+        .functions[programCounter.container.function]
+        .instruction(after: programCounter.step)
     else { throw IRError() }
-    programCounter.instructionInFunction = i
+    programCounter.step = i
   }
 }
 
 extension Program {
   /// The function whose invocation executes the whole program.
-  fileprivate var entry: GlobalFunctionID {
+  fileprivate var entry: GlobalFunctionIdentity {
     let entryModule = identity(module: "Main")!
     let entryFunctionDeclaration = cast(
       select(
--- a/Sources/Interpreter/Interpreter.swift
+++ b/Sources/Interpreter/Interpreter.swift
@@ -46,7 +46,7 @@ private enum InstructionEpilogue {
 
 }
 
-/// The ephermal (on non-Memory) execution state of a function call.
+/// The ephemeral (or non-`Memory`) execution state of a function call.
 private struct StackFrame {
 
   // TODO: add local variables and parameters, which require `Memory`.
--- a/Sources/Interpreter/Interpreter.swift
+++ b/Sources/Interpreter/Interpreter.swift
@@ -40,7 +40,7 @@ private struct GlobalFunctionIdentity {
 
 extension Program {
 
-  /// Returns first instruction of `f`.
+  /// Returns the first instruction of `f`.
   ///
   /// - Precondition: `self` is sufficiently lowered for interpretation.
   fileprivate func firstInstruction(_ f: GlobalFunctionIdentity) -> AnyInstructionIdentity {
@@ -93,7 +93,7 @@ private struct Stack {
   /// Local variables, parameters, and return addresses.
   private var frames: [StackFrame] = []
 
-  /// Adds a frame for a call to the nullary `f`, a nullary function defined in `p`.
+  /// Adds a frame for a call to `f`, a nullary function defined in `p`.
   public mutating func enter(_ f: GlobalFunctionIdentity, definedIn p: Program) {
     // TODO: support parameters.
     let s = InstructionPointer(interpreting: f, definedIn: p)
@@ -139,12 +139,8 @@ public struct Interpreter {
 
   /// The next instruction to execute.
   private var programCounter: InstructionPointer {
-    _read {
-      yield topOfStack.currentStep
-    }
-    _modify {
-      yield &topOfStack.currentStep
-    }
+    get { topOfStack.currentStep }
+    set { topOfStack.currentStep = newValue }
   }
 
   /// `true` iff the program is still running.
--- a/Sources/Interpreter/Interpreter.swift
+++ b/Sources/Interpreter/Interpreter.swift
@@ -68,6 +68,9 @@ private enum InstructionEpilogue {
   /// Control is transferred to the given instruction.
   case jump(to: InstructionPointer)
 
+  /// Control is transferred back to the caller.
+  case `return`
+
 }
 
 /// The ephemeral (or non-`Memory`) execution state of a function call.
@@ -144,7 +147,7 @@ public struct Interpreter {
   }
 
   /// `true` iff the program is still running.
-  public var isRunning: Bool { callStack.count > 1 }
+  public var isRunning: Bool { !callStack.isEmpty }
 
   /// Local variables, parameters and return address.
   private var callStack = Stack()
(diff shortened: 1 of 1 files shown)


Sources/Interpreter/Interpreter.swift
```diff
@@ -146,15 +154,17 @@ public struct Interpreter {
...
-    programCounter = .init(
+
+    let pc = InstructionPointer(
       module: e.module,
       functionInModule: e.function,
       instructionInFunction: i
     )
+    /// The program counter of bottom-most frame would never be used, so we
+    /// fill it with something arbitrary.
+    callStack.push(entry: pc)
```
RishabhRD: to remove this, `InstructionEpilogue` needs to have a case for `return`. IDK if that would be worthy?

dabrahams: Seems consistent with having a case for branching.  Can you explain why it needs that case?

Sources/Interpreter/Interpreter.swift
```diff
@@ -67,20 +65,18 @@ private struct Stack {
   /// Local variables, parameters, and return addresses.
   private var frames: [StackFrame] = []
 
-  /// Adds a new frame on top with the given `returnAddress` and `parameters`.
-  public mutating func push(returnAddress: InstructionPointer) {
+  /// Adds a new call frame for a function invocation with `parameters`,
+  /// starting execution at `entry`.
+  public mutating func push(entry: InstructionPointer) {
```
```suggestion
  /// Adds a frame for a call to the function at `f`, passing `parameters`,
  public mutating func push(frameFor f: InstructionPointer) {
```

But `parameters` doesn't exist anywhere yet, so shouldn't be in the documentation:

```suggestion
  /// Adds a frame for a call to the nullary function at `f`.
  public mutating func push(frameFor f: InstructionPointer) {
```

Is it possible to have something like

```suggestion
  /// Adds a frame for a call to the nullary `f`.
  public mutating func enter(_ f: GlobalFunctionID) {
```

?

Sources/Interpreter/Interpreter.swift
```diff
@@ -46,18 +46,16 @@ private enum InstructionEpilogue {
...
   // TODO: add local variables and parameters, which require `Memory`.
 
   /// The results of instructions.
   public var registers: [AnyInstructionIdentity: Value] = [:]
 
-  /// The program counter to which execution should return when
-  /// popping this frame.
-  public var returnAddress: InstructionPointer
+  /// Next instruction in the function.
+  public var programCounter: InstructionPointer
```
```suggestion
  /// The next instruction to execute.
  public var currentStep: InstructionPointer
```

I'm thinking two things:
1. A “program counter” is typically a single global value for a CPU, or at least for a core.
2. How will this read at use sites?

Sources/Interpreter/Interpreter.swift
```diff
@@ -146,15 +154,17 @@ public struct Interpreter {
...
       instructionInFunction: i
     )
+    /// The program counter of bottom-most frame would never be used, so we
+    /// fill it with something arbitrary.
+    callStack.push(entry: pc)
 
-    // The return address of the bottom-most frame will never be used,
-    // so we fill it with something arbitrary.
-    callStack.push(returnAddress: programCounter)
+    callStack.push(entry: pc)
```
dabrahams: As long as this weird thing is still here, it needs a comment that explains why.

dabrahams (approved):
This looks like it will be a big improvement.  Thanks for pursuing it!  Just a few comments inline

Sources/Interpreter/Interpreter.swift
RishabhRD: Made some refactors to support the last suggestion, requesting re-review for that refactoring.

Sources/Interpreter/Interpreter.swift
RishabhRD: `jump` case requires a valid `InstructionPointer` to jump to. But while returning from main we don't have any valid `InstructionPointer` and thus we push a fake frame. Thus `return` case have slightly different and relaxed requirement from `jump`.
Also, `return` doesn't need `InstructionPointer` data like `jump` does as popping frame automatically sets program counter to `currentStep` of `topOfStack` through computed property.

Sources/Interpreter/Interpreter.swift
```diff
@@ -3,15 +3,28 @@ import Utilities
 
 /// The position of an instruction in the program.
 private struct InstructionPointer {
-  /// The module containing `self`.
-  var module: Module.ID
 
-  /// The function in `module` indicated by `self`.
-  var functionInModule: IRFunction.ID
+  /// A defined function in program.
+  public let functionInProgram: GlobalFunctionID
```
tothambrus11: Maybe a better name could be `parent`?

Sources/Interpreter/Interpreter.swift
```diff
@@ -3,15 +3,28 @@ import Utilities
...
 
   /// The position relative to `functionInModule` indicated by `self`.
   var instructionInFunction: AnyInstructionIdentity
```
tothambrus11:
```suggestion
  var offset: AnyInstructionIdentity
```

Sources/Interpreter/Interpreter.swift
```diff
@@ -23,6 +36,13 @@ private struct GlobalFunctionID {
   /// The function in `module` indicated by `self`.
   public let function: IRFunction.ID
 
+  /// Entry instruction of `self` declared in `p`.
+  public func entry(declaredIn p: Program) -> AnyInstructionIdentity? {
```
tothambrus11: I feel this shouldn't be a member of the identity, but rather a member of the whole, e.g. an extension of `Program`.

Sources/Interpreter/Interpreter.swift
```diff
@@ -54,8 +74,8 @@ private struct StackFrame {
   /// The results of instructions.
   public var registers: [AnyInstructionIdentity: Value] = [:]
 
-  /// Next instruction in the function.
-  public var programCounter: InstructionPointer
+  /// The next instruction to execute.
```
tothambrus11, line 85: By the guideline "omit needless words", I would leave out "The".

Sources/Interpreter/Interpreter.swift
```diff
@@ -95,6 +111,11 @@ private struct Stack {
     }
   }
 
+  /// Number of frames in call stack.
```
tothambrus11:
```suggestion
  /// Number of frames in the call stack.
```
And now I'm contradiction myself...

Sources/Interpreter/Interpreter.swift
```diff
@@ -108,11 +129,18 @@ public struct Interpreter {
...
+      yield topOfStack.currentStep
+    }
+    _modify {
+      yield &topOfStack.currentStep
+    }
+  }
 
   /// `true` iff the program is still running.
-  public private(set) var isRunning: Bool = true
+  public var isRunning: Bool { callStack.count > 1 }
```
tothambrus11: Could you describe how this property is established/derived? To me it's not obvious why a stack with `depth >= 2` means the program is executing. We also have `isEmpty`. Are these related?

Sources/Interpreter/Interpreter.swift
RishabhRD: Also, it would also help to maintain postcondition of `applyCurrentInstruction` that it applies `Memory` and `I/O` effects and popping `StackFrame` is neither of those.

Sources/Interpreter/Interpreter.swift
RishabhRD: This has been a suggestion from Dave. Also, I remember in his one of past review he advised to be careful about use of articles. So, IDK if "the" is needless here. @dabrahams can you guide?

Sources/Interpreter/Interpreter.swift
RishabhRD: But somehow `parent` name is not helping me thinking about address to instruction. I believe you suggested this name as this is the role of variable. Can you explain how `parent` is implying the role of variable here?

tothambrus11: I assumed that an instruction is contained within a function. So we identify its parent container first, then take the offset within that to arrive at the instruction pointer to by `self`. I don't have a strong opinion, so feel free to ignore this if you think it's less clear.

Sources/Interpreter/Interpreter.swift
RishabhRD: Currently the bottom-most frame is a fake frame to emulate call to main. We are discussing above if somehow we can get rid of this frame. Then I can make it `!callStack.isEmpty`.

tothambrus11: Thanks, makes sense. No need to block on this from my end, just mention it as a comment.

Sources/Interpreter/Interpreter.swift
dabrahams: That all makes sense to me, thanks.

Sources/Interpreter/Interpreter.swift
kyouko-taiga: I don't understand the purpose of the suffix `inProgram`. All functions are in a program. Also I assume the interpreter loads one program at a time, so a global function ID is necessarily relatively to that program.

kyouko-taiga: We have used `Identity` rather than `ID` pretty consistently in the compiler. The only exception is when `ID` is used as a type alias in protocols (e.g., `Function.ID`).

Following this convention I would rename `globalFunctionID` to either `globalFunctionIdentity` or `IRFunction.GlobalID`.

Sources/Interpreter/Interpreter.swift
kyouko-taiga: `offset` is the the right name. Also I don't think we should say that an instruction identity is a position "relative" to a function. We could say that it is a position *in* a function.

But more simply, `AnyInstructionIdentity` denotes the identity of a function, which is indeed relative to a particular function, just like an instance of `[Int:Int].Index` is relative to a specific instance of `[Int:Int]`.

Also, I don't understand the purpose of the suffix `inFunction` here. An instruction is always in a function. And if we're anticipating the cases where the interpreter may run multiple functions in parallel, then I would find it puzzling that there's only one identity.

Sources/Interpreter/Interpreter.swift
```diff
@@ -3,15 +3,28 @@ import Utilities
...
 
+  /// Creates an instance pointing to `i` in `f`.
+  ///
+  /// - Precondition: `f` is defined.
+  public init(_ i: AnyInstructionIdentity, in f: GlobalFunctionID) {
+    functionInProgram = f
+    instructionInFunction = i
+  }
+
+  /// Creates an instance pointing to entry instruction of `f` defined in `p`.
```
kyouko-taiga:
```suggestion
  /// Creates an instance pointing to the first instruction of `f`, which is defined in `p`.
```

I don't like the term "pointing". The interpreter doesn't really point at any instruction. Its program counter may be *referring* to a particular instruction, which I think does not mean the same exactly.

kyouko-taiga: Also, the term "entry" is consistently used to refer to a basic block in the compiler. We should avoid overloading that meaning.

Sources/Interpreter/Interpreter.swift
```diff
@@ -3,15 +3,28 @@ import Utilities
...
+  public init(toEntryOf f: GlobalFunctionID, definedIn p: Program) {
```
kyouko-taiga:
```suggestion
  public init(interpreting f: GlobalFunctionID, definedIn p: Program) {
```

Sources/Interpreter/Interpreter.swift
dabrahams: Normally a pointer to a thing doesn't have a “parent;” I don't think we have a really good idiom for this sort of “ID path.”  Perhaps

```suggestion
  /// The function containing the instruction.
  public let container: GlobalFunctionID
```

Sources/Interpreter/Interpreter.swift
```diff
@@ -25,6 +38,17 @@ private struct GlobalFunctionID {
 
 }
 
+extension Program {
+
+  /// Returns first instruction of `f`.
+  fileprivate func entry(_ f: GlobalFunctionID) -> AnyInstructionIdentity? {
```
kyouko-taiga: I would expect that "entry" means a basic block, not an instruction.
```suggestion
  fileprivate func firstInstruction(of f: GlobalFunctionID) -> AnyInstructionIdentity? {
```

Sources/Interpreter/Interpreter.swift
dabrahams: I think

```suggestion
  var step: AnyInstructionIdentity
```

is slightly better than `offset`, looking at the use-sites and thinking about the structure of a function, which is non-linear.

Sources/Interpreter/Interpreter.swift
```diff
@@ -25,6 +38,17 @@ private struct GlobalFunctionID {
...
+    let fn = self[f.module].functions[f.function]
+    guard let e = fn.entry else { return nil }
+    return fn.blocks[e].first!
```
kyouko-taiga:
```suggestion
    self[f.module].functions[f.function].flatMap { (g) in
      g.entry.map({ (b) in g.blocks[b].first })
    }
```
Without any precondition It is wrong to force-unwrap the entry block. It's possible that there is an entry block without any instruction.

Sources/Interpreter/Interpreter.swift
kyouko-taiga: I would keep "The" and I that would be the style of all other docs in the compiler.

dabrahams: Leaving out “the” is not consistent with the way we write these docs: “next instruction…” is not a grammatical answer to the question “what is it?”

Also if we wanted to change that we would need to change many other places.

Sources/Interpreter/Interpreter.swift
```diff
@@ -95,6 +115,11 @@ private struct Stack {
     }
   }
 
+  /// Number of frames in the call stack.
```
```suggestion
  /// The depth of the call stack.
```

I like @tothambrus11 's suggestion; it raises the abstraction level

Sources/Interpreter/Interpreter.swift
```diff
@@ -67,20 +89,18 @@ private struct Stack {
   /// Local variables, parameters, and return addresses.
   private var frames: [StackFrame] = []
 
-  /// Adds a new frame on top with the given `returnAddress` and `parameters`.
-  public mutating func push(returnAddress: InstructionPointer) {
+  /// Adds a frame for a call to the nullary `f` defined in `p`.
```
kyouko-taiga:
```suggestion
  /// Adds a frame for a call to`f`, which is a nullary function defined in `p`.
```

Sources/Interpreter/Interpreter.swift
```diff
@@ -108,11 +133,18 @@ public struct Interpreter {
   /// The program to be executed.
   private let program: Program
 
-  /// Identity of the next instruction to be executed.
-  private var programCounter: InstructionPointer
+  /// Next instruction to be executed.
```
```suggestion
  /// The next instruction to be executed.
```

Sources/Interpreter/Interpreter.swift
kyouko-taiga:
```suggestion
  /// The number of frames in the call stack.
```

dabrahams (commented):
Some suggestions

Sources/Interpreter/Interpreter.swift
```diff
@@ -108,11 +133,18 @@ public struct Interpreter {
   /// The program to be executed.
```
kyouko-taiga:
```suggestion
  /// The program being executed.
```

Sources/Interpreter/Interpreter.swift
kyouko-taiga:
```suggestion
  /// The next instruction to execute.
```

Sources/Interpreter/Interpreter.swift
```diff
@@ -143,40 +175,29 @@ public struct Interpreter {
...
-      instructionInFunction: i
-    )
-
-    // The return address of the bottom-most frame will never be used,
-    // so we fill it with something arbitrary.
-    callStack.push(returnAddress: programCounter)
+
+    let f = p.entry
+
+    // The bottom frame is there to emulate `main` is called from some function.
```
kyouko-taiga:
```suggestion
    // The bottom frame is there to emulate the fact that `main` is called from some function.
```

kyouko-taiga (changes requested):
Nice work. I left some inline comments, mostly about documentation.

Sources/Interpreter/Interpreter.swift
dabrahams: @kyouko-taiga In this PR we should stick with whatever convention you have established, but I question the value of that convention, since ID is a universally understood way to express “identity,” and much shorter.

Sources/Interpreter/Interpreter.swift
dabrahams: I don't know what the problem with “pointing” is.  I think calling the type “InstructionReference,” which would be consistent, would be problematic.  It also fails to connote the idea that you can reach an adjacent instruction from that position.

Sources/Interpreter/Interpreter.swift
dabrahams: I don't agree that `offset` is ideal.  It implies a linear sequence of instructions, but that's not how functions are structured.  `position` is probably better, but I think `step` is possibly better than that.

Sources/Interpreter/Interpreter.swift
dabrahams: Getting better!  But a few needless words…

```suggestion
  /// Adds a frame for a call to the nullary `f`, a nullary function defined in `p`.
```

Sources/Interpreter/Interpreter.swift
dabrahams: But it's not a fact.  And nothing explains why it's important to emulate `main` being called from another function, so this comment doesn't do much for me.

@RishabhRD I thought we would be able to eliminate the bottom frame with these changes?

Sources/Interpreter/Interpreter.swift
dabrahams: The original purpose of those `In`… suffixes (which I agree don't work) was to denote that it was relative to the _property_ `function`, not just _some_ function, which is implied by the type.

Sources/Interpreter/Interpreter.swift
dabrahams: at this point is that really a possibility?
I would have thought the precondition that the program is sufficiently lowered for interpretation would have prevented that.  But then I'm surprised at the `Optional` return type.

tothambrus11: We don't change the type of an IRFunction after lowering and transformations are done. I think this is only practical to enforce as a documented+asserted postcondition of lowering. Note that there may be functions without a function body, like `@extern` and later `@c_ffi`, which should be exempt from this.

RishabhRD: Regarding `Optional` return type, I designed function with wider contract i.e., `fn.isDefined` might be `false` as there is no precondition.
In my mind however, if function is defined, then the entry block should have an instruction. And thus the force unwrapping is there. Is this a wrong assumption @kyouko-taiga ?

Sources/Interpreter/Interpreter.swift
RishabhRD: I used `entry` to avoid repeating type name in function name as per guidelines. How about `firstStep` as we used `step` in `InstructionPointer` too?

Sources/Interpreter/Interpreter.swift
RishabhRD: Yes, and now it is gone! Was pending due to our thread about `return` case above.

Sources/Interpreter/Interpreter.swift
RishabhRD: Moved forward with `step` due to Dave's non-linear sequence of instruction argument. Resolving the conversation, if there are other arguments, please feel free to unresolve the conversation.

Sources/Interpreter/Interpreter.swift
RishabhRD: Moving with `container` due to [this](https://github.com/hylo-lang/hylo-new/pull/236#discussion_r3537841321).

Sources/Interpreter/Interpreter.swift
RishabhRD: Is "pointing" blocking for the PR? If so, how about:
```swift
  /// Creates an instance designating the first instruction of `f`, which is defined in `p`.
```

Sources/Interpreter/Interpreter.swift
kyouko-taiga: The convention is the result of having tried to avoid using abbreviations. It is true that "ID" is well established, but I thought `InstructionIdentity` was not worse than `InstructionID` and so I rolled with the former.

Sources/Interpreter/Interpreter.swift
kyouko-taiga: The use of "pointing" is not a blocker for me. The use of "entry" is but I think it has been changed already.

Sources/Interpreter/Interpreter.swift
kyouko-taiga: Instruction is the name of the type but it is also the name of the thing. I don't know what a step is in that context; a function is a collection of instructions, not a collection of steps.

My objection to the term "entry" is because an entry is a basic block.

Sources/Interpreter/Interpreter.swift
kyouko-taiga: I see. It is fine to force-unwrap.

Sources/Interpreter/Interpreter.swift
dabrahams: looking at `at(step)`, `at(position)` probably would have been better.  I'm on the fence and IIUC @kyouko-taiga doesn't love `step`.

Sources/Interpreter/Interpreter.swift
dabrahams: But _should_ the contract be wide?

RishabhRD: Point taken, we should start with stronger preconditions in general. Changed it accordingly.

RishabhRD:
@kyouko-taiga I think I have resolved all your comments, can you take a look? If okay, we can move forward with the PR. Thank You.

Sources/Interpreter/Interpreter.swift
```diff
@@ -25,6 +38,18 @@ private struct GlobalFunctionID {
 
 }
 
+extension Program {
+
+  /// Returns first instruction of `f`.
```
kyouko-taiga:
```suggestion
  /// Returns the first instruction of `f`.
```

Sources/Interpreter/Interpreter.swift
```diff
@@ -67,20 +93,18 @@ private struct Stack {
   /// Local variables, parameters, and return addresses.
   private var frames: [StackFrame] = []
 
-  /// Adds a new frame on top with the given `returnAddress` and `parameters`.
-  public mutating func push(returnAddress: InstructionPointer) {
+  /// Adds a frame for a call to the nullary `f`, a nullary function defined in `p`.
```
kyouko-taiga:
```suggestion
  /// Adds a frame for a call to `f`, a nullary function defined in `p`.
```

Sources/Interpreter/Interpreter.swift
```diff
@@ -105,14 +134,21 @@ private struct Stack {
 /// A virtual machine that executes Hylo's in-memory IR representation.
 public struct Interpreter {
 
-  /// The program to be executed.
+  /// The program being executed.
   private let program: Program
 
-  /// Identity of the next instruction to be executed.
-  private var programCounter: InstructionPointer
+  /// The next instruction to execute.
```
kyouko-taiga, line 140: I am pretty sure that is less efficient than an ordinary getter because copying an `InstructionPointer` is almost free. Also the debugging experience of `_read` accessors is not as good as that of `get` accessors. So unless we have good reasons to think a `_read` accessor is deserved, I think we should prefer ordinary getters.

```suggestion
    get { topOfStack.currentStep }
```

kyouko-taiga (approved):
LGTM, see comments for typos and suggestions.

PR #240, opened 2026-07-05
Interpreter

dabrahams:
I don't see why this should remain divergent from `main` but maybe CI will explain it.


Tests/InterpreterTests/InterpreterTests.swift
```diff
@@ -0,0 +1,20 @@
+import XCTest
+
+final class InterpreterRunTests: XCTestCase {
+
+  func testProgramRuns() async throws {
+    let u = Bundle.module.url(forResource: "InterpreterTestPrograms", withExtension: nil)!
```
kyouko-taiga, line 6: Why are we not re-using the testing architecture? Ultimately I think interpreter should run on all positive tests that go until execution.

Tests/InterpreterTests/Utilities.swift
```diff
@@ -0,0 +1,87 @@
+import Foundation
+import FrontEnd
+import Interpreter
+
+/// Execute `p` on interpreter and prints all the instruction executed.
+private func execute(_ p: Program) throws {
```
kyouko-taiga, line 6:
```suggestion
private func trace(_ p: Program) throws {
```
?

Tests/InterpreterTests/Utilities.swift
```diff
@@ -0,0 +1,87 @@
...
+  var logger = TreePrinter(program: p)
```
kyouko-taiga, line 7: It is not ideal to keep a single instance of the printer to print all instructions. A printer keeps track of the identifiers it has printed to make sure they are rendered without ambiguity. That makes sense while printing a single entity (e.g., one function) but not for an entire program. Also I think we'll "leak" a lot of memory keeping irrelevant information around.

Tests/InterpreterTests/Utilities.swift
```diff
@@ -0,0 +1,87 @@
...
+
+  /// Runs program at `self` on `Interpreter`.
+  func interpret() async throws {
+    var p = try loadProgramWithStandardLibrary()
+    let m = p.identity(module: "Main")!
+
+    try await p.requireNoError(in: m) { await $0.assignScopes(m) }
+    try p.requireNoError(in: m) { $0.assignTypes(m, loggingInferenceWhere: { _, _ in false }) }
+    try p.requireNoError(in: m) { $0.lower(m) }
+    try p.requireNoError(in: m) { $0.applyTransformationPasses(m) }
```
kyouko-taiga, line 26: Could we have the driver run all these steps? The compilation pipeline is getting more complex as we go. It will be a pain to maintain it in different places eventually.

Tests/InterpreterTests/Utilities.swift
```diff
@@ -0,0 +1,87 @@
...
+    let m = p.demandModule("Main")
+    p[m].addDependency(Module.standardLibraryName)
+    _ = p[m].addSource(s)
+    return p
+  }
+
+}
+
+/// Error generated by compiler while processing a program.
+private struct CompilerError: LocalizedError {
```
kyouko-taiga, line 44: The same type is defined elsewhere IIRC. Perhaps we should refactor the testing infrastructure to avoid code duplication.

Tests/InterpreterTests/Utilities.swift
```diff
@@ -0,0 +1,87 @@
...
+    }
+    return r
+  }
+
+}
+
+extension Program {
+
+  /// Throws iff `m` contains error.
+  private func requireNoError(in m: Module.ID) throws {
```
kyouko-taiga, line 63: A similar functionality exists in the driver.

I like the name, though.

Tests/InterpreterTests/Utilities.swift
tothambrus11: I usually use `program.show(entity)`, then you can forget about TreePrinters altogether.

dabrahams: The printing of instructions was intended to be a temporary observability measure while the interpreter was brought up, not a permanent feature, but @tothambrus11's idea sounds good to me

Tests/InterpreterTests/InterpreterTests.swift
dabrahams: Probably so, ultimately. It's early days yet for the interpreter.

Tests/InterpreterTests/Utilities.swift
dabrahams: What does it mean for the interpreter to trace a program as opposed to executing it?

Tests/InterpreterTests/Utilities.swift
dabrahams: It bothers me a little because "require" is usually associated with preconditions.

dabrahams:
Apologies to all for merging this without review! I thought we were in the old Hylo repo!  Maybe we should archive that repo or something...

Tests/InterpreterTests/Utilities.swift
kyouko-taiga: I suggested "trace" because the docs say that the function prints all instructions that have been executed. I think that is the role of a tracer, which is always also an interpreter.

Tests/InterpreterTests/Utilities.swift
kyouko-taiga: Agreed

kyouko-taiga:
No apologies needed.

I'd like to archive the old repo as soon as we compile standard library, which hopefully will happen before I go on holidays next week.

Tests/InterpreterTests/Utilities.swift
dabrahams: Yeah, the doc shouldn't say that.  IMO ultimately if we have tracing it should be an option on execute, eg.

```suggestion
private func execute(_ p: Program, tracing: Bool) throws {
```

PR #242, opened 2026-07-06
Elaborate calls to memberwise initializers

kyouko-taiga:


Tests/CompilerTests/positive/memberwise-init-1.hylo
```diff
@@ -0,0 +1,10 @@
+//! stage:llvm
+
+struct S is Deinitializable {
+  let x: Int
+  memberwise init
+}
+
+public fun main() {
+  _ = S(x: 123)
```
tothambrus11, line 9: Until we don't execute the code, it could be nice to add an IR matching test for the emitted constructor.

PR #251, opened 2026-07-07
Fix the signature of functions returning indirectly

tothambrus11:
Fixes the comment in https://github.com/hylo-lang/hylo-new/pull/176#discussion_r3490597360


Tests/CompilerTests/positive/big-return.hylo
```diff
@@ -0,0 +1,16 @@
+//! exit-status:42
+
+// A type larger than the size of a pointer should be passed correctly as a function parameter.
```
tothambrus11:
```suggestion
// A type larger than a pointer should be passed correctly.
```

PR #252, opened 2026-07-07
Fix the subfield paths of stored properties

tothambrus11:
`forEachStoredProperty` passed its callback the position of each variable within its own binding pattern rather than its offset among all of the struct's individual variable declarations.

### DIFF
--- a/Sources/FrontEnd/Program.swift
+++ b/Sources/FrontEnd/Program.swift
@@ -1375,9 +1375,7 @@ public struct Program: Sendable {
   /// Calls `action` for each stored property declaration in `d`.
   ///
   /// `action` accepts a variable declaration and an index path identifying its abstract position
-  /// in a record value having the type declared by `d`. That position is the offset of the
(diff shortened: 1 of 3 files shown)


Sources/FrontEnd/Program.swift
```diff
@@ -1375,14 +1375,20 @@ public struct Program: Sendable {
   /// Calls `action` for each stored property declaration in `d`.
   ///
   /// `action` accepts a variable declaration and an index path identifying its abstract position
-  /// in a record value having the type declared by `d`.
+  /// in a record value having the type declared by `d`. That position is the offset of the
+  /// property in the flattened list of all the stored properties of `d`, consistent with the
+  /// fields of the record type of instances of `d`.
```
tothambrus11:
```suggestion
  /// in a record value having the type declared by `d`.
```

kyouko-taiga (approved):
LGTM

PR #255, opened 2026-07-08
Fix missing IRProperty InstructionRelevance

tothambrus11:
`IRProperty` was missing its `InstructionRelevance` in `relevanceInSlideOrPlateau`. A lot more instructions are also missing and using the `none` relevance by default. Should we change this to an exhaustive switch and define the instruction relevance for all instructions?

### DIFF
--- a/Sources/BackEnd/SubscriptDecomposition.swift
+++ b/Sources/BackEnd/SubscriptDecomposition.swift
@@ -551,7 +551,8 @@ extension IRFunction {
     switch tag(of: i) {
     case IRAlloca.self, IRProject.self:
       return .captured
-    case IRAccess.self, IRCase.self, IRGlobalAccess.self, IRPlaceCast.self, IRSubfield.self:
+    case IRAccess.self, IRCase.self, IRGlobalAccess.self, IRPlaceCast.self, IRSubfield.self,
+      IRProperty.self:
       return .redefined
     default:
       return .none


Tests/CompilerTests/positive/nested-plateaus.hylo
```diff
@@ -0,0 +1,26 @@
...
+struct Pair is Deinitializable {
+  var x: Int
+  var y: Int
+
+  memberwise init
+
+  fun deinit() sink {
+    discard(n: self.x)
+    discard(n: self.y)
+  }
```
tothambrus11:
```suggestion

fun discard(_ n: sink Int) {}

struct Pair is Deinitializable {
  var x: Int
  var y: Int

  memberwise init

  fun deinit() sink {
    discard(n: self.x)
    discard(n: self.y)
  }
```

kyouko-taiga:
> Should we change this to an exhaustive switch and define the instruction relevance for all instructions?

I don't think so. It may be useful right this moment but eventually that will just be more code.

Tests/CompilerTests/positive/nested-plateaus.hylo
```diff
@@ -0,0 +1,24 @@
+// Deinitializing a local of a type with a user-defined deinit projects the Deinitializable
+// witness and reads its `deinit` member with a `property` instruction. Two locals dying in the
+// same scope thus produce nested plateaus, the inner one having a `property` and an access to it
+// in its prologue.
```
kyouko-taiga: I don't think this level of description is deserved here, and this comment is likely to be falsified down the line if we implement some sort of opptimizations on the inlining of witness tables.

I don't think this test is the right way to test `property` with subscript decomposition. It would be better if we had a more explicit use of subscripts. Sadly there's nothing emitting `property` instructions except witness tables yet, so I understand why you went that way.

My recommendation is to keep the test but modify/remove this comment because ultimately we should test this particular use case of subscript decomposition differently. Perhaps you could open an issue to keep track of that.

kyouko-taiga (approved):
Thanks!

Looks good to me but I think the comment in the test should change. See my comment.

Tests/CompilerTests/positive/nested-plateaus.hylo
kyouko-taiga: In fact, the changes currently in my branch will invalidate your test (but not the semantics of the program). That's because I'm making the emitter smart enough to avoid generating a call through a witness table if it can dispatch the target function at compile-time.

Can we think of another way to test this change? If you only need an `property` instruction to cross a subscript boundary then you could try using a conformance before and after a projection start.

tothambrus11: That didn't work.

## Attempt 1:
```hylo
trait P {
  fun i() -> Void
}

given Int is P {
  fun i() {}
}

@inline(never)
subscript s() -> Int32 {
  yield 0
}

@inline(never)
fun black_hole(x: Int32){}

public fun main() {
  let a = 0

  a.i()
  let p = s[]
  a.i()

  black_hole(p)  // pun not intended
}
```
This just causes 2 separate IRProperty instructions rather than reusing the first one's result.
image

## Attempt 2
```hylo
trait P {
  fun i() -> Void
}

given Int is P {
  fun i() {}
}

@inline(never)
subscript s() -> Int32 {
  yield 0
}

@inline(never)
fun f<T is P>(x: T) {
  x.i()
}

@inline(never)
fun black_hole(x: Int32){}

public fun main() {
  let a = 0

  f(a)
  let p = s[]
  f(a)

  black_hole(p)  // pun not intended
}
```
This just encapsulates the irproperty instruction within `f`, so it's again called twice independently.
image

Do you have any other ideas?

tothambrus11: I think the fix is worth merging even if it's not testable well at the moment.

kyouko-taiga: I will invalidate your test in the next 24h. I don't think it's worth the code churn.

kyouko-taiga: I don't object to merge your fix, but I would remove your test.

kyouko-taiga: Alternatively, we can just wait for my PR to be live. Then we can think of a test.

Issue #248, opened 2026-07-07
no LLVM metadata representation of the type 'CollectionOfOne<Int>'


tothambrus11:
The following program crashes during LLVM lowering:
```hylo
public struct CollectionOfOne<Element> {
  var contents: Element
  memberwise init
}

given <Element is Deinitializable> => CollectionOfOne<Element> is Deinitializable { 
  fun deinit() sink {
    _ = &self.contents
  }
}

public fun main() {
  let _ = CollectionOfOne(contents: 42)
}
```

```
closure #1 in Program.metadata(of:in:) + 2904 in HyloPackageTests.xctest at /home/ambrus/hylo-language-server/hylo-new/Sources/BackEnd/Program+LLVM.swift:1118:9
Program.metadata<A>(of:in:computedWith:) + 890 in HyloPackageTests.xctest at /home/ambrus/hylo-language-server/hylo-new/Sources/BackEnd/Program+LLVM.swift:983:20
Program.metadata(of:in:) + 111 in HyloPackageTests.xctest at /home/ambrus/hylo-language-server/hylo-new/Sources/BackEnd/Program+LLVM.swift:1102:5
Program.metadata(of:in:) + 2054 in HyloPackageTests.xctest at /home/ambrus/hylo-language-server/hylo-new/Sources/BackEnd/Program+LLVM.swift:1007:7
Program.incorporate(_:in:) + 795 in HyloPackageTests.xctest at /home/ambrus/hylo-language-server/hylo-new/Sources/BackEnd/Program+LLVM.swift:212:13
Program.incorporate(_:in:) + 888 in HyloPackageTests.xctest at /home/ambrus/hylo-language-server/hylo-new/Sources/BackEnd/Program+LLVM.swift:155:14
Program.incorporate(from:in:) + 503 in HyloPackageTests.xctest at /home/ambrus/hylo-language-server/hylo-new/Sources/BackEnd/Program+LLVM.swift:141:14
Program.incorporate(contentsOf:into:in:) + 3124 in HyloPackageTests.xctest at /home/ambrus/hylo-language-server/hylo-new/Sources/BackEnd/Program+LLVM.swift:120:7
Program.incorporate(_:in:) + 1927 in HyloPackageTests.xctest at /home/ambrus/hylo-language-server/hylo-new/Sources/BackEnd/Program+LLVM.swift:82:5
[...]

PR #258, opened 2026-07-08
Manual C interop

tothambrus11:
This PR adds a minimal version of C function calling thunk generation, a subset of what's described in https://github.com/orgs/hylo-lang/discussions/1122#discussioncomment-17090235.

The parameter types going through the C function call must match the type layout of the corresponding C function's parameter types (except the outermost layer of pointer indirection). With a concrete example:
```hylo
// Hylo FFI thunk declaration
@c_ffi("eq_int_indirect")
fun eq(x: Int, y: Int) -> Bool
```
```c
// indirect-ABI C function
void eq_int_indirect(intptr_t const* x, intptr_t* y, char* result) {
  *result = *x == *y;
}
```
Here the type layout of `intptr_t` must match `Int`, and `char` must match `Bool`. This concerns size, alignment, bit representation of integers, offsets of stored properties of structs. As a recommendation, we shouldn't replicate struct layout apart from size and alignment, but access all stored properties through accessor functions/subscripts.

Under the hood, the module of the `eq` declaration will get a thunk emitted that adopts the calling convention from indirect to the Hylo calling convention, so that we can freely call `eq` from Hylo code as a regular function and can be used as an environment-free closure.

In a future iteration I am planning to expose the C symbol as a global constant (hopefully typed) C function pointer. We may achieve some neat grouping between the two entities by creating them in a struct, but it may be trickier to keep `apply()` static. I will explore the possibilities for this in a later post.

## Testing framework additions
Now the testing framework compiles and links `*.c` source files in package based tests. It also compiles the standard library libc shims. There is room for optimizations here - these c shims are recompiled on every execution based test run.


Sources/BackEnd/Program+LLVM.swift
```diff
@@ -1121,6 +1208,8 @@ extension Program {
         let l = ConcreteLayout(
           fields: [], propertyToField: Array(fs.indices), size: .fixed(s), alignment: a)
         return TypeMetadata(llvm: v, layout: l)
+      } else if let properties = program.fields(of: t.erased, visibleFrom: ctx.hylo) {
+        return program.metadata(record: n, fields: properties, in: &ctx)
```
tothambrus11, line 1217: Is this correct? (seems work, fixes the crash that `Pointer<Never>` or `Never`  didn't have a metadata representation)

Sources/Utilities/Process+Extensions.swift
```diff
@@ -20,6 +20,19 @@ extension Process {
     /// The arguments passed to the process.
     public let arguments: [String]
 
+    public var description: String {
```
tothambrus11, line 24:
```suggestion
    /// A textual description of the failure.
    public var description: String {
```

Docs/ABI.md
```diff
@@ -0,0 +1,61 @@
+# Abstract Binary Interface
+
+## Hylo Calling Convention
+- `self` is passed as the first argument in case of member functions.
```
kyouko-taiga, line 5:
```suggestion

- `self` is passed as the first argument in case of member functions.
```

Docs/ABI.md
```diff
@@ -0,0 +1,61 @@
...
+// roughly translates to:
+intptr_t f(intptr_t p1, int32_t* p2, LargeStruct const*);
+```
+
+```cpp
+fun g(a: let Vector2, b: let Vector2) -> Vector2
+// roughly translates to: 
+void g(Vector2 const* a, Vector2 const* b, Vector2* result);
+```
+```cpp
```
kyouko-taiga, line 29:
```suggestion

```cpp
```

Docs/ABI.md
```diff
@@ -0,0 +1,61 @@
...
+}
+// roughly translates to:
+float length(P const* self);
+```
+
+### Closure representation
+TODO
+
+### Indirect C Calling Convention
+For C interoperability, we pass every parameter and the return value indirectly through the language
```
kyouko-taiga:
```suggestion

For C interoperability, we pass every parameter and the return value indirectly through the language
```

Docs/ABI.md
```diff
@@ -0,0 +1,61 @@
...
+We can declare an `@extern_c_indirect`-annotated function in Hylo, and implement it in C. In the 
+indirect C calling convention:
+- All parameters are passed indirectly, except zero-sized ones, which are ignored.
+- The return value is passed as the last parameter, except if zero-sized.
+
+Example:
+
+```hylo
+// Hylo FFI thunk declaration
+@c_ffi("eq_int_indirect")
```
kyouko-taiga: I am confused, is it `@c_ffi` or `@extern_c_indirect`?

Sources/FrontEnd/Program.swift
```diff
@@ -1274,6 +1274,22 @@ public struct Program: Sendable {
     }
   }
 
+  /// Returns the name of the C function implementing `d` iff `d` is annotated with 
+  /// `@extern_c_indirect`.
+  ///
+  /// The returned function follows the indirect calling convention: it accepts one pointer for
+  /// each non-erased parameter of `d`, in order, followed by a pointer to the storage receiving
+  /// the result, and returns `void`.
```
kyouko-taiga: This documentation does not really belong here. The function is about returning names; it should not be concerned with signatures.

Sources/FrontEnd/Program.swift
```diff
@@ -1274,6 +1274,22 @@ public struct Program: Sendable {
...
+  public func externCName(of d: DeclarationIdentity) -> String? {
```
kyouko-taiga, line 1279: Do we need to make a distinction between external C names and `@extern`? There was some support already so I wonder if we can re-use it or scrape it.

Sources/BackEnd/Program+LLVM.swift
```diff
@@ -65,14 +65,36 @@ extension Program {
...
     // Don't compile generic functions.
     if !self[ctx.hylo].ir[f].isMonomorphic { return }
-    // Don't compile functions without a definition.
-    if !self[ctx.hylo].ir[f].isDefined { return }
+
+    // Don't compile functions without a definition, except those annotated with
+    // `@extern_c_indirect`, whose declaring module defines the indirect-to-Hylo thunk.
+    if !self[ctx.hylo].ir[f].isDefined {
+      if case .lowered(let d) = self[ctx.hylo].ir[f].name, d.module == ctx.hylo,
+        let foreign = externCName(of: d)
```
kyouko-taiga: I would put all this logic into one function, so that the test is easier to read. I suspect it will make control-flow simpler to express, as I am not a fan of the `return` statement sitting by itself at line 95.

Sources/BackEnd/Program+LLVM.swift
```diff
@@ -944,6 +966,71 @@ extension Program {
     }
   }
 
+  /// Defines the calling convention adopting thunk compiled from a `@extern_c_indirect`
+  /// Hylo declaration `f`, calling `foreign` with indirect C convention.
```
kyouko-taiga: The documentation is confusing. The calling function is defined in the docs and it does not have to be computed at run-time. It is also unclear to me what the function does if I just look at the signature.

kyouko-taiga: Reading the body it gets clearer. I think the documentation should say something along these lines:

> Defines a thunk presenting a Hylo function that calls the C function `foreign`.

I don't know exactly how you want to phrase the stuff about indirections but hopefully my suggestions helps.

Sources/Driver/Driver.swift
```diff
@@ -62,6 +62,14 @@ public struct Driver {
     #endif
   }
 
+  #if USE_BUNDLED_STANDARD_LIBRARY // Set compiler flag in distributable builds.
+  /// The root folder of the standard library's sources.
+  let chosenStandardLibraryRoot = bundledStandardLibrarySources
```
kyouko-taiga: Should this be `public` or `private`?

Sources/Driver/Driver.swift
```diff
@@ -194,20 +202,25 @@ public struct Driver {
   /// - Throws: if the parent folder of `output` doesn't exist.
   public mutating func generateExecutable(
     from module: Module.ID,
+    withCSources cSources: [URL] = [],
```
kyouko-taiga, line 205: Hmm, this parameter suggests to me that the C sources should be part of the module.

Sources/Driver/Driver.swift
```diff
@@ -241,6 +254,22 @@ public struct Driver {
     }
   }
 
+  /// Compiles `source` using `clang` to an object file.
+  ///
+  /// Returns the path to the object file within `destinationDirectory`.
+  public func compileC(source: URL, destinationDirectory: URL) throws -> URL {
+    let o = destinationDirectory.appendingPathComponent(
+      "\(UUID())-\(source.deletingPathExtension().appendingPathExtension("o").lastPathComponent))",
```
kyouko-taiga: I don't like the use of a UUID. I think it will make it very difficult to create reproducible builds.

Sources/Driver/Driver.swift
```diff
@@ -241,6 +254,22 @@ public struct Driver {
     }
   }
 
+  /// Compiles `source` using `clang` to an object file.
+  ///
+  /// Returns the path to the object file within `destinationDirectory`.
+  public func compileC(source: URL, destinationDirectory: URL) throws -> URL {
```
kyouko-taiga: Looking at the call site I find it off that this function does not mention object files. Also it seems we use `compile` with a different meaning when we talk about Hylo source files.

I would rename this function to something suggesting that it is also *writing* object files somewhere.

Sources/BackEnd/Program+LLVM.swift
kyouko-taiga: It will do the job for now but we'll need something more clever. I have ongoing work in a local branch but it's fine to go ahead with this temporary solution.

Sources/FrontEnd/Parser/Parser.swift
```diff
@@ -181,11 +181,29 @@ public struct Parser {
       if let (x, xs) = arguments.headAndTail {
         if x.value != .string("always") && x.value != .string("never") {
           report(expected("'always' or 'never'", at: x.site))
-        } else if let y = xs.first {
+        }
+        if let y = xs.first {
```
kyouko-taiga, line 185: Why? I don't think it is necessarily better to emit more error messages at once.

Sources/FrontEnd/Parser/Parser.swift
```diff
@@ -181,11 +181,29 @@ public struct Parser {
...
+          at: identifier.site))
+        break
+      }
+      if case .string(let v) = x.value {
+        if v.isEmpty {
+          report(expected("non-empty string argument", at: x.site))
+        }
+      } else {
+        report(expected("string argument", at: x.site))
+      }
```
kyouko-taiga, line 202:
```suggestion
      guard case .string(let v) = x.value, !v.isEmpty else {
        report(expected("non-empty string argument", at: x.site))
        break
      }
```
I think it is simpler to read and sufficiently precise as an error message.

Sources/FrontEnd/Typer/Typer.swift
```diff
@@ -624,10 +624,28 @@ public struct Typer {
...
     check(body: program[d].body, of: .init(d), expectingOutputType: a.output)
     checkCaptures(of: d)
+    checkExternCIndirect(d)
 
     // TODO: Redeclarations
   }
 
+  /// Checks invariants of `@extern_c_indirect` if the annotation is present.
+  private mutating func checkExternCIndirect(_ d: FunctionDeclaration.ID) {
+    guard let a = program.annotation("extern_c_indirect", appliedTo: d) else { return }
```
kyouko-taiga, line 634: We are repeating this character string ("extern_c_indirect") in many places. Do you think it may be worth defining a constant?

Sources/FrontEnd/Program.swift
```diff
@@ -1491,13 +1507,14 @@ public struct Program: Sendable {
...
-  /// A declaration requires a definition unless it is a trait requirement, an FFI, an external
-  /// function, or a memberwise initializer.
+  /// A declaration requires a definition unless it is a trait requirement, an external function,
+  /// or a memberwise initializer.
   public func requiresDefinition(_ d: DeclarationIdentity) -> Bool {
     switch tag(of: d) {
     case FunctionDeclaration.self:
       let f = castUnchecked(d, to: FunctionDeclaration.self)
-      return !isRequirement(f) && !isForeign(f) && !isExtern(f) && !self[f].isMemberwiseInitializer
+      return !isRequirement(f) && !isExtern(f) && !isExternCIndirect(f)
```
kyouko-taiga, line 1512: I think we either have to merge `extern` and `externCIndirect` or we have to rename one of the two.

Sources/Utilities/Process+Extensions.swift
```diff
@@ -20,6 +20,20 @@ extension Process {
     /// The arguments passed to the process.
     public let arguments: [String]
 
+    /// A textual description of the failure.
+    public var description: String {
+      """
+      NonZeroExit:\(exitCode)
+      Of command: \(executable) \(arguments.joined(separator: " "))
```
kyouko-taiga:
```suggestion
      \(executable) \(arguments.joined(separator: " ")) exited with status \(exitCode)
```

StandardLibrary/StandardLibrary.swift
```diff
@@ -15,3 +15,6 @@ public let localStandardLibrarySources = URL(fileURLWithPath: #filePath)
 /// its sources are compiled with the flag `USE_BUNDLED_STANDARD_LIBRARY` set.
 public let bundledStandardLibrarySources = Bundle.module.url(
   forResource: "Sources", withExtension: nil)!
+
+/// The file name of the standard library's C shim source file within the standard library root.
+package let cShimSource = "shims.c"
```
kyouko-taiga, line 20: Where is this necessary?

Tests/CompilerTests/positive/c-ffi-calls.package/foreign.c
```diff
@@ -0,0 +1,59 @@
...
+  intptr_t y;
+};
+
+void Point_init_indirect(
+  struct Point* self, intptr_t const* x, intptr_t const* y
+) {
+  self->x = *x;
+  self->y = *y;
+}
+
```
kyouko-taiga, line 16:
```suggestion
```

Tests/CompilerTests/positive/c-ffi-calls.package/foreign.c
```diff
@@ -0,0 +1,59 @@
...
+  struct Point const* a, struct Point const* b, struct Point* result
+) {
+  result->x = a->x + b->x;
+  result->y = a->y + b->y;
+}
+
+void Int_add_indirect(
+  intptr_t const* a, intptr_t const* b, intptr_t* result
+) {
+    *result = *a + *b;
```
kyouko-taiga:
```suggestion
  *result = *a + *b;
```

Tests/CompilerTests/positive/c-ffi-calls.package/LibC.hylo
```diff
@@ -0,0 +1,12 @@
+// These are linked from the standard library shims.
+// Eventually, we should move this file to `StandardLibrary/Sources/`.
```
kyouko-taiga, line 2: Well, we have to think about `--freehosted`

Tests/UtilitiesTests/HostTests.swift
kyouko-taiga, line 62:
```suggestion
XCTAssertThrowsError(try Process.executionOutput(executable, arguments: arguments), "") { (error) in
  if let e = error as? Process.NonzeroExit {
    XCTAssertEqual(e.exitCode, 42)
    XCTAssertEqual(e.executable, executable)
    XCTAssertEqual(e.arguments, arguments)
    XCTAssert(e.description.starts(with: "NonZeroExit:42"))
  } else {
    XCTFail("Expected Process.NonzeroExit, got \(error)")
  }
}
```
Can you use XCTUnwrap rather than implementing pattern matching manually?

kyouko-taiga (changes requested):
Beautiful work!

I'd like to sort out the meaning of `extern` and `externCIndirect`. I suspect we can perhaps merge the two features, or eliminate one of them. I also wonder if `externCIndirect` is the right name and think we should workshop some ideas.

The other concern I have is with the handling of the C files. I strongly believe they should be part of a module so that we don't have to carry extra parameters throughout many APIs.

I left other inline comments and questions.

Docs/ABI.md
tothambrus11: Oh yes, I forgot to update it here.

tothambrus11:
```suggestion
@extern_c_indirect("eq_int_indirect")
```

StandardLibrary/StandardLibrary.swift
tothambrus11: We use this in the driver to add this to the extra compiled C sources unless being in freestanding more. I added this as a variable here so that we make the contract that the file is there more explicit.

Tests/CompilerTests/positive/c-ffi-calls.package/LibC.hylo
tothambrus11: I think we can put these outside the `Core/` folder, and only link `shims.c` and use the files outside of `Core/` when hosted standard library is enabled.

Tests/UtilitiesTests/HostTests.swift
tothambrus11: Unfortunately no, because of a Swift XCTest library bug. The overload of `XCTAssertThrowsError` that accepts a throwing closure is not consistently available on all platforms. https://github.com/swiftlang/swift-corelibs-xctest/issues/543
We discussed this solution here: https://github.com/hylo-lang/hylo-new/pull/93#discussion_r3153948762

Sources/FrontEnd/Program.swift
tothambrus11: `@extern("some_name")` means that the Hylo function's implementation resides in an external source, assuming Hylo ABI. If another language implements Hylo ABI, they can just link to that directly. `@extern_c_indirect` emits an ABI adopting thunk in addition. I don't think we will have uses for `@extern` for the foreseeable future, so we could technically remove that. I'm not sure if I'm aware of all use cases, many languages tend to have an `@extern` equivalent, so I would keep it until we are more confident about our needs.

`@extern` may be used in interface files when a precompiled closed-source library is published. Alternatively, we could just use a `.hylointerface` file to have everything assumed `@extern` by default, so we can reduce the noise.

Sources/BackEnd/Program+LLVM.swift
tothambrus11: Changed to:
```swift
// Don't compile functions without a definition, unless `@extern_c_indirect`-annotated.
guard self[ctx.hylo].ir[f].isDefined else {
 return defineIndirectToHyloThunkIfNeeded(in: f, in: &ctx)
}
```

Sources/BackEnd/Program+LLVM.swift
tothambrus11: Thanks, this makes it clearer that the added function is a Hylo function.

Sources/Driver/Driver.swift
tothambrus11: For developing external modules, we only want the bundled standard library sources. `Driver` is also not an exposed module in this package. We can make this private.

Sources/Driver/Driver.swift
tothambrus11: I don't think we should pollute `Module` with build system related impurities. This is just a temporary solution, and eventually the Hylo compiler should not be responsible for compiling C sources.

Sources/Driver/Driver.swift
tothambrus11: It will be always hard to guarantee global uniqueness of the generated object files in this function and ensure we don't overwrite source files with the same name. For now I changed it to hash the source URL, which is consistent locally, but not across different computers. We still include the original file name, which should be sufficient for diagnosing problems.

tothambrus11: Note that we already use temporary directories with UUIDs in their names for the generated artifacts within a test case. LMK if you want that changed as well. We could probably take the encoded test case name as a uniqueness-ensuring prefix.

Sources/FrontEnd/Parser/Parser.swift
tothambrus11: For the IDE experience it's nicer to be aware of as many high-confidence errors as possible. It's like a mountain you are climbing. You think you see the top, and when you finally get there, it's revealed that it was just a bump and the real mountaintop is much further.

I would also not mandate any arbitrary order in which these high-confidence mistakes are presented to the user when possible. The only exception I see is when we cannot be confident about a certain diagnostic unless some prerequisites are fixed.

Sources/FrontEnd/Parser/Parser.swift
tothambrus11: My reasoning was that if the user types a number, it is enough for them to see that it's supposed to be a string, so I wouldn't bother them with the details. They usually don't need to be informed that it's not supposed to be empty, unless in the unlikely case that they forget to fill it in.

For the reason mentioned in the previous comment, I would also like this to be an independent diagnostic from the annotation having too many arguments, as we can reliably detect both.

Sources/FrontEnd/Typer/Typer.swift
tothambrus11: We would need to repeat it in documentation anyways, so when we change it, we need to do a global search and replace anyways. Do you think it's worth the readability regressions caused by string interpolations and a `externCIndirectAnnotation: String` constant?

Sources/FrontEnd/Program.swift
tothambrus11: Yes, I realize I can just remove this whole section now.

Sources/Driver/Driver.swift
tothambrus11: I renamed it to `compileCToObject`.

Sources/BackEnd/Program+LLVM.swift
tothambrus11: https://github.com/hylo-lang/hylo-new/issues/248

This is where this solution breaks down. Probably the witness has a dynamic size, so `record(fields:in:)` crashes with "unexpected dynamic size". 

It could be nice to add that example to your test cases in your local branch.

Sources/Driver/Driver.swift
tothambrus11: Resolved in private discussion.

> I re-reviewed the question of adding foreign functions to be part of the module and [implemented a version of that](https://github.com/hylo-lang/hylo-new/pull/294/changes#top). I think this is a kind of refactoring that is not worth the intrusiveness, and will need to be refactored anyways when we integrate a proper build system. It overfits the C compilation model (source files can be compiled separately), and would be a lot more intrusive changes following for adding include paths and foreign compiler options per source file. Including the foreign sources in the module bundle is an interesting idea, but I would spend more time researching this before we start implement anything complicated. Build system interoperability is a topic I've researched already for some days, and it never seemed to be as easy of a task to just take a list of source files to compile. E.g. consider the folder structure of a C project - includes needs to be resolved relatively, so just materializing the C sources from text accurately is already a bit involved.

> My current design keeps the compiler take care of purely Hylo concerns and only adds the complexity of foreign interaction within focused places in the driver, which may be easier to get rid of later.

tothambrus11:
I think I addressed all points. We can come back to some of the decisions when the time comes. I'll merge this for now so that we can use this for testing.

PR #267, opened 2026-07-09
Implement synthetic deinitializers for structs

kyouko-taiga:
This PR implements the generation of synthetic deinitializers for structs. It does not yet handle user-defined initializers of empty structs. It does not handle synthetic deinitializers for other data structures (e.g., enums).

### DIFF
--- a/Tests/CompilerTests/negative/recursive-implicit-deinit.diagnostics.expected
+++ b/Tests/CompilerTests/negative/recursive-implicit-deinit.diagnostics.expected
@@ -1,0 +1,3 @@
+negative/recursive-implicit-deinit.hylo:4.22: error: implicit deinitialization of instances of 'S' causes infinite recursion in this context
+  fun deinit() sink {}
+                     ^
--- a/.typos.toml
+++ b/.typos.toml
@@ -3,7 +3,11 @@
 
 [files]
 # Swifty-LLVM is a separate project. Its spelling is managed upstream.
-extend-exclude = ["Swifty-LLVM"]
+# IR source files may contain mangled and generated symbols.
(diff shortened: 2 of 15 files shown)


Tests/CompilerTests/negative/recursive-implicit-deinit.diagnostics.expected
```diff
@@ -0,0 +1,3 @@
+negative/recursive-implicit-deinit.hylo:4.22: error: implicit deinitialization of instances of 'S' causes infinite recursion in this context
+  fun deinit() sink {}
+                     ^
```
tothambrus11, line 3: I understood that we would emit a structural member wise deinit in this case. Did we change direction, or is this a temporary diagnostic until that's implemented?

kyouko-taiga: It is a temporary diagnostic. But I have not yet reached a definite conclusion on what should be done if someone writes a deinitializer that could have been synthesized.

On one had, if we can emit memberwise deinitialization, then the deinitializer was synthesizable all along and there was no need to write a custom one. On the other hand, I guess it is useful to be able to write that:
```
struct S is Deinitializable {
  let x: Int
  fun deinit() sink { print("good bye") }
}
```

tothambrus11: Yeah, I think this is a worthwhile use case. We can emit a static analysis suggestion if somebody happens to exactly reimplement the synthesized behavior and it could be simplified, but there is nothing inherently wrong with that.

PR #269, opened 2026-07-10
Exclude static properties from memberwise initializers

### LINKED ISSUE #213: Filter non-static members when synthesizing memberwise initializers
The following code currently results in a memberwise initializer with 3 members - including the static `nil` member variable.
 
```hylo
struct Point {
  var x: Int
  var y: Int

  memberwise init

  fun infix+(other: Point) -> Point {
    Point(x: self.x + other.x, y: self.y + other.y)
  }

  fun infix-(other: Point) -> Point {
    .new(x: self.x - other.x, y: self.y - other.y)
  }

  static let nil: Point = Point(x: 0, y: 0)

}
```

Image

kyouko-taiga:
Fixes #213

### DIFF
--- a/Sources/FrontEnd/Typer/Typer.swift
+++ b/Sources/FrontEnd/Typer/Typer.swift
@@ -1349,7 +1349,7 @@ public struct Typer {
       // We don't use `program.storedProperties(of:)` because we have to ensure that all stored
       // properties are typed before we can form a corresponding parameter.
       for b in program.collect(BindingDeclaration.self, in: program[s].members) {
-        // Static properties are not excluded.
+        // Static properties are excluded.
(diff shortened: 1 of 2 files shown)


Sources/FrontEnd/Typer/Typer.swift
```diff
@@ -1349,6 +1349,10 @@ public struct Typer {
       // We don't use `program.storedProperties(of:)` because we have to ensure that all stored
       // properties are typed before we can form a corresponding parameter.
       for b in program.collect(BindingDeclaration.self, in: program[s].members) {
+        // Static properties are not excluded.
```
```suggestion
        // Static properties are excluded.
```

tothambrus11:
LGTM, except a likely double negation.

Issue #246, opened 2026-07-06
refactor the testing infrastructure to avoid code duplication.


dabrahams:
The same type is defined elsewhere IIRC. Perhaps we should refactor the testing infrastructure to avoid code duplication.

_Originally posted by @kyouko-taiga in https://github.com/hylo-lang/hylo-new/pull/240#discussion_r3526765396_

tothambrus11:
Fixed in #270

Issue #245, opened 2026-07-06
Could we have the driver run all these steps


dabrahams:
Could we have the driver run all these steps? The compilation pipeline is getting more complex as we go. It will be a pain to maintain it in different places eventually.

_Originally posted by @kyouko-taiga in https://github.com/hylo-lang/hylo-new/pull/240#discussion_r3526762847_

Issue #244, opened 2026-07-06
re-using the testing architecture


dabrahams:
Why are we not re-using the testing architecture? Ultimately I think interpreter should run on all positive tests that go until execution.

_Originally posted by @kyouko-taiga in https://github.com/hylo-lang/hylo-new/pull/240#discussion_r3526746744_

Issue #243, opened 2026-07-06
I usually use `program.show(entity)`, then you can forget about TreePrinters altogether.


dabrahams:
I usually use `program.show(entity)`, then you can forget about TreePrinters altogether.

_Originally posted by @tothambrus11 in https://github.com/hylo-lang/hylo-new/pull/240#discussion_r3527326732_

PR #270, opened 2026-07-10
Interpreter tests using driver

### LINKED ISSUE #239: URL extension
https://github.com/hylo-lang/hylo-new/blob/cdd2b58f5659043599d6bd2578bd97447f1cc80f/Tests/InterpreterTests/Utilities.swift#L19-L39

I don't love these as extensions on `URL`.  It is too general a tool to have Hylo-specific methods like these unless some EDSL justifies it.  `URL("file://foo.bar").interpret()` has no obvious meaning.  Probably they should be moved to `SourceFile`.

Secondly, `interpret` has two logical steps: 
1. Lower the program to an interpretable state
2. Execute the lowered program

The first step should be its own function.

RishabhRD:
Fixes: #239, #243, #245, #246.

Regarding #244, it would need some time as @dabrahams pointed [here](https://github.com/hylo-lang/hylo-new/pull/240#discussion_r3530857082), it is early days of interpreter, so we can run very small subset of programs.

NOTE: The diffs for `Utilities.swift` are little messy as that file is almost completely changed. So, looking at the final updated file without diffs would be more friendly.


tothambrus11:
Regarding #244: we can add a test flag to opt out from the interpreter/llvm based execution. By default, try to run everything with both. Slowly we will be able to remove the opt-outs.

Tests/InterpreterTests/Utilities.swift
```diff
@@ -1,87 +1,48 @@
...
-  /// Returns `Program` at `self` containing `Main` module alongwith standard library.
-  private func loadProgramWithStandardLibrary() throws -> Program {
-    let s = try SourceFile(contentsOf: self)
-    var p = Program()
-    let m = p.demandModule("Main")
-    p[m].addDependency(Module.standardLibraryName)
-    _ = p[m].addSource(s)
-    return p
+  /// Returns a program loaded from `u` and lowered for interpretation.
+  static func loadForInterpretation(from u: URL) async throws -> Program {
```
dabrahams: `loadForInterpretation(from: someURL)` does not tell me what is expected at that URL.  One could imagine a pre-lowered program.  I think we want


```suggestion
  /// Returns the sources rooted at `r`, lowered for interpretation.
  static func loadForInterpretation(sourceRoot r: URL) async throws -> Program {
```

Tests/InterpreterTests/Utilities.swift
```diff
@@ -1,87 +1,48 @@
...
+    var d = Driver(targetSpecification: try .host())
+    try await d.loadStandardLibrary()
+    try await d.load("Main", withSourcesAt: u, dependingOnStandardLibrary: true)
```
dabrahams: I find it weird that we have to both load the standard library _and_ tell the driver that the program depends on it.  One of those _should_ be sufficient.

dabrahams (approved):
couple comments.  Probably @kyouko-taiga should look

Tests/InterpreterTests/Utilities.swift
tothambrus11: Loading a module just adds it to Program. Expressing the dependency graph between modules is a separate concern.

dabrahams:
I would rather see one PR per issue; that would certainly make review easier.

Tests/InterpreterTests/Utilities.swift
tothambrus11: In a helper method we might unify the two, just check if a module with the standard library's name is already present in the program, and load if not.

dabrahams: I realize what the calls do. Obviously I don't think they are separate concerns, when you know it's the standard library being loaded, or I wouldn't have written what I wrote.

RishabhRD: Yes, it will also help me get rid of extension method I wrote. I had same feeling while writing those 2 calls, but I was not sure. Thanks.
As it is a separate concern, let me open a new PR these driver changes.

RishabhRD:
> I would rather see one PR per issue; that would certainly make review easier.

Apologies. I would take care of it in future. I misjudged the complexity based on number of lines. If you want I can break the PR into multiple ones.

Tests/InterpreterTests/Utilities.swift
RishabhRD: #283 is merged and thus marking this comment as resolved.

Sources/Driver/Driver.swift
```diff
@@ -274,18 +274,14 @@ public struct Driver {
...
   /// not set, the module is unconditionally compiled from sources and no archive is stored.
   public mutating func load(_ module: Module.Name, withSourcesAt root: URL) async throws {
-    // Compute a fingerprint of all source files.
-    var sources: [SourceFile] = []
-    try SourceFile.forEach(in: root) { (s) in
-      sources.append(s)
-    }
+    let sources = try sources(at: root)
 
     // Attempt to load the module from disk.
```
dabrahams, line 288: Lines like this are a red flag that the section following should probably be a separate function with a name that communicates what's in the comment.  There's another like this later in the function.

Sources/Driver/Driver.swift
```diff
@@ -334,12 +334,14 @@ public struct Driver {
...
 
-  /// Loads the standard library with `load(_:withSourcesAt:)`.
+  /// Loads the standard library with `load(_:withSourcesAt:)` and makes
+  /// standard library a dependency of modules loaded thereafter.
   /// 
   /// Use the `USE_BUNDLED_STANDARD_LIBRARY` compiler flag to control whether the  bundled or local
   /// standard library is used. Defaults to local.
   public mutating func loadStandardLibrary() async throws {
     try await load(Module.standardLibraryName, withSourcesAt: chosenStandardLibraryRoot)
+    noStandardLibrary = false
```
dabrahams: Seems to me maybe `noStandardLibrary` ought to be inverted and spelled `standardLibraryLoaded`, at least based on the few use sites I've seen.

Sources/Driver/Driver.swift
```diff
@@ -433,6 +435,22 @@ public struct Driver {
 
 }
 
+/// Returns the source files at `location`.
+///
+/// If `location` identifies a file, the result contains that file.
+/// If `location` ientifies a directory, the result contains all source files in that directory.
```
dabrahams: And subdirectories, or not?  Please clarify.

Sources/Driver/Driver.swift
```diff
@@ -433,6 +435,22 @@ public struct Driver {
...
+private func sources(at location: URL) throws -> [SourceFile] {
+  if location.pathExtension == "hylo" {
+    return [try SourceFile(contentsOf: location)]
+  } else {
+    var s: [SourceFile] = []
+    try SourceFile.forEach(in: location) {
```
dabrahams: Every URL indicates a location.  Saying `path` helps compensate for weak type information

```suggestion
/// Returns the source files at `path`.
///
/// If `path` is a file, the result contains that file.
/// If `path` is a directory, the result contains all source files in that directory.
private func sources(at path: URL) throws -> [SourceFile] {
  if path.pathExtension == "hylo" {
    return [try SourceFile(contentsOf: path)]
  } else {
    var s: [SourceFile] = []
    try SourceFile.forEach(in: path) {
```

However `p` would be fine too

Sources/Driver/Driver.swift
tothambrus11: I would call it something like `(shall)LoadStandardLibrary` to clarify that it's imperative rather than describing a current/observed state.

dabrahams: Oh, didn't realize that semantics.  But an imperative `Bool`?  I don't think so.  How about `usesStandardLibrary`?

Sources/Driver/Driver.swift
dabrahams: And the only reason not to use the role for the name is that the role is the “source” of the loading.

Sources/Driver/Driver.swift
RishabhRD: Sadly, it was already present in `Driver`. Will open a PR to fix these.

RishabhRD:
I was holding this PR for #377, but since #397 is handling that now, I don't see a reason to keep this one on hold. @dabrahams have addressed your comments after approval; if it looks good to you, we can merge this work.

tothambrus11 (approved):
LGTM!

PR #272, opened 2026-07-10
Fix record alignment in layout computation

### LINKED ISSUE #268: Incorrect alignment returned in layout algorithm
The alignment of a type must be the largest alignment of all members (assuming alignments are powers of 2, otherwise we need to take smallest common multiple of all)

```suggestion
    let a = if let max = ps.last { rs[max].layout.alignment } else { 1 }
```
`rs` is not reordered, only `ps`.

_Originally posted by @tothambrus11 in https://github.com/hylo-lang/hylo-new/pull/176#discussion_r3487868905_

tothambrus11:
Fixes #268

### DIFF
--- a/Sources/BackEnd/Program+LLVM.swift
+++ b/Sources/BackEnd/Program+LLVM.swift
@@ -1218,7 +1218,7 @@ extension Program {
       size = next + fieldSize
     }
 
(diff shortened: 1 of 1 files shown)


kyouko-taiga (approved):
LGTM, thanks!

PR #277, opened 2026-07-11
Attempt to update CodeCov's server record of our YAML

dabrahams:
The website shows a JSON schema error, and

> This is the default YAML for the current repository, after validation. This YAML takes precedence over the global YAML, but will be overwritten if a YAML change is included in a commit.

So this makes a whitespace change.

### DIFF
--- a/.github/codecov.yml
+++ b/.github/codecov.yml
@@ -1,6 +1,5 @@
 ignore:
   - '!(Sources|StandardLibrary)/'
-
 comment:
   require_changes: true
 


tothambrus11:
I haven't noticed the problem. Where do you see it? We can merge this, though I'm not sure what it will fix.

tothambrus11 (approved):

dabrahams:
I saw it in the web interface to their server

dabrahams:
Huh.  Didn't work.

If you use [this syntax](https://app.codecov.io/github/hylo-lang/hylo-new/config/yaml) for `require_changes` [here](https://app.codecov.io/account/github/hylo-lang/yaml/), it complains and won't let you save.  Probably a benign-ish bug on their side.

tothambrus11:
You could try reporting it here: https://github.com/codecov/feedback/issues/new/choose

PR #278, opened 2026-07-12
Introduce unlink, link, relink abstractions in `List`

tothambrus11:
I extracted this commit out of my old while loop PR. The node relinking abstraction in `List` may be useful in the future.

### DIFF
--- a/Sources/StableCollections/List.swift
+++ b/Sources/StableCollections/List.swift
@@ -399,8 +399,8 @@ public struct List<Element> {
     linkNeighborsToPreallocated(address, before: b)
   }
 
-  /// Moves the bucket at `address` to appear after `a`.
-  ///
+  /// Moves the bucket at `address` to appear after `a`, without invalidating addresses.
+  /// 
   /// - Requires: `address` != `a`.
   public mutating func move(_ address: Address, after a: Address) {
     precondition(isInBounds(address), "address out of bounds")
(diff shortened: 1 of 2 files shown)


kyouko-taiga (commented):
I am not a fan of this API. It is leaking too many implementation details about the list and it is not very clear what "linking" means. If you're just moving things around, then the names of the functions should more closely match the vocabulary of `RangeReplaceableCollection `.

kyouko-taiga:
> may be useful in the future.

Only "may" be? If you don't need new features for work that's already in your pipeline, let's not add dead API endpoints to our data structures "just in case".

tothambrus11:
>  It is leaking too many implementation details about the list and it is not very clear what "linking" means. If you're just moving things around, then the names of the functions should more closely match the vocabulary of RangeReplaceableCollection .

Thanks, good point. I renamed the operations to `move`, which doesn't imply anything about links. I added documentation about address stability, which now became less evident. `RangeReplaceableCollection` didn't have a primitive for moving, just for removal/addition, which we already had, except that couldn't guarantee address stability.

tothambrus11:
I think this change also simplified the existing implementation of `insert` and `remove` by abstracting away `unlink` and `linkNeighborsToPreallocated`. I would argue that `move` is really part of the efficient basis of the `List` data structure, regardless of whether we need it or not. But for now don't waste time on reviewing this, we can integrate it later or if somebody else has time to review it.

tothambrus11:
This is not useful for now. If we need it, we can add it later.

PR #280, opened 2026-07-12
Generate synthetic deinitializers for enums

kyouko-taiga:


Sources/FrontEnd/IR/Instructions/IREnumTag.swift
```diff
@@ -0,0 +1,55 @@
+import Archivist
+
+/// Copies the discriminator of an enum.
+///
+///     enum_tag <source : value>
```
tothambrus11, line 5: I was confused for a moment about seeing `: value` here, because I thought the notation describes the distinction between value / place. Then I saw in the description that it's in fact a place. Could we make a short markdown document describing the conventions to use in IR specification?

Sources/FrontEnd/IR/Instructions/IREnumTag.swift
```diff
@@ -0,0 +1,55 @@
...
+///
+/// `enum_tag` copies the tag identifying the active payload of the initialized enum held in the
+/// place `source`, defining a register with the value of that discrminiator as a `Builtin.word`.
```
tothambrus11, line 8: It's okay to keep it `word` for now, but in the future I would make it define an integer type corresponding to the custom representation of the enum tag, which may be `Int128` as well.

tothambrus11: It may be beneficial to always use as small of a type for the tag as possible, and if the user declares cases with explicit raw values, have those values computed on demand. I think Swift does this for Strings anyway, but we can generalize it further. In most cases, one byte should be enough to store the tag.

Sources/FrontEnd/IR/Instructions/IRPlaceCast.swift
```diff
@@ -4,10 +4,12 @@ import Archivist
 ///
 ///     place_cast <source : value> as <access : effect> <target : type>
 ///
-/// `place_cast` projects the value of `source` as a value of type `target` into a region, using
-/// `effect` to determine the expected memory state of the source before and after that region. The
-/// conversion is not checked. Accessing the resulting place has undefined behavior unless `target`
-/// is layout-compatible with the type of `source`.
+/// `place_cast` projects the contents of `source`, which is the result an access instruction, as
+/// contents of type `target` into a region, using `effect` to determine the memory state of the
+/// source before and after that region. `source` is a place
```
tothambrus11:
```suggestion
/// source before and after that region. `source` is a place.
```

Sources/FrontEnd/IR/Instructions/IRSwitch.swift
```diff
@@ -0,0 +1,80 @@
+import Archivist
+
+/// Transfers control flow to the start one of several basic blocks.
```
tothambrus11:
```suggestion
/// Transfers control flow to the start of one of several basic blocks, according to `scrutinee`.
```

Sources/FrontEnd/IR/IREmitter.swift
```diff
@@ -2078,6 +2063,13 @@ internal struct IREmitter {
     insert(T.End(start: start, anchor: currentAnchor))
   }
 
+  /// Inserts an `enum_tag` instruction.
+  internal mutating func _enum_tag(_ source: IRValue) -> IRValue {
```
tothambrus11, line 2067: Sometimes I see snake case, other times camel case. Should we make these consistent or is there a pattern when to use what?

Sources/FrontEnd/IR/Instructions/IRCase.swift
```diff
@@ -1,6 +1,17 @@
 import Archivist
+import Utilities
 
 /// Accesses to the payload of an enum.
```
tothambrus11:
```suggestion
/// Accesses the payload of an enum.
```

Sources/FrontEnd/IR/Transforms/IRFunction+LifetimeNormalization.swift
```diff
@@ -828,6 +878,22 @@ private struct Transfer: AbstractTransferFunction {
...
+    _ v: IRValue, openedWith k: AccessEffect, closedBy e: AnyInstructionIdentity,
+    in f: inout IRFunction
+  ) {
+    switch k {
+    case .let, .set, .inout:
+      checkInitialized(place: v, in: f, at: f.at(e).anchor.site)
+    case .sink:
+      ensureDeinitialized(place: v, before: e, in: &f)
+    case .auto:
+      fatalError("invalid IR")
```
tothambrus11, line 893: It would be helpful to add a comment or expand the error message to describe why this is a problem here, why is it illegal to have `.auto` in the IR at this place.

Tests/CompilerTests/positive/synthesized-deinit.hylo
tothambrus11, line 6: Could you add a test case for enums?

kyouko-taiga: Enums are already covered by existing use cases.

Sources/FrontEnd/IR/IREmitter.swift
kyouko-taiga: We use snake case to name IR instruction constructors and camel case everywhere else.

Sources/FrontEnd/IR/Instructions/IREnumTag.swift
kyouko-taiga: That is already what we do. We use `Builtin.word` in the abstract semantics of Hylo, which does not have to reflect exactly the types of the concrete LLVM implementation. The distinction helps keeping the IR simpler and more abstract.

I don't see a compelling reason for using something different here. `Builtin.word` is certainly large enough to hold all possible enum cases and that is the type we use for indexing elsewhere, so I think it is a good default. Again, that is orthogonal to the actual representation of the enum in the backend, which may be as small as an empty object for enums with less than two cases.

Tests/CompilerTests/positive/synthesized-deinit.hylo
tothambrus11: I didn't find a test case where there is an empty `fun deinit () {}` for enums.

kyouko-taiga: It is already covered by trivially deinitializable cases.

Sources/FrontEnd/IR/Transforms/IRFunction+LifetimeNormalization.swift
kyouko-taiga: I agree, but I don't think such documentation belongs to this PR. There is a lot of code in the transformation passes already that expect invariants of the sort. What should be done is a proper description of the pre/postconditions of each mandatory pass, which is work that goes beyond the scope of this patch.

Sources/FrontEnd/IR/Instructions/IREnumTag.swift
tothambrus11: I see, nice trick.

Sources/BackEnd/Program+LLVM.swift
```diff
@@ -1159,32 +1214,37 @@ extension Program {
...
+    if payloadOffset > 0 {
+      let padding = payloadOffset - tagSize
+      fields.append(ctx.llvm.arrayType(padding, ctx.llvm.i8).t)
+      fields.swapAt(1, 2)
     }
+
+    let pair = ctx.llvm.structType(named: name, fields, packed: true)
+    let layout = ConcreteLayout(
+      fields: fields, propertyToField: [0, fields.count - 1],
+      size: .fixed(tagSize + payloadSize),
```
tothambrus11: We should add the padding's size if present.

kyouko-taiga: Good catch.

Sources/BackEnd/Program+LLVM.swift
```diff
@@ -1159,32 +1214,37 @@ extension Program {
...
+    let tag = ctx.integerTypeToRepresent(cases.count)
+    let tagSize = ctx.llvm.layout.storageSize(of: tag)
+    var fields: [LLVMType] = [
+      tag.t,
+      ctx.llvm.arrayType(payloadSize, ctx.llvm.i8).t
+    ]
+
+    // Add padding after the tag to satisfy the alignment of the payload if necessary.
+    let payloadOffset = tagSize.rounded(upToNearestMultipleOf: payloadAlignment)
+    if payloadOffset > 0 {
```
```suggestion
    if payloadOffset > tagSize {
```
The current condition is probably always true.

PR #281, opened 2026-07-12
Report hylo errors legibly during testing

dabrahams:
That the errors come out looking like compiler output matters both for humans and for tools.

### DIFF
--- a/Sources/FrontEnd/Diagnostic.swift
+++ b/Sources/FrontEnd/Diagnostic.swift
@@ -1,5 +1,3 @@
-import Foundation
-
 /// A diagnostic related to a region of Hylo source code.
 public struct Diagnostic: Hashable, Sendable {
 
@@ -75,11 +73,7 @@ extension Diagnostic: Comparable {
 extension Diagnostic: CustomStringConvertible {
 
   public var description: String {
-    var r = ""
-    render(
-      into: &r, showingPaths: .relative(to: FileManager.default.currentDirectoryURL),
-      style: .unstyled)
-    return r
+    "\(site.gnuStandardText()): \(level): \(message)"
   }
 
 }
--- a/Sources/FrontEnd/Diagnostic.swift
+++ b/Sources/FrontEnd/Diagnostic.swift
@@ -1,3 +1,5 @@
+import Foundation
+
 /// A diagnostic related to a region of Hylo source code.
 public struct Diagnostic: Hashable, Sendable {
 
@@ -73,7 +75,11 @@ extension Diagnostic: Comparable {
 extension Diagnostic: CustomStringConvertible {
 
   public var description: String {
-    "\(site.gnuStandardText()): \(level): \(message)"
+    var r = ""
+    render(
+      into: &r, showingPaths: .relative(to: FileManager.default.currentDirectoryURL),
+      style: .unstyled)
+    return r
   }
 
 }
--- a/Sources/FrontEnd/DiagnosticSet.swift
+++ b/Sources/FrontEnd/DiagnosticSet.swift
@@ -1,4 +1,5 @@
 import OrderedCollections
+import Foundation
 
 /// A set of diagnostics.
 public struct DiagnosticSet: Hashable, Sendable, CustomStringConvertible {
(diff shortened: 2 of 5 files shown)


dabrahams:
This probably won't be accepted due to  not using the rendering framework and the precedent that CompilerTests renders paths relative to the root of the project.  That will work for some systems; it doesn't work for emacs. To work, a relative path needs to be relative to the directory where the compilation (or test) command was initiated.

tothambrus11:
IMO this is a clear improvement from what we had before. We can merge it and handle relative paths later if it becomes a problem. I think absolute paths are the most reliable to be resolvable by editors.

tothambrus11:
image

This emits relative paths on linux, relative to `CompilerTests`. I confirmed that it gets recognized and is clickable in VSCode, but not in JetBrains IDEs (probably none of us use it anyway). To make it print absolute paths, I would recommend changing `FileName.description`'s implementation to this:
```swift
    url.isFileURL ? url.absoluteURL.path : url.absoluteString
```
Does that solve the issue with emacs?

tothambrus11:
I tested the change, it renders absolute paths like `/home/ambrus/hylo-new3/StandardLibrary/Sources/Core/Numbers/Int.hylo:13.9: error: consecutive statements on the same line must be separated by ';'`, which are clickable in VSCode and IntelliJ.

Note that the test cases still render the relative paths, your changes only impact the standard library's diagnostic printing because it's going through a different path of the driver (loading a module using the driver vs compiling step by step in the test framework). I wouldn't change the test framework to print absolute paths because the printed diagnostics are part of the captured expectations, but it's fine to make this change to the loading diagnostics, as it's a clear improvement from what we had before.

kyouko-taiga:
Where are these absolute paths rendered? I agree with the rationale that disseminating the full path to a file is not necessarily a great idea for security reasons. I also would not want to get absolute paths in error messages when I use my compiler in an interactive terminal session.

tothambrus11:
We could change the relative URL rendering strategy to use the current working directory as the base. We would need to update test cases to include the `Tests/CompilerTests/` prefix in the expectations.

dabrahams:
@tothambrus11:

```
    url.isFileURL ? url.absoluteURL.path : url.absoluteString
```

is broken because `.path` doesn't get you a path in the filesystem in the syntax recognized by the system.  On Windows it's got forward slashes and the drive letter cannot be represented correctly.  This is why I hate to use `URL` as a generalized filesystem path.

The only correct formula for that is [here](https://github.com/dabrahams/SPMBuildToolSupport/blob/598aa38867b6ed40ba9f256959684b81c1adf766/SPMBuildToolSupport.swift#L133-L136).

dabrahams:
I'm not attached to absolute paths.  Relative to the directory where `swift test` is invoked would be fine.

dabrahams:
Changed to relative paths and uses the render framework

Sources/FrontEnd/Diagnostic.swift
```diff
@@ -73,7 +75,11 @@ extension Diagnostic: Comparable {
 extension Diagnostic: CustomStringConvertible {
 
   public var description: String {
-    "\(site.gnuStandardText()): \(level): \(message)"
+    var r = ""
+    render(
+      into: &r, showingPaths: .relative(to: FileManager.default.currentDirectoryURL),
+      style: .unstyled)
```
tothambrus11: In the future we can detect whether the user is running the program from a terminal and change styling accordingly. See [isatty](https://github.com/dtolnay/isatty).

Sources/FrontEnd/DiagnosticSet.swift
```diff
@@ -33,4 +33,8 @@ public struct DiagnosticSet: Hashable, Sendable {
     for d in ds { insert(d) }
   }
 
+  public var description: String {
+    "\n" + elements.lazy.map(\.description).joined(separator: "\n")
```
tothambrus11:
```suggestion
    elements.lazy.map({ "\n\($0)" }).joined()
```
We could avoid the special casing for the first line. This avoids an extra newline when the collection is empty.

dabrahams: good point!  But I had to drop that code.

PR #283, opened 2026-07-13
Driver: loading stdlib should enable standard library dependency

RishabhRD:
Posted by @dabrahams as [comment](https://github.com/hylo-lang/hylo-new/pull/270#discussion_r3561096491)
> I find it weird that we have to both load the standard library and tell the driver that the program depends on it. One of those should be sufficient.

### DIFF
--- a/Sources/Driver/Driver.swift
+++ b/Sources/Driver/Driver.swift
@@ -278,7 +278,7 @@ public struct Driver {
     // Compile the module from sources.
     let m = program.demandModule(module)
 
-    if addStandardLibraryDependency {
+    if addStandardLibraryDependency && module != Module.standardLibraryName {
       program[m].addDependency(Module.standardLibraryName)
     }
 
--- a/Sources/Driver/Driver.swift
+++ b/Sources/Driver/Driver.swift
@@ -42,9 +42,6 @@ public struct Driver {
   /// `true` iff the standard library and its shim should be excluded from compilation and linking.
   public var noStandardLibrary: Bool = false
 
-  /// `true` iff a standard library dependency should be added to loaded modules.
-  private var addStandardLibraryDependency: Bool = false
-
   /// The program being compiled by the driver.
   public var program: Program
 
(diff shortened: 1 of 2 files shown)


Sources/Driver/Driver.swift
```diff
@@ -275,6 +278,10 @@ public struct Driver {
     // Compile the module from sources.
     let m = program.demandModule(module)
 
+    if addStandardLibraryDependency {
+      program[m].addDependency(Module.standardLibraryName)
```
tothambrus11, line 279: You may need to guard for checking if the current module being loaded is the standard library itself.

Sources/Driver/Driver.swift
```diff
@@ -42,6 +42,9 @@ public struct Driver {
   /// `true` iff the standard library and its shim should be excluded from compilation and linking.
   public var noStandardLibrary: Bool = false
 
+  /// `true` iff a standard library dependency should be added to loaded modules.
+  private var addStandardLibraryDependency: Bool = false
```
tothambrus11: It's a bit unfortunate that the Driver exposes both the high-level load function and the lower-level `assignScopes`, `lower`, etc. The variable only has effect on the former. Maybe we should have a driver-driver, or GPS, something higher-level. It is still useful to expose the lower-level primitives for the LSP, which needs to drive the pipeline slightly differently. 

Maybe the docs could be a bit clearer if we replace `to loaded modules` with `when loading modules with load(_:withSourcesAt:)`.

Apart from this, LGTM!

kyouko-taiga: What's the purpose of this property? Why it isn't simply `!noStandardLibrary`?

kyouko-taiga (changes requested):
Thanks for taking a jab at this. I think we don't need the additional property.

Sources/Driver/Driver.swift
RishabhRD: > Maybe we should have a driver-driver, or GPS, something higher-level.

I think this is out  of scope of this PR, we can track it in some other PR though.

tothambrus11 (approved):

RishabhRD:
Removed the additional property that solves @kyouko-taiga comment. I think we can merge this one.

PR #284, opened 2026-07-13
Implement built-in support for converting pointers to places and vice versa

kyouko-taiga:
~The test `pointer-place-roundtrip.hylo` is only a temporary measure. It is most likely UB.~

### DIFF
--- a/Sources/FrontEnd/IR/Instructions/IRPointerToPlace.swift
+++ b/Sources/FrontEnd/IR/Instructions/IRPointerToPlace.swift
@@ -1,17 +1,14 @@
 import Archivist
 
-/// Projects a value at a different type.
+/// Converts a built-in pointer to a place.
 ///
-///     place_cast <source : value> as <access : effect> <target : type>
+///     pointer_to_place <source : value> as <access : effect> <target : type>
 ///
-/// `place_cast` projects the contents of `source`, which is the result an access instruction, as
-/// contents of type `target` into a region, using `access` to determine the memory state of the
-/// source before and after that region.
-///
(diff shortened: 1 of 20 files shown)


Sources/FrontEnd/IR/Instructions/IRPointerToPlace.swift
```diff
@@ -0,0 +1,65 @@
+import Archivist
+
+/// Converts a built-in pointer to a place.
+///
+///     pointer_to_place <source : value> as <access : effect> <target : type>
+///
+/// `pointer_to_place` defines a place of type `target` with the value of `source`, which denotes
+/// the value of a `Builtin.ptr`, using `access` to determine the memory state of the place. The
+/// resulting place unsafely refers to the memory referenced by `source`.
```
tothambrus11, line 9: Does this allow taking a pointer-to-tuple and case it to its first element? (assuming fixed, paddingless memory layout)

Sources/FrontEnd/IR/Transforms/IRFunction+DeadCodeElimination.swift
tothambrus11:
```suggestion
  /// wrapped into a type application so that it matches the expected type. This method can
```

Sources/FrontEnd/IR/Instructions/IRPointerToPlace.swift
kyouko-taiga: Yes, that would be an unsafe reinterpret cast.

```hylo
let xs = (1, 2)
let p = Builtin.address(of: xs)
let x = p as* (let Int)
print(x)
use(xs)
```

PR #296, opened 2026-07-15
Add Hylo-demangle tool

lucteo:
The tool parses the input and replaces the symbols that look like Hylo mangled symbols with their demangled string, keeping everything else the same.

Add "$h" as a prefix for Hylo mangled names.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -19,6 +19,7 @@ let package = Package(
   ],
   products: [
     .executable(name: "hc", targets: ["hc"]),
+    .executable(name: "hylo-demangle", targets: ["hylo-demangle"]),
     .library(name: "HyloStandardLibrary", targets: ["StandardLibrary"]),
     .library(name: "HyloFrontEnd", targets: ["FrontEnd"]),
   ],
@@ -56,6 +57,14 @@ let package = Package(
       name: "hc-tests",
       dependencies: [
         .product(name: "ArgumentParser", package: "swift-argument-parser")
+      ],
+      swiftSettings: commonSwiftSettings),
+
+    .executableTarget(
+      name: "hylo-demangle",
+      dependencies: [
+        .target(name: "FrontEnd"),
+        .product(name: "ArgumentParser", package: "swift-argument-parser"),
       ],
       swiftSettings: commonSwiftSettings),
 
--- a/Sources/FrontEnd/IR/Mangling/DemangledSymbol.swift
+++ b/Sources/FrontEnd/IR/Mangling/DemangledSymbol.swift
@@ -3,7 +3,8 @@ internal indirect enum DemangledSymbol: Hashable, Sendable {
 
   /// Creates an instance decoding the symbol mangled in `s`.
   ///
-  /// `self` is assigned to an error if `s` is malformed.
+  /// `self` is assigned to `.notMangled` if `s` doesn't start with the mangling prefix, or to
+  /// `.error` if `s` is malformed.
   internal init(_ s: String) {
     guard s.starts(with: ManglingEncoding.manglingPrefix) else {
       self = .notMangled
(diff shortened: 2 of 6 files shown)


Sources/hylo-demangle/CommandLine.swift
```diff
@@ -0,0 +1,101 @@
...
+  /// Executes the command.
+  public func run() throws {
+    let input = try sourceText()
+
+    if listOnly {
+      input.enumerateTokens { (token) in
+        if let demangled = String(token).demangled() {
+          print("\(token) => \(demangled)")
+        }
+      }
```
tothambrus11: This could be factored out to a function, then we can avoid the early return and just have a symmetrical-looking `if`/`else`.

tothambrus11 (changes requested):
Nice work! 
If we are fine with keeping the whole processed text in memory, we could refactor the printing to accumulate into a string, so that the code becomes easier to test (the corner cases of the index handling and making sure nothing from the content gets dropped).

Tests/BackEndTests/SimpleFunctionEmitterTest.swift
```diff
@@ -31,7 +31,7 @@ final class SimpleFunctionEmitterTest: XCTestCase {
     XCTAssert(
       llvmIR.containsOnce(
         """
-        define private void @"M6testvUpvirtualu6rky9fkv9x3F06mainlT01tR50tR5$gP71L1L1L"() {
```
tothambrus11, line 34: It's really nice that most of the mangling tests could remain the same, while we still got this indicative change here :D

Package.swift
```diff
@@ -59,6 +60,14 @@ let package = Package(
       ],
       swiftSettings: commonSwiftSettings),
 
+    .executableTarget(
+      name: "hylo-demangle",
```
tothambrus11, line 64: The current pattern seems to be to prefix CLI tools with `hc-` (e.g., `hc-tests`).

Sources/FrontEnd/IR/Mangling/DemangledSymbol.swift
```diff
@@ -5,7 +5,13 @@ internal indirect enum DemangledSymbol: Hashable, Sendable {
   ///
   /// `self` is assigned to an error if `s` is malformed.
```
tothambrus11: I'm wondering if we should describe `.notMangled` here.

Package.swift
lucteo: I tried to keep the name we had in the old compiler. I feel that using `hc-` prefix is not thee most appropriate, as this is not a direct part of the compiler

tothambrus11: I think I also like the `hylo-` prefix more, it's also less obfuscated.

Sources/hylo-demangle/CommandLine.swift
```diff
@@ -0,0 +1,105 @@
...
+        print(token, terminator: "")
+      }
+      lastIndex = token.endIndex
+    }
+    if lastIndex < input.endIndex {
+      print(input[lastIndex..<input.endIndex], terminator: "")
+    }
+  }
+
+  /// Prints the demangled form of each matched mangled symbol om `input`.
```
tothambrus11:
```suggestion
  /// Prints the demangled form of each matched mangled symbol in `input`.
```

Sources/hylo-demangle/CommandLine.swift
```diff
@@ -0,0 +1,105 @@
...
+        print(demangled, terminator: "")
+      } else {
+        print(token, terminator: "")
+      }
+      lastIndex = token.endIndex
+    }
+    if lastIndex < input.endIndex {
+      print(input[lastIndex..<input.endIndex], terminator: "")
+    }
+  }
```
tothambrus11, line 68: It would be nice to have some tests for this code, especially its edge cases. We could accumulate into a string and print only at the call site.

tothambrus11: We decided that it wasn't worth the refactor for now.

tothambrus11 (approved):

PR #303, opened 2026-07-16
Implement basic cross-module inlining

kyouko-taiga:
This patch implements basic cross-module inlining for functions marked with `@inline(always)`, which mandates inlining.

Inlining is not yet guaranteed to apply recursively on all `@inline(always)` functions as this patch does not implement any call-graph analysis. However, the change is sufficient to support the compilation of user-programs using basic functions from the standard library, without compiling the standard library into its own object file.

See `factorial.hylo` for a demonstration.

### DIFF
--- a/Sources/BackEnd/Program+LLVM.swift
+++ b/Sources/BackEnd/Program+LLVM.swift
@@ -256,7 +256,7 @@ extension Program {
     if case .function(let n, _) = s.callee {
       let f = demandFunction(named: n, in: &ctx.module)
       let x = insertArguments(s.arguments, mappedWith: f.prototype.mapping, in: &ctx)
-      insertCall(applying: f, to: x, writingResultTo: s.result, in: &ctx)
+      insertCall(f, with: x, writingResultTo: s.result, in: &ctx)
     }
 
     // Is the call indirect?
@@ -265,7 +265,7 @@ extension Program {
       let a = types.seenAsTermAbstraction(ctx.ir.result(of: s.callee)!.type)!
       let p = prototype(types[a], in: &ctx.module)
       let x = insertArguments(s.arguments, mappedWith: p.mapping, in: &ctx)
-      insertCall(applying: f, describedBy: p, to: x, writingResultTo: s.result, in: &ctx)
+      insertCall(f, describedBy: p, with: x, writingResultTo: s.result, in: &ctx)
     }
 
     return ctx.ir.instruction(after: i.erased)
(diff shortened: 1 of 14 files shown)


Sources/BackEnd/Program+LLVM.swift
```diff
@@ -956,6 +977,19 @@ extension Program {
...
+  /// `xs` is an array of IR values denoting places of type `t`, which are defined in the current
+  /// context. This method inserts a load for each of these values in the compiled function.
+  private mutating func insertLoad(
+    _ xs: [FrontEnd.IRValue], of t: MachineType.ID, in ctx: inout FunctionGenerationContext
+  ) -> [LLVMValue] {
+    let m = metadata(of: t, in: &ctx.module)
+    let lhs = ctx.module.llvm.insertLoad(m.llvm, from: ctx.value[xs[0]]!, at: ctx.insertionPoint!)
+    let rhs = ctx.module.llvm.insertLoad(m.llvm, from: ctx.value[xs[1]]!, at: ctx.insertionPoint!)
+    return [lhs.v, rhs.v]
+  }
```
tothambrus11, line 991: It would be more suitable to ask for a tuple here, or generalize this with `xs.map { ... }`

Sources/FrontEnd/Syntax/Builtins/BuiltinFunction.swift
```diff
@@ -768,8 +768,8 @@ extension BuiltinFunction: Showable {
     //      return (p != .ignore) ? "shl_\(p)_\(t)" : "shl_\(t)"
     //    case .udiv(let e, let t):
     //      return e ? "udiv_exact_\(t)" : "udiv_\(t)"
-    //    case .sdiv(let e, let t):
-    //      return e ? "sdiv_exact_\(t)" : "sdiv_\(t)"
+    case .sdiv(let e, let t):
+      return printer.format(e ? "sdiv_exact_\(t)" : "sdiv_\(t)", [t.erased])
```
tothambrus11:
```suggestion
      return printer.format(e ? "sdiv_exact_%T" : "sdiv_%T", [t.erased])
```

StandardLibrary/Sources/Core/Numbers/Int32.hylo
```diff
@@ -13,6 +13,94 @@ public struct Int32 is Deinitializable {
...
+      return .new(value: result)
+    }
+  }
+
+  /// Returns the result of `self` divided by `other`.
+  ///
+  /// - Requires: `other` is different from zero.
+  @inline(always)
+  public fun infix/ (other: Self) -> Self {
+    if other == (0 as Self) {
```
tothambrus11, line 60:
```suggestion
    if other == .zero() || ((-2147483648 as Int32) == self && other == (-1 as Self)) {
```
division of Int32.min / -1 causes overflow, as it would be represented as Int32.max + 1

tothambrus11: I'll apply this in the integer type generator.

PR #304, opened 2026-07-16
Fix formatting injection

### LINKED ISSUE #297: Invalid placeholder
I get an inscrutable fatal error when I test the `invalid-placeholder-bug` branch with `swift test --filter testStandardLibraryLoading`

tothambrus11:
Fixes #297 

The problem was that some string was interpolated into the formatting string, which ended up becoming parsed by `Program.format` and failed. We should never use raw string interpolation except if we are sure the injected string won't contain `%`.

### DIFF
--- a/Sources/FrontEnd/Diagnostic.swift
+++ b/Sources/FrontEnd/Diagnostic.swift
@@ -123,7 +123,7 @@ extension Program {
     _ f: AnyTypeIdentity, mutably isAppliedMutably: Bool, at site: SourceSpan
   ) -> Diagnostic {
     let x = isAppliedMutably ? "mutably" : "immutably"
-    let m = format("cannot call bundle of type '%T' \(x)", [f])
+    let m = format("cannot call bundle of type '%T' %S", [f, x])
     return .init(.error, m, at: site)
   }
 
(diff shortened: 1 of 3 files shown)


kyouko-taiga (approved):
LGTM

dabrahams:
> We should never use raw string interpolation except if we are sure the injected string won't contain `%`.  

Seems to me we should never use `%` formatting and always use string interpolation.  It shocks me more than a little to see that we have implemented our own system for this.

tothambrus11:
The motivation is that to print each individual component, the `Showable`  instances must have access to `Program`, they can't print themselves otherwise. We would need to write this otherwise: `"x: \(program.show(x)); y: \(program.show(y)"`, which adds some noise to what we want to express. Do you have any idea that could help?

dabrahams:
```swift
func <T: Showable> show(_ x: T) { program.show(x) }

"x: \(show(x)); y: \(show(y))"
```

```swift
func <T: Showable> s(_ x: T) { program.show(x) }

"x: \(s(x)); y: \(s(y))"
```

```swift
prefix operator %
prefix func <T: Showable>%(_ x: T) { program.show(x) }

"x: \(%x); y: \(%y)"
```

kyouko-taiga:
```
func show<T: Showable>(_ x: T) { program.show(x) }

"x: \(show(x)); y: \(show(y))"
```

Where does `program` come from? Are you suggesting we write this function in all types that have a `program` property? That will not work as well as you think at call sites because we re-project all sorts of `self` in closures.

> It shocks me more than a little to see that we have implemented our own system for this.

We need the program and we need a mutable instance of printer, which persists information through the visit of the things being printed.

I think you can trust that I would not have invented a whole formatting system just because I'm not aware of or don't like `\()`.

PR #305, opened 2026-07-16
Clarify Metatype layout documentation.

dabrahams:
Otherwise, this comment says that the buffer of pointers contains those things.  The first commit clarifies with just a comma.  The second commit goes a bit further.

### DIFF
--- a/Sources/BackEnd/ModuleGenerationContext.swift
+++ b/Sources/BackEnd/ModuleGenerationContext.swift
@@ -61,7 +61,8 @@ internal struct ModuleGenerationContext: ~Copyable {
 
   /// The header of the structure representing information stored in all type witnesses.
   ///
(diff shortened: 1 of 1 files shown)


Sources/BackEnd/ModuleGenerationContext.swift
```diff
@@ -61,7 +61,8 @@ internal struct ModuleGenerationContext: ~Copyable {
 
   /// The header of the structure representing information stored in all type witnesses.
   ///
-  /// A type witness is a 4-tuple optionally followed by a buffer of pointers containing:
+  /// A type witness is a 4-tuple followed by a tail buffer of zero or more pointers. 
+  /// The tuple contains:
   /// - A string describing the type being witnessed;
```
dabrahams, line 66: This should really be clearer about what kind of string representation is used.  E.g. nul-terminated utf8.

tothambrus11 (approved):

PR #314, opened 2026-07-17
Fix bugs in SortedSet

tothambrus11:
- `union` was O(n^2) instead of O(n+m)
- `update` had its nil-returning semantics opposite to what the [protocol required](https://developer.apple.com/documentation/swift/setalgebra/update(with:)).
- `union` and `symmetricDifference` didn't include the rest of the elements when one side was emptied

### DIFF
--- a/Sources/Utilities/SortedSet.swift
+++ b/Sources/Utilities/SortedSet.swift
@@ -161,10 +161,10 @@ extension SortedSet: SetAlgebra {
   public mutating func update(with e: consuming Element) -> Element? {
     let (inserted, k) = index(insertingIfAbsent: e)
     if inserted {
+      return nil
+    } else {
       swap(&e, &elements[k])
       return e
-    } else {
-      return nil
     }
   }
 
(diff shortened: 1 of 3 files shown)


Sources/Utilities/SortedSet.swift
```diff
@@ -177,29 +177,45 @@ extension SortedSet: SetAlgebra {
...
-
-    // Insert into the largest container to minimize the number of lookups.
-    if lhs.count < rhs.count { swap(&lhs, &rhs) }
+    if self.isEmpty {
+      return other
+    } else if other.isEmpty {
+      return self
+    }
 
+    var result: ContiguousArray<Element> = []
```
RishabhRD, line 190: Nice optimization. IDK how worthy it would be to reserve capacity of result to avoid allocation while insertion. It would atleast be `max(self.elements.count, other.elements.count)`.

Tests/UtilitiesTests/SortedSetTests.swift
```diff
@@ -143,4 +143,226 @@ final class SortedSetTests: XCTestCase {
...
+    func hash(into hasher: inout Hasher) {
+      hasher.combine(key)
+    }
+
+  }
+
+  /// Returns lists of pseudo-random sets to combine, covering a range of lengths, sizes, and
+  /// overlaps, drawn from `g`.
+  private func randomSetLists(
+    using g: inout SeedableRandomNumberGenerator
```
RishabhRD: These private functions can be private extension of `SeedableRandomNumberGenerator`. That would help the readability.

RishabhRD (approved):
LGTM! couple of comments.

Sources/Utilities/SortedSet.swift
tothambrus11: That's a good idea! Initially I wanted to reserve `n + m` but decided not to in order to not over-reserve extra memory. This is clearly better!

PR #317, opened 2026-07-18
Add hc --version option

### LINKED ISSUE #316: Add `--version` option and fill it in the release workflow
~~During development, this should be set as a commit hash, or `<some-hash>-DIRTY` indicating if the working tree is modified. The version file can be generated by a build plugin, taking the version from an environment variable.~~

A simpler solution is to just have a `Version.swift` file in the driver folder and have it replaced by the CI when needed. It doesn't provide a tag in local builds but I think that's not worth the extra complexity for now.

floze-the-genius:
## Summary

Closes #316.

Adds a built-in `hc --version` option and embeds the release tag into every
published distributable.

## Changes

- adds `Sources/hc/Version.swift` with `development` as the local-build value
- wires that value into `ArgumentParser`'s command configuration
- replaces the source value from `github.ref_name` before release builds
- verifies the unpacked distributable reports the expected tag-derived version
  on every release matrix target

The release workflow strips only the conventional leading `v`, so a tag such
as `v0.2.0-test-lsp` produces `hc --version` output of `0.2.0-test-lsp`.

## Testing

- parsed `.github/workflows/release.yml` with PyYAML
- built `hc` locally with Swift 6.3 and LLVM 20.1.8
- verified the default binary reports `development`
- temporarily embedded `1.2.3-test`, rebuilt, and verified the binary reports
  exactly `1.2.3-test`
- ran `git diff --check`

`swift test` reaches test-target compilation but cannot import `XCTest` because
this machine has Apple Command Line Tools rather than `Xcode.app`. The `hc`
product itself builds successfully; the repository CI runs the full test suite
with its configured Swift and LLVM environment.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -73,7 +73,7 @@ jobs:
       - name: Embed release version
         run: |
           $version = $env:RELEASE_TAG -replace '^v', ''
-          "let hyloVersion = `"$version`"" | Set-Content Sources/hc/Version.swift
+          "internal let hyloVersion = `"$version`"" | Set-Content Sources/hc/Version.swift
         shell: pwsh
 
       - name: Download Recent Visual Studio and Windows SDK
@@ -147,7 +147,6 @@ jobs:
 
           $binaryName = if ($env:RUNNER_OS -eq 'Windows') { 'hc.exe' } else { 'hc' }
           $binary = Join-Path $extractDir $binaryName
-          & $binary --help
           $expectedVersion = $env:RELEASE_TAG -replace '^v', ''
           $actualVersion = (& $binary --version).Trim()
           if ($actualVersion -ne $expectedVersion) {
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -48,6 +48,7 @@ jobs:
     runs-on: ${{ matrix.os }}
     env: &distributable-env
       ARTIFACT_NAME: hylo-${{ github.ref_name }}-${{ matrix.target }}
+      RELEASE_TAG: ${{ github.ref_name }}
 
     steps:
       - uses: actions/checkout@v7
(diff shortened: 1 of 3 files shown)


Sources/hc/CommandLine.swift
```diff
@@ -12,7 +12,9 @@ private typealias Module = FrontEnd.Module
 @main struct CommandLine: AsyncParsableCommand {
 
   /// Configuration for this command.
-  public static let configuration = CommandConfiguration(commandName: "hc")
+  public static let configuration = CommandConfiguration(
+    commandName: "hc",
+    version: hyloVersion)
```
```suggestion
  public static let configuration = CommandConfiguration(commandName: "hc", version: hyloVersion)
```

It's still below the 100 character line length limit, so we don't need to break.

.github/workflows/release.yml
```diff
@@ -141,6 +148,11 @@ jobs:
           $binaryName = if ($env:RUNNER_OS -eq 'Windows') { 'hc.exe' } else { 'hc' }
           $binary = Join-Path $extractDir $binaryName
           & $binary --help
```
tothambrus11:
```suggestion
```
Now that we check the version, no need to check `--help`. That used to be a workaround for not having a `--version`.

Sources/hc/Version.swift
```diff
@@ -0,0 +1 @@
+let hyloVersion = "development"
```
tothambrus11:
```suggestion
internal let hyloVersion = "development"
```
Even though `internal` is the default, we use the convention to add access modifiers explicitly unless something is already constrained to be `private`, e.g. a method within a `private struct`.

tothambrus11 (changes requested):
Hi, thanks a lot for the contribution! The changes look good, there are just some project-specific nitpicks that I wanted to point out. Our [coding style guide](https://github.com/hylo-lang/hylo-new/pull/102/changes) is not yet officially merged in, sorry for the inconvenience.

I saw you had some problems with the development environment setup, do you need any help? Feel free to reach out on Slack.

I would like to kindly ask you to not use AI for writing the PR description / comments, as it's usually much longer and verbose than it needs to be, and it's better to know we are talking to a human. For the future, code contributions should be also written by humans to avoid lack of ownership, guarantee code quality and for ethical reasons. IMO AI tends to abstract away the pain and the learning of insights that we need when developing a new language.

tothambrus11 (approved):
LGTM!

PR #318, opened 2026-07-18
Interpreter type layout cache

RishabhRD:
The PR doesn't cover the `Arrow` case. Will cover it at time of function call implementation as that requires little bit of exploration of `FunctionPointer` and `Arrow` type.


Sources/Interpreter/TypeLayoutCache.swift
```diff
@@ -0,0 +1,148 @@
+import FrontEnd
+import Utilities
+
+/// A memoizing computer of type layouts.
+struct TypeLayoutCache {
+
+  /// The ABI for which the types will be laid out.
+  let abi: any TargetABI
```
tothambrus11, line 8: Do we ever want to switch the `abi` at runtime to a different type of TargetABI? If not, we could add this as a type parameter to `TypeLayoutCache` and store a concrete TargetABI type here, so we can avoid existentials. The struct can be still instantiated during testing with a different type of TargetABI.

RishabhRD: We would never switch the ABI at runtime. That said, this code originally came from @dabrahams, so I'm curious whether this was an intentional design choice or simply a convenience. I'm happy with either approach.

dabrahams: I don't remember, sorry.  Do whatever makes sense!

Sources/Interpreter/TypeLayoutCache.swift
```diff
@@ -0,0 +1,256 @@
...
+      in: &p)
+  }
+
+  /// Returns the layout for a record `t` in `p` having members `ms`.
+  private mutating func computeLayout(
+    record t: MonomorphicTypeIdentity,
+    havingMembers ms: [(type: MonomorphicTypeIdentity, label: String?)],
+    in p: inout Program
+  ) -> TypeLayout {
+    var b = TypeLayout.Bytes(alignment: 1, size: 0)
```
dabrahams: I think this might not be consistent with the layout algorithm in Hylo-new, which is described [here](https://github.com/hylo-lang/hylo-new/blob/1d9afc037064e6a2a5136da7dd84e981d09c74b1/Tests/RuntimeTests/RuntimeTests.swift#L227-L241).  (There are tests in that file too, which you could crib)

Sources/Interpreter/TypeLayoutCache.swift
```diff
@@ -0,0 +1,256 @@
...
+
+  /// Returns the layout for a tagged enum `t` in `p`.
+  private mutating func computeLayout(
+    taggedEnum t: MonomorphicTypeIdentity,
+    in p: inout Program
+  ) -> TypeLayout {
+    let basis = storage(nominal: t.underlying, in: &p)
+      .map { layout(MonomorphicTypeIdentity($0), in: &p) }
+
+    if basis.count == 0 {
```
dabrahams: This should treat the enum as a record type containing the discriminator and an undiscriminated union, and should use the record layout algorithm to determine the ordering of those two parts.

Sources/Interpreter/TypeLayoutCache.swift
```diff
@@ -0,0 +1,256 @@
...
+      let a = type(t, as: TypeApplication.self, in: p)
+      let v = tag(a.abstraction, in: p)
+      return v == Enum.self
+    } else {
+      return false
+    }
+  }
+
+  /// Returns true iff enum `t` in `p` has a representation.
+  private func isTaggedEnum(_ t: AnyTypeIdentity, in p: Program) -> Bool {
```
dabrahams: I don't understand the name.  What's a "tagged" enum?

Sources/Utilities/UnsignedFixedWidthInteger+Extensions.swift
```diff
@@ -0,0 +1,22 @@
+extension FixedWidthInteger where Self: UnsignedInteger {
+
+  /// The number of bytes needed to hold `self` bits.
+  public var bitsToBytes: UInt {
+    (UInt(self) + 7) / 8
+  }
+
+  /// The number of bits in `self` bytes.
+  public var bytesToBits: UInt {
+    UInt(self) * 8
```
dabrahams: Something about making this a property of `FixedWidthInteger` rubs me the wrong way.  Maybe it's the fact that bytes and bits are just units, and we probably don't want to clutter all integers with conversions between every kind of unit.  Don't know what to suggest though.

dabrahams (commented):
Left some comments

Sources/Interpreter/TypeLayoutCache.swift
tothambrus11: You may have meant tagged/discriminated union. That's the default in Hylo when we talk about enums, so it could be sufficient to say `isEnum`. We can use "untagged union" if we ever end up adding that feature.

Sources/Interpreter/TypeLayoutCache.swift
```diff
@@ -0,0 +1,256 @@
...
+    } else if u == TypeApplication.self {
+      let a = type(t, as: TypeApplication.self, in: p)
+      let v = tag(a.abstraction, in: p)
+      return v == Enum.self
+    } else {
+      return false
+    }
+  }
+
+  /// Returns true iff enum `t` in `p` has a representation.
```
tothambrus11: Is it significant that it has _a_ representation or whether it has _enum_ representation? I would update either the docs or the name to clarify this.

Sources/Interpreter/TypeLayoutCache.swift
RishabhRD: Done.

Sources/Interpreter/TypeLayoutCache.swift
RishabhRD: It was there to differentiate between raw value enum (that is currently not completely implemented in compiler) and normal enum that I called tagged enum. However, I got the way to get rid of this "tagged" enum name.

Sources/Utilities/UnsignedFixedWidthInteger+Extensions.swift
RishabhRD: For now, I inlined `bytesToBits` on use-site as there is only one. `bitsToBytes` was just there for completeness and thus now removed the both.

Sources/Interpreter/TypeLayoutCache.swift
RishabhRD: It is significant as having representation implies raw value enum like thing from swift (that is not fully implemented right now in compiler). However, updated the docs to directly talk about raw value enum as readers might not care about implementation detail as you suggested.

Sources/Interpreter/TargetABI.swift
```diff
@@ -53,10 +53,8 @@ struct UnrealABI: TargetABI {
 extension TargetABI {
 
   /// Returns a discriminator type for the enum of `n` types in `p`.
-  ///
-  /// Precondition: `n` is positive
   func enumDiscriminator(count n: Int, in p: inout Program) -> MonomorphicTypeIdentity {
-    if n == 1 { return MonomorphicTypeIdentity(.void) }
+    if n <= 1 { return MonomorphicTypeIdentity(.void) }
```
dabrahams, line 60: I'd have left the comparison alone and added a precondition check.  That there's at least one is implied by the fact that it's an enum, IIUC

RishabhRD: > That there's at least one is implied by the fact that it's an enum, IIUC

But empty enum compiles for me:
```hylo
enum EmptyEnum {}
public fun main() {}
```
Or, am I misunderstanding your statement?

Although, I should add a precondition check for `n >= 0` for sure.

dabrahams: Ok yes I wasn't thinking straight

Sources/Interpreter/MonomorphicTypeIdentity.swift
```diff
@@ -0,0 +1,19 @@
...
+
+  /// The underlying type identity having no generic parameters or
+  /// unification variables.
+  public let underlying: AnyTypeIdentity
+
+  /// Creates an instance wrapping `t` as `MonomorphicTypeIdentity`.
+  ///
+  /// - Precondition: `t` is monomorphic.
+  public init(_ t: AnyTypeIdentity) {
+    precondition(!t[.hasGenericParameter] && !t[.hasVariable])
```
dabrahams, line 15: @kyouko-taiga what kind of type has a variable but no generic parameter?  There must be some gap in my understanding.

Sources/Interpreter/TargetABI.swift
```diff
@@ -0,0 +1,67 @@
...
+  /// An instance.
+  public init() {}
+
+  /// The size of a word in `bits`
+  private let bitsInAWord = 64
+
+  /// The maximal alignment of a builtin type in bytes.
+  private let maxAlignment = 128 / 8
+
+  /// Returns the layout for a `bitWidth`-bit type.
```
dabrahams:
```suggestion
  /// Returns the layout for a `bitWidth`-bit builtin type.
```

Sources/Interpreter/TargetABI.swift
```diff
@@ -0,0 +1,67 @@
...
+      bitWidth > 0 && bitWidth.nonzeroBitCount == 1,
+      "bit width \(bitWidth) is not a power of 2.")
+    let sizeInBytes = (bitWidth + 7) / 8
+    return .init(
+      alignment: min(sizeInBytes, maxAlignment),
+      size: sizeInBytes)
+  }
+
+  /// Returns the layout of `t`.
+  public func layout(_ t: MachineType) -> TypeLayout.Bytes {
```
dabrahams, line 37: What's the difference between a builtin type and a `MachineType`?

Sources/Interpreter/TypeLayout.swift
```diff
@@ -0,0 +1,146 @@
...
+
+    /// The type of the part.
+    public let type: MonomorphicTypeIdentity
+
+    /// The byte offset of the part with respect to the layout.
+    public let offset: Int
+  }
+
+  /// Aggregate layout values of this layout.
+  public let bytes: Bytes
```
dabrahams:
```suggestion
  /// Aggregate properties of this layout.
  public let whole: Bytes
```

Sources/Interpreter/TypeLayout.swift
```diff
@@ -0,0 +1,146 @@
...
+  }
+
+  /// True iff `self` is the layout of an enum type, which changes how
+  /// its `parts` are interpreted.
+  public let isEnumLayout: Bool
+}
+
+extension TypeLayout {
+
+  /// The discriminator of an enum layout.,
```
dabrahams:
```suggestion
  /// The discriminator of an enum layout.
```

Sources/Interpreter/TypeLayout.swift
```diff
@@ -0,0 +1,146 @@
...
+
+extension TypeLayout.Bytes {
+
+  /// Returns the layout of the tuple `(S, T)`, where `S` and `T` are types whose layout is
+  /// represented by `self` and `t` respectively.
+  ///
+  /// - Note: the `T` instance is stored `t.size` bytes before the end of the tuple.
+  func appending(_ t: Self) -> Self {
+    let r = self.size.rounded(upToNearestMultipleOf: t.alignment)
+    return .init(alignment: max(self.alignment, t.alignment), size: r + t.size)
```
dabrahams: This `max` call may have come from code I wrote.  In any case I think it presumes power-of-2 alignment, which is not (no longer) valid for the NFE.

Sources/Interpreter/TypeLayoutCache.swift
```diff
@@ -0,0 +1,255 @@
...
+  /// An instance for laying out types according to `abi`.
+  public init(for abi: any TargetABI) {
+    self.abi = abi
+  }
+
+  /// The layout for `t` in `p`.
+  public mutating func layout(
+    _ t: MonomorphicTypeIdentity,
+    in p: inout Program
+  ) -> TypeLayout {
```
dabrahams:
```suggestion
  /// Returns the layout of `t`, defined in `p`.
  public mutating func layout(
    _ t: MonomorphicTypeIdentity, in p: inout Program) -> TypeLayout
{
```

Is the comment change correct?  The fact that I have to ask implies there's some ambiguity in the original comment. Always a good idea to add some information that isn't purely derivable from the signature. Apply also elsewhere.

Sources/Interpreter/TypeLayoutCache.swift
```diff
@@ -0,0 +1,255 @@
...
+    if let r = storage[t] { return r }
+    let r = computeLayout(t, in: &p)
+    storage[t] = r
+    return r
+  }
+
+  /// Returns the layout for `t` in `p`.
+  ///
+  /// - Precondition: `t` is not an alias.
```
dabrahams: Wouldn't this precondition also apply to the function above?

Sources/Interpreter/TypeLayoutCache.swift
```diff
@@ -0,0 +1,255 @@
...
+    } else if hasRecordLayout(t.underlying, in: p) {
+      return computeLayout(record: t, in: &p)
+    } else if hasEnumLayout(t.underlying, in: p) {
+      return computeLayout(enum: t, in: &p)
+    } else {
+      unreachable("\(p.show(t.underlying)) doesn't have any layout)")
+    }
+  }
+
+  /// Returns the layout for record `t` in `p`.
```
dabrahams:
```suggestion
  /// Returns the layout of the record type `t` in `p`.
```

Can't leave out “type” in this case because saying something is a record means something different from saying that it's a record type.  Adding the type information from the signature to the wording makes it seem ambiguous.  Apply also elsewhere.

dabrahams: Consider whether the label should change to `recordType`

Sources/Interpreter/TypeLayoutCache.swift
```diff
@@ -0,0 +1,255 @@
...
+    return TypeLayout(
+      bytes: discriminatorLayout.bytes,
+      type: t,
+      parts: [.init(name: "discriminator", type: discriminator, offset: 0)],
+      isEnumLayout: true
+    )
+  }
+
+  /// Returns true iff `t` in `p` has a record layout.
+  private func hasRecordLayout(_ t: AnyTypeIdentity, in p: Program) -> Bool {
```
dabrahams: is this not `isRecordType`?
I don't see what it has to do with layout

Sources/Interpreter/TypeLayoutCache.swift
```diff
@@ -0,0 +1,255 @@
...
+      let d = type(t, as: Enum.self, in: p).declaration
+      return p[d].representation != nil
+    } else {
+      let a = type(t, as: TypeApplication.self, in: p)
+      let d = type(a.abstraction, as: Enum.self, in: p).declaration
+      return p[d].representation != nil
+    }
+  }
+
+  /// Returns the types of stored parts of record `t` in `p`.
```
dabrahams:
```suggestion
  /// Returns the types of `t`'s stored parts, where `t` is defined in `p`.
```

Sources/Interpreter/TypeLayoutCache.swift
```diff
@@ -0,0 +1,255 @@
...
+    p.types.tag(of: t).value
+  }
+}
+
+/// Returns the storage layout and part offsets of a record having members
+/// with layouts `ms` in declaration order.
+func storageLayoutOfRecord(
+  havingMembers ms: [TypeLayout.Bytes]
+) -> (bytes: TypeLayout.Bytes, partOffsets: [Int]) {
+  let storageOrder = ms.enumerated().sorted {
```
dabrahams, line 212: The builtin `sort` is stable, so I think you can simplify this code

Sources/Interpreter/TypeLayoutCache.swift
```diff
@@ -0,0 +1,255 @@
...
+      return computeLayout(rawValueEnum: t, in: &p)
+    }
+    let cases = storage(nominal: t.underlying, in: &p).map { c in
+      layout(.init(c), in: &p)
+    }
+
+    let d = layout(abi.enumDiscriminator(count: cases.count, in: &p), in: &p)
+
+    let payload = TypeLayout.Bytes(
+      alignment: cases.map(\.alignment).max() ?? 1,
```
dabrahams: I think you need the LCM of alignments.  Please check my reasoning though.

Sources/Interpreter/TypeLayoutCache.swift
```diff
@@ -0,0 +1,255 @@
...
+
+  /// Returns the layout for `t` in `p`.
+  ///
+  /// - Precondition: `t` is not an alias.
+  private mutating func computeLayout(
+    _ t: MonomorphicTypeIdentity,
+    in p: inout Program
+  ) -> TypeLayout {
+    if t.underlying[.hasAliases] {
+      let l = layout(.init(p.types.dealiased(t.underlying)), in: &p)
```
dabrahams: If you have this code, why the precondition?

dabrahams (approved):
A few critical things, mostly minor ones in comments.

Sources/Interpreter/TypeLayout.swift
```diff
@@ -0,0 +1,146 @@
...
+    isEnumLayout ? 2 : parts.count
+  }
+}
+
+extension BinaryInteger {
+
+  /// Returns `self` rounded up to the nearest multiple of `n`.
+  ///
+  /// - Precondition: `n > 0`.
+  fileprivate func rounded(upToNearestMultipleOf n: Self) -> Self {
```
dabrahams: I realize this is quite trivial, but I think the following is actually more consistent with [this coding guideline](https://www.swift.org/documentation/api-design-guidelines/#give-prepositional-phrase-argument-label).

```suggestion
  fileprivate func roundedUp(toNearestMultipleOf n: Self) -> Self {
```

tothambrus11: Maybe `nearest` could be dropped, it's quite evident that we wouldn't round to a non-nearest multiple. We could keep the clarification in the documentation though.

tothambrus11: It's not envident whether this rounds away from zero or strictly up. Should we clarify this in the docs?

dabrahams: It says "rounded up."  How much clearer could it be?  I don't think we need to add redundant information for people who might choose to question the words that are plainly written there.  Is there' some reason you think people will ignore “up?”

Furthermore, the precondition says that `n` is positive so “away from zero” and “up” are synonymous.  What am I missing?

dabrahams: I agree that “nearest” could be dropped, if you know what rounding is :-)

```suggestion
fileprivate func roundedUp(toAMultipleOf n: Self) -> Self {
```

tothambrus11: Fair enough, I was too paranoid. 

>  Is there' some reason you think people will ignore “up?”

:DD

> precondition says that n is positive

Yes, but `self` may be negative.

Sources/Interpreter/TargetABI.swift
RishabhRD: `BuiltinType` has been renamed as `MachineType` in NFE. Will use machine type in docs too for consistency.

Sources/Interpreter/TypeLayoutCache.swift
RishabhRD: Yes. Got rid of special handling of type aliases, so that solves this now.

Sources/Interpreter/MonomorphicTypeIdentity.swift
tothambrus11: Type variables are used for constraint solving. They don't necessarily correspond to generic parameters. E.g. a function call adds a new type variable for the `callee` and a constraint that that type must be callable with the given argument list. During constraint solving, `TypeVariable` types get gradually substituted with concrete types (which may in fact be the type of a generic parameter). Requiring that `t` is not a variable here means that constraint solving managed to successfully substitute all variables within the type. That's always true for types that the typer presents on a successfully type-checked program: if we can't discharge a type variable, that usually means ambiguity or the lack of enough constraints.

Sources/Interpreter/MonomorphicTypeIdentity.swift
```diff
@@ -0,0 +1,19 @@
...
+/// The identity of a monomorphic type.
+struct MonomorphicTypeIdentity: Regular {
+
+  /// The underlying type identity having no generic parameters or
+  /// unification variables.
+  public let underlying: AnyTypeIdentity
+
+  /// Creates an instance wrapping `t` as `MonomorphicTypeIdentity`.
+  ///
+  /// - Precondition: `t` is monomorphic.
```
tothambrus11:
```suggestion
  /// - Precondition: `t` is monomorphic and not a type variable.
```

Sources/Interpreter/TargetABI.swift
tothambrus11: `MachineType` represents all the builtin types. There are also some ways to refer to builtin entities like Never, Void, MetaType, that aren't types themeselves, but have a type assigned by the compiler. See https://github.com/hylo-lang/hylo-new/pull/416/changes#diff-aacc3e39c69eb662478d1907efe75a477e52677af9e9b6515d8802a32ba9c174 for some more context.

RishabhRD:
All comments have been resolved, thus merging the PR.

Sources/Interpreter/TypeLayout.swift
dabrahams: Fair enough; @RishabhRD please investigate. Take care to avoid a complicated semantic description; it would be better to add preconditions or extend only on unsigned types.

Issue #112, opened 2026-04-25
Optimize standard library cache loading in tests


tothambrus11:
Currently, we deserialize the standard library for each test case, which takes up significant time. We should read the standard library cache and deserialize it only once for the whole test suite.

Issue #80, opened 2026-04-07
Reuse standard library module cache instead of reading it from disk


tothambrus11:
Profiling suggested that currently loading libraries from cache takes significant time in test cases. We could try loading the cache once, and subsequently just passing copies of the loaded program to the test cases.

kyouko-taiga:
See #106 for another optimization.

tothambrus11:
Marking this as blocked by LLVM lowering, as that's highest priority for now, and these might have conflicting changes.

Issue #288, opened 2026-07-14
Build the standard library module as part of `swift build`


dabrahams:
The standard library should be built by `swift build`.  We want the build step to detect all compile-time errors.

Ideally, the release build would incorporate the standard library module into the released executable.
Because that involves building the compiler without the standard library, then compiling the standard library, it may mean another link step to incorporate the library into the resulting executable package.  If so, we would want to avoid doing that in debug builds.

tothambrus11:
Gating the build of the compiler on whether the standard library can be compiled seems a bit too strict for me during regular development. In cases when we temporarily make a source-breaking change to the compiler, it's good to be able to write focused, independent tests that run without the standard library. Once we confirmed our change works in isolation, we can try it on the standard library and on the rest of the test cases.

Currently if you want to type-check the standard library for development purposes, I would recommend running one of the test cases that loads the standard library: https://github.com/hylo-lang/hylo-new/blob/main/Tests/CompilerTests/StandardLibraryLoadingTests.swift We can make some top-level bash script for making that convenient. I can also try to get the LSP working for standard library development, currently it's not prepared for that.

Before releases, we already run the tests, so a broken standard library shouldn't get through. We could bundle the prebuilt standard library, but I think it's cheaper to compile it on demand and cache it. Otherwise, we would need a version of the standard library compiled for every platform, which increases our archive size a lot. As a compromise, we can bundle the standard library for the host platform and do lazy compilation for everything else.

dabrahams:
Then there should be a target you can build that gets you the compiler without the standard library.  That's probably needed anyway.  The question is does `swift build` with no arguments succeed or fail if the library is broken.
The standard library is a part of what we are shipping.  If `swift build` isn't going to be the command that creates the artifact we ship, what command will do that?

tothambrus11:
I'm not sure I understand your argument yet. Before shipping something, one is expected to build AND run all tests. What do we want to optimize for here? The suggestions you proposed so far appear to me as they would make our build system system more complex and with some additional effort needed to opt out of the library check on builds - another flag to worry about.

A bit more context on the releasing: Unfortunately, SPM is currently doesn't provide enough support to prepare all of what we need for a distributable. We need to bundle dynamic library dependencies, and in the future probably do some more sophisticated post-processing to generate the proper archive for package managers with the correct layout. The current bundles are (almost) self-contained portable bundles distributed as a tar.

[...]

PR #336, opened 2026-07-21
Enable caching of standard library program in parallel test runs

### LINKED ISSUE #330: Investigate the reason for lack of parallelism in test suite
Currently, running tests in parallel don't seem to be significantly improve performance. I would expect a 10-20x speedup, but it's more like 10%-200% depending on how the muons are hitting my CPU. Maybe there is some shared resource that's causing the contention?

tothambrus11:
XCTest creates a new process per test case in parallel test runs, which made every test case compile the standard library from scratch. This PR migrates the CompilerTests to use Swift Testing, which uses a lightweight, task based scheduling, so we can share the standard library's Program between every test case.

Debug builds before:
image

After:
image

Fixes #330, https://github.com/hylo-lang/hylo-new/issues/112 and https://github.com/hylo-lang/hylo-new/issues/80


Tests/CompilerTests/positive/generic-call-1.llvm-ir.expected
```diff
@@ -0,0 +1,29 @@
+define i32 @"$hM6TestUggenericcall1cDpTcTK1C3P1gcTK20tR50$fJ5J"(ptr %0, ptr %1) {
```
tothambrus11, line 1: These files were forgotten to be renamed at some point, so they got outdated. Refreshing them now.

tothambrus11:
@kyouko-taiga Are you fine with moving the tests (at least the CompilerTests) to Swift testing?

I remember you had some opposing arguments against it a while ago, but I don't remember these exactly. My reluctance came from bad CMake support, but that's irrelevant now that we dropped support for that.

So far my experience was pretty good. It has great performance benefits in parallel test runs. I'm fine with keeping the regular tests in XCTest, but I'd also be happy to migrate everything to Swift testing for consistency. From the fraction of code I migrated, it seems that it makes tests more readable.

tothambrus11:
Debug in parallel runs:

before:
image

after:

when stdlib is a cache miss:
image


when stdlib is a cache hit:

image

Tests/CompilerTests/positive/generic-call-1.llvm-ir.expected
dabrahams: How could that happen?  Do we need to update the testing infrastructure to catch these cases?

dabrahams:
I defer entirely to @kyouko-taiga  on this one.

Tests/CompilerTests/positive/generic-call-1.llvm-ir.expected
tothambrus11: Yes, I think that would be great. At least during PRs.

tothambrus11:
@kyouko-taiga Are you fine with moving this part of the testing suite to use Swift testing?

Tests/CompilerTests/positive/generic-call-1.llvm-ir.expected
dabrahams: Would you please file an issue? I don't understand it well enough I think. Thank you!

tothambrus11: https://github.com/hylo-lang/hylo-new/issues/436

Sources/Driver/Driver.swift
```diff
@@ -204,24 +198,22 @@ public struct Driver {
     from module: Module.ID,
     withCSources cSources: [URL] = [],
     writingTo output: URL
-  ) throws -> PhaseResult {
+  ) async throws -> PhaseResult {
+    // FIXME: Enable this after we can lower the standard library
+    // modulesToLink.append(program.modules[.standardLibrary]!.identity)
+    let shimObject = noStandardLibrary
```
kyouko-taiga: How is that a shim?

Sources/Driver/Driver.swift
```diff
@@ -204,24 +198,22 @@ public struct Driver {
...
-        cSources.append(chosenStandardLibraryRoot.appending(component: cShimSource))
-      }
-
       try FileManager.default.withUniqueTemporaryDirectory { (d) in
         let hyloObjects = try writeObjectFiles(for: modulesToLink, into: d)
-        let cObjects = try cSources.map { (s) in
+        var cObjects = try cSources.map { (s) in
           try compileCToObject(source: s, destinationDirectory: d)
         }
+        if let shimObject { cObjects.append(shimObject) }
```
```suggestion
        if let o = shimObject { cObjects.append(o) }
```
or
```suggestion
        cObjects.append(contentsOf: shimObject)
```

Sources/Driver/Driver.swift
```diff
@@ -434,6 +443,37 @@ public struct Driver {
...
+
+    /// Returns an object file compiled from the standard library's C shim with `relocation`,
+    /// compiling it at most once per process into a temporary directory that lives until the
+    /// process exits.
+    func object(compiledWith relocation: RelocationModel) throws -> URL {
+      if let o = objects[relocation] { return o }
+
+      let d = try FileManager.default.createUniqueTemporaryDirectory()
+      let s = Driver.chosenStandardLibraryRoot.appending(component: cShimSource)
+      let o = d.appendingPathComponent("shims.o", isDirectory: false)
```
kyouko-taiga: This name seems likely to show up in user projects. Is that an issue?

Tests/CompilerTests/CompilerTests.swift
```diff
@@ -222,80 +225,110 @@ final class CompilerTests: XCTestCase {
 
   }
 
-  /// A temporary folder for caching compilation artifacts during testing.
+
+  /// The record of any test failures during a CompilerTest.
```
kyouko-taiga, line 233:
```suggestion
  /// A record of any test failures during a CompilerTest.
```
Each instance can be a different record, right?

kyouko-taiga (approved):
I am fine with these changes but I'd like to test the UX over some period of time to see if they do not disrupt my dev routine too much, so please wait before merging. I'll do a proper test over the next weekend.

In the meantime I left some comments.

Sources/Driver/Driver.swift
tothambrus11: I don't want to require global uniqueness of object file names. CMake and linkers at least work with absolute paths.

Tests/CompilerTests/CompilerTests.swift
tothambrus11: It's intended to be used once per test case, so that this accumulates the test results over multiple assertions.

Sources/Driver/Driver.swift
tothambrus11: See the line below. It's the object file compiled from `shims.c` unless `noStandardLibrary` is true.

Sources/Driver/Driver.swift
tothambrus11: It's pretty neat that Optional conforms to Sequence/Collection.

Sources/Driver/Driver.swift
tothambrus11: Would it help readability to use an if expression here?

dabrahams:
This may be a good change in general but it may be the wrong way to solve the problem.  Building the standard library can be a build step so that it's available by the time testing of the standard library begins… if you don't want it to be a part of a complete build it can be a dependency target of the tests that use the standard library.  And finally, I don't understand why running separate processes should be an obstacle for caching,  since the library can be written to a file on disk.

Sources/Driver/Driver.swift
kyouko-taiga: Ah, there is a line break! Yeah, maybe the `if` expression would help.

tothambrus11:
> And finally, I don't understand why running separate processes should be an obstacle for caching, since the library can be written to a file on disk.

The goal is partially also to avoid re-reading the archive from disk, currently that would also take significant time per test case. The in-process caching is one aspect of what made this much faster. The other reason is that we already had a standard library cache, but it was removed at the end of each process, which meant many test cases had to start over without a cache. That would be in fact solved if a build plugin produced the standard library cache. The in-process cache reuse would have to be added over either solutions.

> if you don't want it to be a part of a complete build it can be a dependency target of the tests that use the standard library.

I think that would work. If the build system is guaranteed to be maximally lazy, a broken standard library compilation would only break the tests that require a complete standard library.

A concern may be that we have to create two test targets, and supposedly a third common shared test library target for the common test utilities.

If I understand correctly, the goal behind having the build system generate the standard library is a) reduce build system emulating complexity on our side; b) avoid running test cases that rely on the standard library if the standard library fails compilation.

I'm afraid `a)` wouldn't be met because there is also a significant complexity in instrumenting SPM to track the files properly, and the build would involve 2-3 extra targets. I actually think `b)` is not yet practical for us, but will be in the near future. SPM build plugin crashes and failures [used to be hidden and very hard to diagnose](https://github.com/swiftlang/swift-package-manager/issues/10042), but the fix for this just got in yesterday. Once it gets into a Swift release, I think this becomes a viable option.

Let me know if I didn't capture all aspects accurately. (For reference, our previous discussion: https://github.com/hylo-lang/hylo-new/issues/288)

kyouko-taiga:
I don't have a strong opinion about the setup but I'd just like to point out that I am not convinced by the necessity to have a "minimal" standard library. Is there a chance we could simplify?

If caching becomes good enough I guess it doesn't matter whether a test uses the minimal or the full version. And as we progress I expect more and more tests will require the latter anyway.

tothambrus11:
A minimal standard library would be faster in terms of name resolution, but I think we shouldn't solve our problems  by adding a minimal standard library, but by a) using builtins and `no-std` where possible; b) speeding up the compiler (possibly with more caches) so we don't have to worry about this. Either way, I agree that it's not worth adding support for minimal standard library at the moment.

There are some parts in the compiler that trigger a crash is a std symbol is not resolved. Maybe we should make those optional, then we could enable no-std in more places.

Sources/Driver/Driver.swift
```diff
@@ -65,10 +65,10 @@ public struct Driver {
 
   #if USE_BUNDLED_STANDARD_LIBRARY // Set compiler flag in distributable builds.
   /// The root folder of the standard library's sources.
-  private let chosenStandardLibraryRoot = bundledStandardLibrarySources
+  private static let chosenStandardLibraryRoot = bundledStandardLibrarySources
```
dabrahams: I know it's not part of this change, but why is “chosen” a part of this name?  The doc comment doesn't explain its presence or say anything that would leave me to believe it can't be dropped! /cc @kyouko-taiga

tothambrus11: Yeah, good point. I have already changed the name of this to `standardLibraryRoot` in another PR.

tothambrus11:
@kyouko-taiga did you get a chance to test this workflow? I think now you shouldn't need to erase the build folder for the tests to show up thanks to the correction in our build dependencies for the test generator.

PR #345, opened 2026-07-22
Avoid creating fresh type variables when lowering built-in applications

kyouko-taiga:
This patch improves the performance of the emitter by avoiding creating new fresh type variables before lowering `apply_builtin` instructions.

Before the patch, we called `BuiltinFunction.type(uniquingTypesWith:)` in `IREmitter._emitApply(builtin:to:)` in order to get the access effect of each parameter. This call caused the creation of new fresh type variables for the generic built-in functions (e.g., `Builtin.address(of:)`), eventually requiring a reallocation of the type store. This patch fixed that issue by reusing types inferred during typing and by constructing universal types for generic built-in functions, so that the result of `BuiltinFunction.type(uniquingTypesWith:)` is idempotent.

### DIFF
--- a/Sources/FrontEnd/IR/IREmitter.swift
+++ b/Sources/FrontEnd/IR/IREmitter.swift
@@ -738,7 +738,7 @@ internal struct IREmitter {
       } else {
         let t = program.type(assignedTo: program[e].callee, assuming: Arrow.self)
         lowering(e) { (me) in
-          let x0 = me._emitApply(builtin: f, typed: t, to: me.program[e].arguments)
+          let x0 = me._emitApply(builtin: f, ofType: t, to: me.program[e].arguments)
           me._emitInitialize(target, with: x0)
         }
       }
@@ -2639,18 +2639,20 @@ internal struct IREmitter {
     _ = _apply_builtin(.trap, ofType: u, to: [])
   }
 
-  /// Generates IR for applying `callee` to `arguments`.
+  /// Calls `f(arguments...)` where `f` is a builtin function of type `t`
+  /// other than `assume_[un]initialized`.
   ///
-  /// `callee` is any built-in function but `assume_[un]initialized`.
+  /// - Note: Calls to `Builtin.assume_[un]initialized` are lowered directly to `assume_state`
+  ///   instructions rather than function applications.
   private mutating func _emitApply(
-    builtin callee: BuiltinFunction, typed calleeType: Arrow.ID,
+    builtin f: BuiltinFunction, ofType t: Arrow.ID,
     to arguments: [LabeledExpression],
   ) -> IRValue {
-    let xs = zip(program.types[calleeType].inputs, arguments).map { (p, a) in
+    let xs = zip(program.types[t].inputs, arguments).map { (p, a) in
       let x0 = lowered(lvalue: a.value)
       return _access([p.access], from: x0)
     }
-    return _apply_builtin(callee, ofType: calleeType, to: xs)
+    return _apply_builtin(f, ofType: t, to: xs)
   }
 
   /// Generates IR for defining a place projecting `source` as a place of type `target` with
--- a/Sources/FrontEnd/IR/IREmitter.swift
+++ b/Sources/FrontEnd/IR/IREmitter.swift
@@ -736,8 +736,9 @@ internal struct IREmitter {
           me._assume_state(target, initialized: true)
         }
       } else {
+        let t = program.type(assignedTo: program[e].callee, assuming: Arrow.self)
         lowering(e) { (me) in
-          let x0 = me._emitApply(builtin: f, to: me.program[e].arguments)
+          let x0 = me._emitApply(builtin: f, typed: t, to: me.program[e].arguments)
           me._emitInitialize(target, with: x0)
         }
       }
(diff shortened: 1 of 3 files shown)


tothambrus11:
Single thread test suite times in debug mode:
Before:
image
After:
image


In release:
Before:
image

After: 
image

Sources/FrontEnd/Syntax/Builtins/BuiltinFunction.swift
```diff
@@ -397,37 +397,42 @@ public enum BuiltinFunction: Hashable, Sendable {
...
     case .addressOf:
-      let t0 = s.fresh().erased
+      let t0 = s.demand(GenericParameter.nth(0, .proper))
       let t1 = s.demand(MachineType.ptr).erased
-      return s.demand(Arrow(inputs: [.init(label: "of", access: .let, type: t0)], output: t1))
+      let a = Arrow(inputs: [.init(label: "of", access: .let, type: t0.erased)], output: t1)
+
+      let t2 = s.demand(a).erased
+      let t3 = s.demand(UniversalType(parameters: [t0], head: t2))
+      return t3.erased
```
tothambrus11, line 415: Previously this was an Arrow with a type variable as its parameter, and now it's a UniversalType with an Arrow head. Could you explain what's the difference? How does a UniversalType gets called?

Sources/FrontEnd/IR/IREmitter.swift
```diff
@@ -2072,12 +2073,12 @@ internal struct IREmitter {
 
   /// Inserts a `apply_builtin` instruction.
   internal mutating func _apply_builtin(
-    _ callee: BuiltinFunction, typed f: Arrow.ID, to arguments: [IRValue]
+    _ callee: BuiltinFunction, typed calleeType: Arrow.ID, to arguments: [IRValue]
```
dabrahams:
```suggestion
  /// Calls `f(arguments...)`, where `f` has type `t`.
  internal mutating func _apply_builtin(
    _ f: BuiltinFunction, ofType t: Arrow.ID, to arguments: [IRValue]
```

Sources/FrontEnd/IR/IREmitter.swift
```diff
@@ -2633,22 +2634,23 @@ internal struct IREmitter {
...
+    _ = _apply_builtin(.trap, typed: u, to: [])
   }
 
   /// Generates IR for applying `callee` to `arguments`.
   ///
   /// `callee` is any built-in function but `assume_[un]initialized`.
   private mutating func _emitApply(
-    builtin callee: BuiltinFunction, to arguments: [LabeledExpression],
+    builtin callee: BuiltinFunction, typed calleeType: Arrow.ID,
+    to arguments: [LabeledExpression],
```
dabrahams:
```suggestion
  /// Calls `f(arguments...)` where `f` is a builtin function of type `t`
  /// other than `assume_[un]initialized`.
  ///
  /// - Note: <describe how you call the assume functions>
  private mutating func _emitApply(
    builtin f: BuiltinFunction, ofType t: Arrow.ID,
    to arguments: [LabeledExpression],
```

Sources/FrontEnd/Syntax/Builtins/BuiltinFunction.swift
dabrahams: good questions

Sources/FrontEnd/Syntax/Builtins/BuiltinFunction.swift
```diff
@@ -437,31 +442,31 @@ extension BuiltinFunction {
     //    case .srem(let t):
     //      return .init(^t, ^t, to: ^t)
     case .and(let t):
-      return s.demand(Arrow(t, t, to: t))
+      return s.demand(Arrow(t, t, to: t)).erased
```
dabrahams, line 445: should we have `s.demandAny`?

dabrahams (approved):
LGTM

Sources/FrontEnd/Syntax/Builtins/BuiltinFunction.swift
```diff
@@ -397,37 +397,42 @@ public enum BuiltinFunction: Hashable, Sendable {
...
       let t1 = s.demand(MachineType.ptr).erased
-      return s.demand(Arrow(inputs: [.init(label: "of", access: .let, type: t0)], output: t1))
+      let a = Arrow(inputs: [.init(label: "of", access: .let, type: t0.erased)], output: t1)
+
+      let t2 = s.demand(a).erased
+      let t3 = s.demand(UniversalType(parameters: [t0], head: t2))
+      return t3.erased
 
     case .assumeInitialized:
       let t0 = s.fresh().erased
```
tothambrus11, line 418: Should we change this to a UniversalType too?

Sources/FrontEnd/Syntax/Builtins/BuiltinFunction.swift
kyouko-taiga: It was always an arrow. A universal type is just a type abstraction, introducing a type parameter in the body of the abstraction.

A type variable can be bound or free. The former are represented by an instance of `GenericParameter`, the latter by `TypeVariable`. So before we returned an arrow with free type variables. Now we return a type in which all type variables are bound (aka a closed type).

It is generally a bad idea to keep open types around. Previously we could get away with that because `type(uniquingTypesWith:)` was only called during type inference, by functions like `Typer.inferredType(of:in:)`. These functions have a typing context, which (implicitly) keeps track of all free variables. But then we also called `type(uniquingTypesWith:)` during lowering, outside of a typing context, thus creating an unbounded number of new types. The PR fixes this specific issue.

A universal type never gets "called", or at least not like I think you are thinking. A universal type introduces a type parameter which can be eliminated with a type application. It's the typer's job to figure out when that should happen. The whole process is rather sophisticated but in a nutshell, when the typer sees a term `e` of type `<T> X` in a position where some other type `Y` is expected, it will infer a type argument and elaborate a term `e<Z>` of type `Y`.

Sources/FrontEnd/Syntax/Builtins/BuiltinFunction.swift
kyouko-taiga: Yes, we should. Good catch.

Sources/FrontEnd/Syntax/Builtins/BuiltinFunction.swift
kyouko-taiga: Maybe indeed. It exists by the way, but it is private.

PR #354, opened 2026-07-22
Print nicer error message when a runtime witness table couldn't be lowered

tothambrus11:
Masking the hard-to-diagnose LLVM assertion failure with something easier to understand while working with the compiler.

### DIFF
--- a/Sources/BackEnd/Program+LLVM.swift
+++ b/Sources/BackEnd/Program+LLVM.swift
@@ -662,6 +662,16 @@ extension Program {
     }
 
     let tableType = metadata(of: s.witnessType, in: &ctx.module)
+
+    assert(entries.allSatisfy(\.unsafe[].isConstant), """
+      Runtime-defined IRWitnessTable operands are not yet supported.
+      See https://github.com/hylo-lang/hylo-new/issues/342
+
+      Problem occurred during LLVM lowering of:
+      \(show(ctx.ir))
+
+      """)
+
     let table = ctx.module.llvm.structConstant(
       of: StructType.UnsafeReference(tableType.llvm)!, aggregating: entries)
 


Sources/BackEnd/Program+LLVM.swift
```diff
@@ -662,6 +662,16 @@ extension Program {
     }
 
     let tableType = metadata(of: s.witnessType, in: &ctx.module)
+
+    assert(entries.allSatisfy(\.unsafe[].isConstant), """
```
kyouko-taiga, line 666: There's an `unimplemented` function for that. We should use it.

tothambrus11: I initially wrote this using an if + unimplemented, but I couldn't use the `inout` parameter `ctx` in the never-returning path  (maybe a compiler bug, or just inconvenience), but I just used assert for simplicity.

tothambrus11: image

tothambrus11: Changed to this:
```swift
    if !entries.allSatisfy(\.unsafe[].isConstant) {
      let ir = show(ctx.ir)
      unimplemented("""
        Runtime-defined IRWitnessTable operand. https://github.com/hylo-lang/hylo-new/issues/342

        In:
        \(ir)

        """)
    }
```

kyouko-taiga: Swift and its half-broken access checker ...

Then we should create an `unimplemented(if:_:)` function (out of scope for this path, of course). I'd like to avoid having 15 different ways of formatting debug locations in `assert`, `fatalError`, etc.

tothambrus11: Good idea, I'll create `unimplemented(if:_:)`. Soon I'll also finish my PR for the crash handler, which will also eliminate the need for manually printing the IR here.

Issue #331, opened 2026-07-20
Add Builtin instruction/constant for Int's and UInt's min/max


tothambrus11:
Currently, the standard library integer type generator assumes Int/UInt is 64-bit. We should add an intrinsic that lets the backend define their min/max values.
See `GenerateIntegeTypes.swift` -> `minimumValue`

We would need to add some Builtin intrinsics: `Builtin.max_s_word()`, `Builtin.min_s_word()`, `Builtin.max_u_word()`, `Builtin.min_u_word()` for signed/unsigned types correspondinly. This will need to touch all layers of the compiler, so it's a very nice first issue. You can follow the `sdiv` builtin throughout the compiler and use it as an example to implement the behavior on all layers.

Please don't use AI for implementing this, it's a learning opportunity for a new contributor.

PR #356, opened 2026-07-23
Report out-of-bounds integer literals

### LINKED ISSUE #335: Report errors for out of bounds integer literals
Currently, bounds of integer literals are never checked, so we can trigger overflow with `let x: UInt8 = 300`.

ColumbusLabs:
## Summary
- validate integer literals against the exact signed or unsigned range of their standard-library target type during lowering
- report an error at the literal while continuing lowering so multiple invalid literals are diagnosed in one pass
- cover the issue reproduction (`let x: UInt8 = 300`), every standard integer type, and accepted min/max boundaries

`Int` and `UInt` use 64-bit bounds to match the standard library generator’s current `Builtin.word` model and its existing issue #331 TODO.

## Verification
- focused positive/negative compiler tests: 2/2
- full `swift test` with the repository-pinned LLVM 20.1.6 toolchain: 437/437
- max Swift line length: 100 characters
- `git diff --check`

Closes #335

### DIFF
--- a/Sources/FrontEnd/IR/IREmitter.swift
+++ b/Sources/FrontEnd/IR/IREmitter.swift
@@ -1446,20 +1446,47 @@ internal struct IREmitter {
   ) -> IRValue {
     let value = BigInt(hyloLiteral: program[source].value)!
 
+    let specification: (width: Int, isSigned: Bool, storage: MachineType, name: String)
     switch target {
-    case program.standardLibraryType(.int), program.standardLibraryType(.uint):
-      return .integer(value, program.types.demand(MachineType.word))
-    case program.standardLibraryType(.int8), program.standardLibraryType(.uint8):
-      return .integer(value, program.types.demand(MachineType.i(8)))
-    case program.standardLibraryType(.int16), program.standardLibraryType(.uint16):
-      return .integer(value, program.types.demand(MachineType.i(16)))
-    case program.standardLibraryType(.int32), program.standardLibraryType(.uint32):
-      return .integer(value, program.types.demand(MachineType.i(32)))
-    case program.standardLibraryType(.int64), program.standardLibraryType(.uint64):
-      return .integer(value, program.types.demand(MachineType.i(64)))
+    // Keep word-sized integer bounds in sync with the generated standard library.
+    case program.standardLibraryType(.int):
+      specification = (64, true, .word, "Int")
+    case program.standardLibraryType(.uint):
+      specification = (64, false, .word, "UInt")
+    case program.standardLibraryType(.int8):
+      specification = (8, true, .i(8), "Int8")
+    case program.standardLibraryType(.uint8):
+      specification = (8, false, .i(8), "UInt8")
+    case program.standardLibraryType(.int16):
+      specification = (16, true, .i(16), "Int16")
+    case program.standardLibraryType(.uint16):
+      specification = (16, false, .i(16), "UInt16")
+    case program.standardLibraryType(.int32):
+      specification = (32, true, .i(32), "Int32")
+    case program.standardLibraryType(.uint32):
+      specification = (32, false, .i(32), "UInt32")
+    case program.standardLibraryType(.int64):
+      specification = (64, true, .i(64), "Int64")
+    case program.standardLibraryType(.uint64):
+      specification = (64, false, .i(64), "UInt64")
     default:
       program.unexpected(target)
     }
+
+    let unit = BigInt(1)
+    let minimum = specification.isSigned ? -(unit << (specification.width - 1)) : 0
+    let maximum =
+      specification.isSigned
+      ? (unit << (specification.width - 1)) - 1
+      : (unit << specification.width) - 1
+    if !(minimum...maximum).contains(value) {
+      let message =
+        "integer literal '\(program[source].value)' is out of bounds for "
+        + "'\(specification.name)'"
+      report(.error, message, about: source)
+    }
+
+    return .integer(value, program.types.demand(specification.storage))
   }
 
   /// Returns the value denoted by `source` interpreted as the floating point
--- a/Sources/FrontEnd/IR/IREmitter.swift
+++ b/Sources/FrontEnd/IR/IREmitter.swift
@@ -1446,47 +1446,20 @@ internal struct IREmitter {
   ) -> IRValue {
     let value = BigInt(hyloLiteral: program[source].value)!
 
-    let specification: (width: Int, isSigned: Bool, storage: MachineType, name: String)
     switch target {
-    // Keep word-sized integer bounds in sync with the generated standard library.
-    case program.standardLibraryType(.int):
-      specification = (64, true, .word, "Int")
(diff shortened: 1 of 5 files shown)


Tests/CompilerTests/negative/integer-literal-bounds.hylo
```diff
@@ -0,0 +1,28 @@
+//! stage:lowering
+
+fun integer_literal_bounds() {
+  let a1: Int8 = -129
+  let a2: Int8 = 128
+  let a3: UInt8 = -1
+  let a4: UInt8 = 300
```
tothambrus11:
```suggestion
  let a4: UInt8 = 256
```
It's good to ensure that the boundary is at exactly the right place.

Sources/FrontEnd/IR/IREmitter.swift
```diff
@@ -1446,20 +1446,47 @@ internal struct IREmitter {
...
-      return .integer(value, program.types.demand(MachineType.word))
-    case program.standardLibraryType(.int8), program.standardLibraryType(.uint8):
-      return .integer(value, program.types.demand(MachineType.i(8)))
-    case program.standardLibraryType(.int16), program.standardLibraryType(.uint16):
-      return .integer(value, program.types.demand(MachineType.i(16)))
-    case program.standardLibraryType(.int32), program.standardLibraryType(.uint32):
-      return .integer(value, program.types.demand(MachineType.i(32)))
-    case program.standardLibraryType(.int64), program.standardLibraryType(.uint64):
-      return .integer(value, program.types.demand(MachineType.i(64)))
+    // Keep word-sized integer bounds in sync with the generated standard library.
```
tothambrus11: You can use `program.show()` to display the string representation of types/declarations, which should simplify this here. Also we no longer need to eagerly create all these strings, only on error.

tothambrus11: It would be nice to factor out this switch into its own function(s), as now it takes up a great part of the function, distracting from the main logic the function is responsible for.

Sources/FrontEnd/IR/IREmitter.swift
```diff
@@ -1446,20 +1446,47 @@ internal struct IREmitter {
...
+      specification = (32, false, .i(32), "UInt32")
+    case program.standardLibraryType(.int64):
+      specification = (64, true, .i(64), "Int64")
+    case program.standardLibraryType(.uint64):
+      specification = (64, false, .i(64), "UInt64")
     default:
       program.unexpected(target)
     }
+
+    let unit = BigInt(1)
```
tothambrus11: Do you think we could cache minimum and maximum for all the standard library integer types? BigInt operations are relatively expensive.

Sources/FrontEnd/IR/IREmitter.swift
```diff
@@ -1446,20 +1446,47 @@ internal struct IREmitter {
...
+    let minimum = specification.isSigned ? -(unit << (specification.width - 1)) : 0
+    let maximum =
+      specification.isSigned
+      ? (unit << (specification.width - 1)) - 1
+      : (unit << specification.width) - 1
```
tothambrus11:
```suggestion
    let maximum = if isSigned {
        (unit << (specification.width - 1)) - 1
      } else {
        (unit << specification.width) - 1
      }
```
I would prefer explicit `if` expressions when longer expressions are involved. My reasoning is that `?` and `:` is relatively smaller than `if`/`else`, so they don't look significant enough when looking at a bigger expression.

Sources/FrontEnd/IR/IREmitter.swift
```diff
@@ -1446,20 +1446,47 @@ internal struct IREmitter {
...
+    if !(minimum...maximum).contains(value) {
+      let message =
+        "integer literal '\(program[source].value)' is out of bounds for "
+        + "'\(specification.name)'"
+      report(.error, message, about: source)
+    }
```
tothambrus11: Could this check happen at type-checking time rather than during IR emission? That way we get earlier feedback in editors when something is wrong (IREmitter only runs after we have a well-typed program).

As a reminder, we have parsing -> scope assignment -> type checking -> IR emission -> mandatory IR passes -> LLVM emission / interpretation.

Tests/CompilerTests/negative/integer-literal-bounds.hylo
ColumbusLabs: Changed the `UInt8` overflow case to the exact adjacent value `256` and updated the expected diagnostic.

Sources/FrontEnd/IR/IREmitter.swift
ColumbusLabs: The range construction is now isolated in explicit `signed(_:)` and `unsigned(_:)` helpers, so the longer ternary expressions are gone.

Sources/FrontEnd/IR/IREmitter.swift
ColumbusLabs: Added a static precomputed `BigInt` bounds table for all ten standard integer types, sharing the 64-bit ranges for `Int`/`Int64` and `UInt`/`UInt64`.

Sources/FrontEnd/IR/IREmitter.swift
ColumbusLabs: Yes — the check now runs during integer-literal type inference, the negative fixture is `stage:typing`, and IR emission is restored to storage-only lowering. The full pinned-LLVM suite passes 437/437.

Sources/FrontEnd/IR/IREmitter.swift
ColumbusLabs: Factored the mapping into `standardLibraryIntegerEntity(of:)`, `standardLibraryIntegerBounds(of:)`, and the cached bounds table so the literal inference path stays focused.

ColumbusLabs: Moved the diagnostic into `Typer`; `program.show(e)` and `program.show(t)` are now evaluated only inside the out-of-range error path.

tothambrus11 (changes requested):
Hi, welcome to the community, and thank you for the contribution!

I left some inline comments that we should discuss/address before merging. I'm happy to chat on Slack or video if you would like some more guidance on getting started.

I would like to ask you to avoid using AI for generating PR descriptions, comment replies and code, for the sake of your own learning opportunity. I've found that AI abstracts away the pain, but also the thinking, which is crucial when designing/implementing a new language. We also try to keep the code to high quality standards, which becomes hard to maintain if we had no control/ownership of the software. I usually use AI to gain understanding of a certain topic/implementation technique, and to do code-review on my changes after I've gone through it manually.

We are looking for some new long-term contributors, so I hope you can understand this. Please send a comment about this if you have acknowledged this.

tothambrus11:
Starts to look better, but unfortunately I'll need to close this until I'm sure you are a human contributor. Feel free to reopen a PR if you are ready.

PR #361, opened 2026-07-23
Add more builtin instructions; add Builtin.* tests

### LINKED ISSUE #358: Add more of the basic intrinsics, builtin tests, and standard library functions

tothambrus11:
This PR adds a handful of missing instructions and adds execution tests for all builtin instructions via the naming convention `Builtin.instruction_name`. Fixes https://github.com/hylo-lang/hylo-new/issues/358

I had to modify the driver to add a linking flag on Linux, as the floating point math may require linking the math library.

Float16 and float128 support is deferred to #435.

## Added instructions

Integer arithmetic
- `add`, `sub`, `mul`  wrapping and poisonous nuw/nsw variants
- `urem`, `srem`

Overflow-reporting arithmetic (return `{result, i1}` tuples)
- `uadd_with_overflow`, `usub_with_overflow`, `umul_with_overflow`

Bitwise logic
- `and`, `or`, `xor`

Integer width conversions
- `trunc`
- `zext`
- `sext`

Conversion between integer and floating-point numbers
- `uitofp`, `sitofp`
- `fptoui`, `fptosi`

Floating-point conversions
- `fptrunc`
- `fpext`

Floating-point arithmetic (all accept fast-math flags)
- `fadd`, `fsub`, `fmul`, `fdiv`, `frem`

Floating-point comparison (accepts fast-math flags and predicate selection)
- fcmp

Pointer arithmetic
- `advanced_by_bytes`

### DIFF
--- a/Sources/BackEnd/Program+LLVM.swift
+++ b/Sources/BackEnd/Program+LLVM.swift
@@ -299,9 +299,9 @@ extension Program {
       ctx.value[v] = metadata(of: t, in: &ctx.module).llvm.unsafe[].null
     case .advancedByBytes(let t):
       let p = insertLoad([s.arguments[0]], of: types.demand(MachineType.ptr), in: &ctx)[0]
-      let n = insertLoad([s.arguments[1]], of: t, in: &ctx)
+      let offsets = insertLoad([s.arguments[1]], of: t, in: &ctx)
       ctx.value[v] = ctx.module.llvm.insertGetElementPointerInBounds(
-        of: p, typed: ctx.module.llvm.i8, indices: n, at: ctx.insertionPoint!).v
+        of: p, typed: ctx.module.llvm.i8, indices: offsets , at: ctx.insertionPoint!).v
 
     case .add(let o, let t):
       let xs = insertLoad(s.arguments, of: t, in: &ctx)
--- a/Sources/BackEnd/Program+LLVM.swift
+++ b/Sources/BackEnd/Program+LLVM.swift
@@ -297,6 +297,61 @@ extension Program {
       ctx.value[v] = ctx.value[s.arguments[0]]!
     case .zeroinitializer(let t):
       ctx.value[v] = metadata(of: t, in: &ctx.module).llvm.unsafe[].null
+    case .advancedByBytes(byteOffset: let t):
+      assert(s.arguments.count == 2)
+      let p = insertLoad([s.arguments[0]], of: types.demand(MachineType.ptr), in: &ctx)[0]
+      let offsets = insertLoad([s.arguments[1]], of: t, in: &ctx)
+      ctx.value[v] = ctx.module.llvm.insertGetElementPointerInBounds(
+        of: p, typed: ctx.module.llvm.i8, indices: offsets, at: ctx.insertionPoint!).v
+
+    case .add(let o, let t):
+      let xs = insertLoad(s.arguments, of: t, in: &ctx)
+      ctx.value[v] = ctx.module.llvm.insertAdd(
+        overflow: o.llvm, xs[0].v, xs[1].v, at: ctx.insertionPoint!).v
+    case .signedAdditionWithOverflow(let t):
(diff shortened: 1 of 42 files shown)


dabrahams:
I'm sure this overlaps with #412 somewhat.

tothambrus11:
Yes, though I didn't add all instructions from there yet. I think we should factor out the built-in and standard library additions from that PR into separate, more complete PRs rather than one-off additions.

Do you think the offset_by_bytes could be expressed as some pointer casting to uint8 pointer, then advance and casting back?

dabrahams:
The advance part is exactly what this builtin implements.  I researched how to do this pretty carefully for the OFE.  If you want to try something else, be my guest, but I trust my earlier work.

dabrahams:
As for "more complete" or “one-off additions,” the latter are much easier to review.  Take care that you're not creating a lot of verification work for someone.

tothambrus11:
Cool, then I'll add that to this batch.

tothambrus11:
@dabrahams This PR now subsumes all builtins that are needed for #412.

dabrahams:
You should be able to get this to 100% patch coverage.

tothambrus11:
It turns out, we can't achieve 100% coverage here easily because [float16 and float128 are not yet supported](https://github.com/hylo-lang/hylo-new/issues/435). The remaining uncovered line is an `unreachable()` call.

Tests/CompilerTests/positive/Builtin.advanced_by_bytes.hylo
```diff
@@ -0,0 +1,42 @@
...
+/// Four bytes laid out contiguously, since all properties have the same alignment.
+struct Quad is Deinitializable {
+  public memberwise init
+  public var a: UInt8
+  public var b: UInt8
+  public var c: UInt8
+  public var d: UInt8
+}
+
+public fun main() {
```
tothambrus11, line 14: @RishabhRD Could you check if I'm introducing UB here? If yes, could you suggest a better test case to test `advanced_by_bytes` with?

dabrahams: There's no UB here.

Sources/BackEnd/Program+LLVM.swift
```diff
@@ -297,57 +297,151 @@ extension Program {
...
-      ctx.value[v] = ctx.module.llvm.insertUnsignedDiv(
-        exact: e, xs[0], xs[1], at: ctx.insertionPoint!).v
-    case .sdiv(let e, let t):
-      let xs = insertLoad(s.arguments, of: t, in: &ctx)
-      ctx.value[v] = ctx.module.llvm.insertSignedDiv(
-        exact: e, xs[0], xs[1], at: ctx.insertionPoint!).v
-    case .urem(let t):
+    case .advancedByBytes(let t):
+      let p = insertLoad([s.arguments[0]], of: types.demand(MachineType.ptr), in: &ctx)[0]
+      let n = insertLoad([s.arguments[1]], of: t, in: &ctx)
```
dabrahams: Reordering these things makes the diff bigger and harder to evaluate.  I don't love replacing `offsets` with `n`, which tends to imply its a single integer, rather than array.  I realize it will only have 1 element, but still…

Sources/BackEnd/Program+LLVM.swift
```diff
@@ -297,57 +297,151 @@ extension Program {
...
+      ctx.value[v] = ctx.module.llvm.insertGetElementPointerInBounds(
+        of: p, typed: ctx.module.llvm.i8, indices: n, at: ctx.insertionPoint!).v
```
```suggestion
      let offsets = insertLoad([s.arguments[1]], of: t, in: &ctx)
      ctx.value[v] = ctx.module.llvm.insertGetElementPointerInBounds(
        of: p, typed: ctx.module.llvm.i8, indices: offsets , at: ctx.insertionPoint!).v
```

PR #367, opened 2026-07-24
Optimize 'Typer.canDeriveCoercions'

kyouko-taiga:
This patch slightly optimizes the implementation of `Typer.canDeriveCoercions`, which is used during name resolution as conservative test to determine whether the typer should attempt to prove the equality of two types.

Three optimizations are implemented.
- The first simply applies memoization.
- The second leverages the idea that if a scope contains no local given, then any query to the implicit context can be forwarded to the parent scope.
- The third uses but masking to avoid branches and reduce memory footprint.

These optimizations result in a tiny but noticeable speedup on my laptop.

### DIFF
--- a/Sources/FrontEnd/Typer/Typer.swift
+++ b/Sources/FrontEnd/Typer/Typer.swift
@@ -91,7 +91,7 @@ public struct Typer {
     fileprivate var scopeToTypeOfSelf: [ScopeIdentity: AnyTypeIdentity?]
 
     /// The cache of `Typer.canDeriveCoercions(_:_:in:where:)`.
-    fileprivate var scopeToPossibleCoercions: [ScopeIdentity: [TypePair: Bool]]
+    fileprivate var scopeToPotentialCoercions: [ScopeIdentity: [TypePair: Bool]]
 
     /// The cache of `Typer.typeOfTraitSelf(in:)`.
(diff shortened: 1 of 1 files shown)


tothambrus11:
Very nice!

debug:
- before: 145s
- after: 100s
30% improvement

release:
- before: 25.9s
- after: 22.3s
13% improvement

Sources/FrontEnd/Typer/Typer.swift
```diff
@@ -105,8 +108,8 @@ public struct Typer {
     /// The cache of `Typer.standardLibraryType(_:)`.
     fileprivate var standardLibraryEntityToType: [Program.StandardLibraryEntity: AnyTypeIdentity]
 
-    /// The cache of `Typer.canDeriveCoercion(_:_:applying:)`.
-    fileprivate var canDeriveCoercion: [Given: [TypePair: (Bool, Bool)]]
+    /// The cache of `Typer.canDeriveCoercion(_:applying:)`.
+    fileprivate var canDeriveCoercion: [Given: [TypePair: UInt8]]
```
tothambrus11, line 112: I would add an `OptionSet` with `UInt8` raw representation. That way we can easily check the activeness of each member without manual bit manipulations, and it's a more descriptive in the API.

PR #377, opened 2026-07-26
Driver extract reusable functions from load module function

RishabhRD:
This PR is intended to resolve @dabrahams' [comment](https://github.com/hylo-lang/hylo-new/pull/270#discussion_r3608465797) on #270.

### DIFF
--- a/Sources/Driver/Driver.swift
+++ b/Sources/Driver/Driver.swift
@@ -303,8 +303,8 @@ public struct Driver {
     try a.write(into: f)
   }
 
-  /// Attempts to load `m` from the module cache if its fingerprint matches `s`,
-  /// returning whether `m` was loaded.
+  /// Loads `m` from the module cache and returns `true`
+  /// if `m`'s fingerprint matches `s`; returns `false` otherwise.
   private mutating func loadCached(
     _ m: Module.Name,
     matching s: [SourceFile]
@@ -332,7 +332,7 @@ public struct Driver {
     }
   }
 
-  /// Compile module named `n` from `sources` and returns its identifier.
+  /// Compiles the module named `n` from `sources`, returning its identifier.
   private mutating func compile(
     _ n: Module.Name,
     from sources: [SourceFile]
--- a/Sources/Driver/Driver.swift
+++ b/Sources/Driver/Driver.swift
@@ -285,32 +285,56 @@ public struct Driver {
   ) async throws {
     let sources = additionalSources + (try sources(at: root))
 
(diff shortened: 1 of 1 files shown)


dabrahams:
I believe I filed an issue for that comment

Sources/Driver/Driver.swift
```diff
@@ -285,32 +285,56 @@ public struct Driver {
...
+    if cachingIsEnabled {
+      if try loadCached(module, matching: sources) {
+        return
+      }
+    }
+    let m = try await compile(module, from: sources)
+    if cachingIsEnabled {
+      let a = try program.archive(module: m)
+      let f = moduleCachePath!.appending(component: module + ".hylomodule")
+      try a.write(into: f)
```
dabrahams: should this be a function?  Just asking.

Sources/Driver/Driver.swift
```diff
@@ -285,32 +285,56 @@ public struct Driver {
...
+    }
+  }
+
+  /// Attempts to load `m` from the module cache if its fingerprint matches `s`,
+  /// returning whether `m` was loaded.
```
dabrahams:
```suggestion
  /// Loads `m` from the module cache and returns `true` 
  // if `m`'s fingerprint matches `s`; returns `false` otherwise.
```

It could fail to load by throwing and then it would not return whether it was loaded.  What it really returns, I think, is whether its fingerprint matches.

Sources/Driver/Driver.swift
```diff
@@ -285,32 +285,56 @@ public struct Driver {
...
         Maybe the archive was compiled with a different version of the compiler. \
         Try erasing the module cache.
         """
       throw Error(message: m)
     }
+  }
 
-    // Compile the module from sources.
-    let m = program.demandModule(module)
+  /// Compile module named `n` from `sources` and returns its identifier.
```
dabrahams:
```suggestion
  /// Compiles the module named `n` from `sources`, returning its identifier.
```

dabrahams (commented):
comments inline

Sources/Driver/Driver.swift
RishabhRD: Yes. Extracted it to a separate function.

tothambrus11:
@RishabhRD I adopted your changes in #397. Could you review the broader changes?

dabrahams:
So is this PR superseded and thus closable?

dabrahams (approved):
LGTM, regardless

PR #390, opened 2026-08-01
Set the working directory of executable test cases to their parent folder

tothambrus11:
The generated binary is invoked with working directory set to the package test's root directory or
the parent of a single-file test case. This is needed for reliably locating test resources.


Tests/CompilerTests/positive/c-ffi-calls.package/main.hylo
```diff
@@ -1,57 +1,57 @@
-struct Point is Deinitializable {
```
tothambrus11, line 1: This is just changing CRLF -> LF line separators.

dabrahams: I've seen problems with newline encoding in other files recently.  I wonder if someone has bad git settings or something?

dabrahams (approved):
LGTM

Tests/CompilerTests/positive/c-ffi-calls.package/main.hylo
tothambrus11: I think I did. I used WSL from my Windows laptop. There is some .gitconfig setting that lets us prevent this kind of issue on a project level, I'll look into that.

PR #391, opened 2026-08-01
Add cache to Typer.applies

tothambrus11:
`applies` -> `takeSummonResults` were taking 93% of total CPU time when run on Dave's test. Adding a cache collapsed the 73-second type-checking to 2.8 seconds. The speedup is likely not this big in less pathological examples, but it's quite expensive to recompute `applies` on every statement.

The whole test suite runs 4 seconds faster (51s -> 47s), and a file containing 20 `let x<N> = 1`  in the same scope got speed up from 63ms -> 24ms in type-checking. If there are only 3 integer literals in the same scope, this already saves 3ms.
image
image

Note: I copied the gating of cache saving from `summon`, but I'm not sure if `hasImplicitOnStack` is relevant in this case.

### DIFF
--- a/Sources/FrontEnd/Typer/Typer.swift
+++ b/Sources/FrontEnd/Typer/Typer.swift
@@ -111,6 +111,23 @@ public struct Typer {
     /// The cache of `Typer.canDeriveCoercion(_:applying:)`.
     fileprivate var canDeriveCoercion: [Given: [TypePair: UInt8]]
 
+    /// A key identifying an extension-applicability query.
+    fileprivate struct AppliesKey: Hashable {
+
+      /// The scope in which the extension is being applied.
+      let s: ScopeIdentity
+
+      /// The extension being tested.
+      let d: ExtensionDeclaration.ID
+
+      /// The subject type.
+      let t: AnyTypeIdentity
+
+    }
+
+    /// The cache of `Typer.applies(_:to:in:)`.
+    fileprivate var applies: [AppliesKey: SummonResult?]
+
     /// Creates an instance for typing `m`, which is a module in `p`.
     fileprivate init(typing m: Module.ID, in p: Program) {
       self.moduleToIdentifierToDeclaration = .init(repeating: nil, count: p.modules.count)
@@ -130,6 +147,7 @@ public struct Typer {
       self.predefinedGivens = [:]
       self.standardLibraryEntityToType = [:]
       self.canDeriveCoercion = [:]
+      self.applies = [:]
     }
 
   }
@@ -4604,12 +4622,23 @@ public struct Typer {
     // Fast path: types are trivially equal.
     if t == u { return .init(witness: w, substitutions: .init(), penalties: 0) }
 
+    // Consult the cache.
+    let key = Memos.AppliesKey(s: s, d: d, t: t)
+    if !t[.hasVariable], let m = cache.applies[key] {
+      return m
+    }
+
     // Slow path: use the match judgement of implicit resolution to create a witness describing
     // "how" the type matches the extension.
     let (context, head) = program.types.contextAndHead(t)
     assert(context.usings.isEmpty)
     let thread = formThread(matching: w, to: head, in: .empty)
-    return takeSummonResults(from: [thread], in: s).uniqueElement
+    let result = takeSummonResults(from: [thread], in: s).uniqueElement
+
+    if !t[.hasVariable] && !hasImplicitOnStack() {
+      cache.applies[key] = result
+    }
+    return result
   }
 
   /// Returns the modules that are imported by `f`, which is in the module being typed.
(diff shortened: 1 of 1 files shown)


Sources/FrontEnd/Typer/Typer.swift
```diff
@@ -111,6 +111,23 @@ public struct Typer {
     /// The cache of `Typer.canDeriveCoercion(_:applying:)`.
     fileprivate var canDeriveCoercion: [Given: [TypePair: UInt8]]
 
+    /// A key identifying an extension-applicability query.
+    fileprivate struct AppliesKey: Hashable {
```
kyouko-taiga: It seems to me that the properties should be declared in the same order as the parameters of `applies(_:to:in:)`.

Also, I think the properties should have more descriptive names. I used single letter variables for parameters and local variables but not for stored properties.

Finally I'd like to see this struct declared elsewhere so that the list of memo dictionaries is clearer to see. There are other supporting types, like `TypePair`. Probably it's a good place to declare this key type.

kyouko-taiga: But! Look at other comments before refactoring this struct. I think the memo cache should defined differently anyway.

Sources/FrontEnd/Typer/Typer.swift
```diff
@@ -111,6 +111,23 @@ public struct Typer {
...
+      /// The extension being tested.
+      let d: ExtensionDeclaration.ID
+
+      /// The subject type.
+      let t: AnyTypeIdentity
+
+    }
+
+    /// The cache of `Typer.applies(_:to:in:)`.
+    fileprivate var applies: [AppliesKey: SummonResult?]
```
kyouko-taiga: Following established convention, this should be a map from scope to a map from extension/type to Booleans. It should be named `scopeToSomehingSomething`.

The reason for this convention is that oftentimes the answer in a scope is the same as the answer in a parent scope. Thus we can avoid needless re-computation and minimize memory footprint. You can look at how memoization works in functions like `summon` to understand what I mean.

Sources/FrontEnd/Typer/Typer.swift
```diff
@@ -4604,12 +4622,23 @@ public struct Typer {
     // Fast path: types are trivially equal.
     if t == u { return .init(witness: w, substitutions: .init(), penalties: 0) }
 
+    // Consult the cache.
```
kyouko-taiga, line 4619: Before this step you should look if there are local givens. If not, the query can be forwarded to the parent scope. You can copy this pattern from other functions.

Sources/FrontEnd/Typer/Typer.swift
```diff
@@ -4604,12 +4622,23 @@ public struct Typer {
...
+
     // Slow path: use the match judgement of implicit resolution to create a witness describing
     // "how" the type matches the extension.
     let (context, head) = program.types.contextAndHead(t)
     assert(context.usings.isEmpty)
     let thread = formThread(matching: w, to: head, in: .empty)
-    return takeSummonResults(from: [thread], in: s).uniqueElement
+    let result = takeSummonResults(from: [thread], in: s).uniqueElement
+
+    if !t[.hasVariable] && !hasImplicitOnStack() {
```
kyouko-taiga, line 4630: `hasImplicitOnStack` is important to keep. It is set when the typer is currently checking an implicit that is not yet available for resolution, but will once it has been checked. If we fill the memo cache now, this declaration will essentially remain invisible to implicit resolution.

kyouko-taiga (changes requested):
Good work. I think the structure of the cache should change to avoid duplicating work in nested scopes and reduce memory footprint.

dabrahams:
Follow up PR?

kyouko-taiga:
I wouldn't be opposed to but I think that'd be just churn. The changes I requested likely simplify this PR and maintains consistency.

I am happy to implement them myself if Ambrus is busy (or wants a well-deserved weekend far away from computers).

kyouko-taiga:
If we don't like my changes we can drop the last commit.

Sources/FrontEnd/Typer/Typer.swift
tothambrus11: Oh that's a great optimization! I saw the pattern but couldn't find the reason.

Sources/FrontEnd/Typer/Typer.swift
```diff
@@ -4610,33 +4599,36 @@ public struct Typer {
...
+    // If there aren't any givens in `scopeOfUse`, forward the query to the parent scope.
+    if givens(lexicallyIn: scopeOfUse).isEmpty, let p = program.parent(containing: scopeOfUse) {
+      return applies(d, to: t, in: p)
+    }
+
     let u = extendeeType(d)
     let w = WitnessExpression(value: .reference(.init(d)), type: u)
 
     // Fast path: types are trivially equal.
     if t == u { return .init(witness: w, substitutions: .init(), penalties: 0) }
```
tothambrus11, line 4617: I would keep the fast path before any checking of `givens(lexicallyIn)`, it's cheaper to compute than any cache lookup.

tothambrus11: Or maybe not, we call extendeeType.

tothambrus11:
Thanks for the changes, LGTM!

Issue #392, opened 2026-08-01
Over-pruning coercions due to caching


tothambrus11:
This example demonstrates that a stale result of the cache is used in the presence of local givens:

I expected this to type-check:
```hylo
trait P {}

struct Box<X> {}
struct Pair<X, Y> {}
struct Other {}

fun need<W where W is P>() {}

fun f<T, U, V where T == U, Box<U> == V>(x: Box<T>) {
  given Pair<Box<T>, Other> is P {}
  given <where Box<T> == V> => Pair<Box<T>, Box<T>> is P {}
  need<Pair<V, V>>()
}

```
But:
```
positive/coercion-derivability-cache.hylo:31.3-19: error: no coercion from 'P<Pair<V, V>> ==> [Void]() let -> Void' to '[Void]() let -> Void'
  need<Pair<V, V>>()
  ~~~~~~~~~~~~~~~~
positive/coercion-derivability-cache.hylo:31.3-19: note: no given instance of 'P<Pair<V, V>>' in this scope
  need<Pair<V, V>>()
  ~~~~~~~~~~~~~~~~
````

kyouko-taiga:
This is clearly a torturous example. That's all fine but it would help if you explained a bit what you tried to accomplish, what was the expected result, and what you observed instead.

kyouko-taiga:
Do you expect this program to type check? I don't think you provided any way for the compiler to derive an instance of `P` for `Pair<V, V>`. And therefore the compiler complains that it can't call `need`.

The message is arguably a bit cryptic and requires work, but it says that `need`, which is a context function `<W> => P<W> => Void -> Void` can't be coerced to a function `Void -> Void`, which is the type that is expected from the call. The second part of the diagnostic says _why_ the coercion failed. Here that's because the compiler couldn't find an instance of `P` for `Pair<V, V>` after having substituted `Pair<V, V>` for `W`. AFAICT that is all correct since none of your local givens help.

I'd like to point out that equality constraints are not like aliases and so it is non-trivial and very expensive for the compiler to derive `S<T> == S<U>` from `T == U`. I'm not saying that to absolve the compiler of any bug we may discover, but just as a reminder that equality constraints should be used sparingly in ordinary code. Here your example is best expressed with `f` having only one type parameter and possibly a type alias declaration for `V`. Hence, unless you want to make a point about a failure related specifically to equality constraints, I suggest you simplify your reproducer.

tothambrus11:
[...]

PR #395, opened 2026-08-02
Use OptionSet instead of manual bit-manipulations, add memoization guards

### LINKED ISSUE #375: add an `OptionSet` with `UInt8` raw representation for the Typer's bit-manipulations
I would add an `OptionSet` with `UInt8` raw representation. That way we can easily check the activeness of each member without manual bit manipulations, and it's a more descriptive in the API.

_Originally posted by @tothambrus11 in https://github.com/hylo-lang/hylo-new/pull/367#discussion_r3647392444_

tothambrus11:
Fixes #375 and #392

### DIFF
--- a/Sources/FrontEnd/Typer/Typer.swift
+++ b/Sources/FrontEnd/Typer/Typer.swift
@@ -109,7 +109,7 @@ public struct Typer {
     fileprivate var standardLibraryEntityToType: [Program.StandardLibraryEntity: AnyTypeIdentity]
 
     /// The cache of `Typer.canDeriveCoercion(_:applying:)`.
-    fileprivate var canDeriveCoercion: [Given: [TypePair: UInt8]]
+    fileprivate var canDeriveCoercion: [Given: [TypePair: CoercionCapability]]
 
     /// Creates an instance for typing `m`, which is a module in `p`.
     fileprivate init(typing m: Module.ID, in p: Program) {
@@ -279,6 +279,24 @@ public struct Typer {
       }
     }
     return u
+  }
+
+  /// The ways in which a given may contribute to the derivation of a coercion between a pair of
+  /// types `(a, b)`.
+  fileprivate struct CoercionCapability: OptionSet {
+
+    /// The raw representation.
+    var rawValue: UInt8 = 0
+
+    /// `a` unifies with at least one side of the given's head.
+    static let first = CoercionCapability(rawValue: 0b10)
+
+    /// `b` unifies with at least one side of the given's head.
+    static let second = CoercionCapability(rawValue: 0b01)
+
+    /// The given may witness coercions on both members of the pair.
+    static let both = CoercionCapability(rawValue: 0b11)
+
   }
 
   /// The occurrences of a particular capture in a local definition.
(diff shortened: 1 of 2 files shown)


Sources/FrontEnd/Typer/Typer.swift
```diff
@@ -288,6 +281,24 @@ public struct Typer {
     return u
   }
 
+  /// The ways in which a given may contribute to the derivation of a coercion between a pair of
+  /// types `(a, b)`.
```
tothambrus11, line 285: I had trouble writing a good description here, LMK if you can think of a better one.

kyouko-taiga:
TBH I think it's a lot of code for a very local thing. I would keep the status quo.

tothambrus11:
Ok, we can separate the two sets of changes. What do you think about the other part (related to gating the memoization)?

Sources/FrontEnd/Typer/Typer.swift
```diff
@@ -3960,57 +3972,77 @@ public struct Typer {
 
     // Make sure the cache key does not depend on the order in which `a` and `b` have been passed.
     let p = (b.bits < a.bits) ? Pair(b, a) : Pair(a, b)
-    if let memoized = cache.scopeToPotentialCoercions[scopeOfUse]?[p] { return memoized }
+
+    // The memo isn't keyed by the assumed givens of `environment`, so it can only serve queries
+    // answered against an empty environment. Variable-bearing keys are not memoized, as they are
+    // meaningless outside of the constraint system currently being solved.
```
kyouko-taiga, line 3978: LGTM

Sources/FrontEnd/Typer/Typer.swift
```diff
@@ -3960,57 +3972,77 @@ public struct Typer {
...
+    let memoizable =
+      environment.givens.isEmpty && !p.first[.hasVariable] && !p.second[.hasVariable]
```
kyouko-taiga, line 3980: Maybe not for this PR bur I think we should make an extension of TypePair to write `p.isClosed` as a shortcut for the above.

Sources/FrontEnd/Typer/Typer.swift
```diff
@@ -3960,57 +3972,77 @@ public struct Typer {
...
+  /// `.first` (or `.second`) iff `p.first` (or `p.second`) unifies with one side of `g`'s head.
+  private mutating func canDeriveCoercion(
+    _ p: Memos.TypePair, applying g: Given
+  ) -> CoercionCapability {
     if let memoized = cache.canDeriveCoercion[g]?[p] { return memoized }
 
+    // Do not memoize variable-bearing keys or results computed while givens are on stack.
+    // Such queries may yield different results in the future.
+    let memoizable =
+      !p.first[.hasVariable] && !p.second[.hasVariable] && !hasImplicitOnStack()
```
kyouko-taiga, line 4014: Yeah, now I definitely think we need the extension of `TypePair`.

kyouko-taiga (commented):
The gating part LGTM. I don't think we should go ahead with the option set.

Issue #321, opened 2026-07-18
Compute module interface hash


tothambrus11:
In a build system, we can compile each module to a separate object file, and link them together with a native linker. 

Separate compilation of modules allows implementations to change without having to recompile other modules depending on a changed module. Only when the public interface (signatures, publicly visible type layout, inlinable functions) change, do we need to recompile the dependent modules, otherwise we can get away with just relinking them.

For the build system to notice the changes, the compiler needs to output the signature of the module interface. This can be done using an additional flag `--emit-module-interface-hash-to=<file>` that outputs a hash during the compilation of the module.

kyouko-taiga:
> Separate compilation of modules allows implementations to change without having to recompile other modules depending on a changed module. Only when the public interface (signatures, publicly visible type layout, inlinable functions) change, do we need to recompile the dependent modules, otherwise we can get away with just relinking them.

It's not so simple. We'd also need significant changes to the parser to produce stable identities. Consider the following struct:

```hylo
public struct S {
  public fun f() -> Int { 1 }
  fun g() -> Bool { true }
  public fun h() -> Int { 2 }
}

given 
```

For simplicity we'll assume that `f`, `g`, and `h` get identified by the syntax trees 1, 2 and 3, respectively. The public API of the struct is composed of trees 1 and 3, since `g` is private. But now imagine someone swaps `f` and `g`. That will also swap their identities, causing the module's API to change.

We would have to ensure that identities that do not depend on private parts are resilient to changes to those parts, which I think can only be done by parsing source files twice. Fortunately, parsing is not the bottleneck, so maybe that's acceptable. But note that I'm only sketching a naive solution here. I'm sure other problems will arise if we think carefully.

This feature request is not specific to module dependencies. Incremental compilation could also operate under the same assumptions. That being said, I don't think we should rush into any solution here and I would add the "future" label to this issue. I know everyone dreams of the ability to perform incremental compilation at the character level but unfortunately that is not so simple and there are many things we'll have to figure out first. For a teaser, consider default parameters:

```hylo
fun foo() {}
public fun bar(x: Int = foo())
```

What happens when we import `bar` and call it with `bar()`? Can we resolve `foo`? What if `foo` had context parameters? These are all non-trivial questions that I think we can answer later.

tothambrus11:
I see, it gets tricky with identities. Maybe we could do this on the IR level, collecting the set of public mangled IR function names.

[...]

PR #397, opened 2026-08-02
Build system integration support

tothambrus11:
## Summary

This PR introduces the primitives that an external build system needs to compile a multi-module project with module-level incrementality:
- a way to compile one module at a time into a reusable module archive
- a way to import archives produced by earlier steps
- a stable "interface hash" that lets the build system decide whether dependents actually need to be recompiled.

I already made some proof of concept CMake scripts using this, and it works as expected (although CMake is not prepared for multiple sources producing single object files, so I had to use some hacks). 

## Example:

```bash
hc --module-name Foo 
    --emit object 
    -o Foo.o
    --emit-module-to build/Foo.hylomodule
    --emit-module-interface-hash-to build/Foo.hash 
    Sources/Foo

hc --module-name App
    --import Foo 
    --module-search-path build
    --emit object
    -o App.o
    Sources/App

clang -c "$(hc --print-stdlib-root)/shims.c" -o build/stdlib_shims.o

cd build
clang App.o Foo.o stdlib_shims.o  -o app
```

## New CLI flags:

- `--module-name <name>`: name of the module being compiled (default Main). Replaces the old behaviour of deriving the product name from a single input path, which was a bit arbitrary when multiple sources were given.
- `--import <module>`: make a module visible to the one being compiled. Each is loaded from `<module>.hylomodule` from the module search path.
- `--module-search-path <path>`: where to look for imported archives. This is now distinct from `-L`, which is only for native libraries.
- `--emit-module-to <file>`: write the compiled module's archive so others can import it.
- `--emit-module-interface-hash-to <file>`: write a hash a build system can use as a rebuild key for dependents. The file is only rewritten when the hash changes, so tools looking at modififcation time work efficiently.
- `--print-stdlib-root`: print the standard library root and exit.
- `--module-cache` now defaults to `<user caches dir>/hylo` instead of a .hylocache in the temp directory.
- `--emit object` now writes a single object file at -o (by default at `<module>.o`) instead of a directory of objects.

## Other CLI changes:
- `--verbose` notes now go to standard error to avoid mixing with any printed artifacts expected on stdout.
- CLI flags are validated before running the compilation.

## Driver changes

- A cache entry that fails to parse or whose fingerprint doesn't match is now silently recompiled and overwritten, instead of aborting the build with an error.

## CLI Tests
I added an end-to-end test suite that exercises the compiler through the CLI. This initially didn't contribute to code coverage, so I had to modify the CI script to merge the coverage results.

## Limitations
The module interface hash is currently computed by the hash of the entire archive, so currently any change causes a rebuild. In the future we could implement a more efficient solution, e.g. hash the set of IR functions' mangled names that are present in the module.

### DIFF
--- a/Sources/Driver/Driver.swift
+++ b/Sources/Driver/Driver.swift
@@ -34,11 +34,11 @@ public struct Driver {
   /// The code model for code generation.
   public var codeModel: CodeModel
 
-  /// The directories to pass to the linker as library search paths.
-  public var librarySearchPaths: [URL]
+  /// The linker's library search path.
+  public var librarySearchPath: [URL]
 
   /// The directories in which imported module archives (`.hylomodule` files) are searched.
-  public var moduleSearchPaths: [URL]
+  public var moduleSearchPath: [URL]
 
   /// The names of the native libraries to link (in addition to any imported Hylo dependencies).
   public var librariesToLink: [String]
@@ -88,7 +88,7 @@ public struct Driver {
     optimization: OptimizationLevel = .none,
     relocation: RelocationModel = Driver.defaultRelocationModel,
     codeModel: CodeModel = .default,
-    librarySearchPaths: [URL] = [], moduleSearchPaths: [URL] = [],
+    librarySearchPath: [URL] = [], moduleSearchPath: [URL] = [],
     librariesToLink: [String] = []
   ) {
     self.moduleCachePath = moduleCachePath
@@ -96,8 +96,8 @@ public struct Driver {
     self.optimization = optimization
     self.relocation = relocation
     self.codeModel = codeModel
-    self.librarySearchPaths = librarySearchPaths
-    self.moduleSearchPaths = moduleSearchPaths
+    self.librarySearchPath = librarySearchPath
+    self.moduleSearchPath = moduleSearchPath
     self.librariesToLink = librariesToLink
     self.program = .init()
   }
@@ -386,11 +386,11 @@ public struct Driver {
     return try? Data(contentsOf: c.appending(path: module + ".hylomodule"))
   }
 
-  /// Returns the location and contents of `module`'s archive found in `moduleSearchPaths`, if any.
+  /// Returns the location and contents of `module`'s archive found in `moduleSearchPath`, if any.
   ///
   /// - Throws if an archive exists but cannot be read.
   private func importedArchive(of module: Module.Name) throws -> (url: URL, data: Data)? {
-    for prefix in moduleSearchPaths {
+    for prefix in moduleSearchPath {
       let u = prefix.appending(path: module + ".hylomodule")
       if FileManager.default.fileExists(atPath: u.path) {
         do {
@@ -403,7 +403,7 @@ public struct Driver {
     return nil
   }
 
-  /// Loads `module` and its dependencies from archives found in `moduleSearchPaths`, returning the
+  /// Loads `module` and its dependencies from archives found in `moduleSearchPath`, returning the
   /// identity of `module`.
   @discardableResult
   public mutating func loadArchivedModule(_ module: Module.Name) throws -> Module.ID {
@@ -411,7 +411,7 @@ public struct Driver {
     return try loadArchivedModule(module, modulesStartedLoading: &ms)
   }
 
-  /// Loads `module` and its dependencies from archives found in `moduleSearchPaths`, returning the
+  /// Loads `module` and its dependencies from archives found in `moduleSearchPath`, returning the
   /// identity of `module` and using `modulesStartedLoading` to detect circular dependencies.
   @discardableResult
   private mutating func loadArchivedModule(
@@ -425,7 +425,7 @@ public struct Driver {
     }
 
     guard let (source, data) = try importedArchive(of: module) else {
-      let ps = moduleSearchPaths.map(\.path).joined(separator: ", ")
+      let ps = moduleSearchPath.map(\.path).joined(separator: ", ")
       throw Error(message: "no archive found for module '\(module)' in module search paths [\(ps)]")
     }
 
@@ -494,7 +494,7 @@ public struct Driver {
   /// - Throws: if the parent folder of `output` doesn't exist.
   private func linkExecutable(from objectFiles: [URL], writingTo output: URL) async throws {
     var arguments = ["-o", output.path]
-    arguments += librarySearchPaths.map({ "-L\($0.path)" })
+    arguments += librarySearchPath.map({ "-L\($0.path)" })
     arguments += librariesToLink.map({ "-l\($0)" })
     arguments += objectFiles.map(\.path)
 
--- a/Sources/Driver/Driver.swift
+++ b/Sources/Driver/Driver.swift
@@ -34,10 +34,13 @@ public struct Driver {
   /// The code model for code generation.
   public var codeModel: CodeModel
 
-  /// The list of directories in which to search for libraries to link.
+  /// The directories to pass to the linker as library search paths.
   public var librarySearchPaths: [URL]
 
-  /// The names of the native libraries to link.
+  /// The directories in which imported module archives (`.hylomodule` files) are searched.
+  public var moduleSearchPaths: [URL]
+
+  /// The names of the native libraries to link (in addition to any imported Hylo dependencies).
   public var librariesToLink: [String]
 
   /// `true` iff compilation and linking depend on the standard library and its shim.
@@ -65,11 +68,19 @@ public struct Driver {
 
   #if USE_BUNDLED_STANDARD_LIBRARY // Set compiler flag in distributable builds.
   /// The root folder of the standard library's sources.
-  private let chosenStandardLibraryRoot = bundledStandardLibrarySources
+  public static let standardLibraryRoot = bundledStandardLibrarySources
   #else
   /// The root folder of the standard library's sources.
-  private let chosenStandardLibraryRoot = localStandardLibrarySources
+  public static let standardLibraryRoot = localStandardLibrarySources
   #endif
+
+  /// The file within `standardLibraryRoot` implementing the standard library's
+  /// `@extern_c_indirect` declarations.
+  ///
+  /// Build systems linking Hylo objects must compile and link this except in freestanding mode.
+  public static var standardLibraryCShim: URL {
+    standardLibraryRoot.appending(component: cShimSource)
+  }
 
   /// Creates an instance with the given properties.
   public init(
@@ -77,7 +88,8 @@ public struct Driver {
     optimization: OptimizationLevel = .none,
     relocation: RelocationModel = Driver.defaultRelocationModel,
     codeModel: CodeModel = .default,
-    librarySearchPaths: [URL] = [], librariesToLink: [String] = []
+    librarySearchPaths: [URL] = [], moduleSearchPaths: [URL] = [],
+    librariesToLink: [String] = []
   ) {
     self.moduleCachePath = moduleCachePath
     self.target = targetSpecification
@@ -85,6 +97,7 @@ public struct Driver {
     self.relocation = relocation
     self.codeModel = codeModel
     self.librarySearchPaths = librarySearchPaths
+    self.moduleSearchPaths = moduleSearchPaths
     self.librariesToLink = librariesToLink
     self.program = .init()
   }
@@ -110,9 +123,7 @@ public struct Driver {
 
   /// Assigns the trees in `module` to their scopes.
   @discardableResult
-  public mutating func assignScopes(
-    of module: Module.ID
-  ) async -> PhaseResult {
+  public mutating func assignScopes(of module: Module.ID) async -> PhaseResult {
     let clock = ContinuousClock()
     let elapsed = await clock.measure {
       await program.assignScopes(module)
@@ -135,9 +146,7 @@ public struct Driver {
 
   /// Lowers the contents of `module` to IR.
   @discardableResult
-  public mutating func lower(
-    _ module: Module.ID
-  ) async -> PhaseResult {
+  public mutating func lower(_ module: Module.ID) async -> PhaseResult {
     let clock = ContinuousClock()
     let elapsed = clock.measure {
       program.lower(module)
@@ -147,9 +156,7 @@ public struct Driver {
 
   /// Applies mandatory transformation passes on the IR of `module`.
   @discardableResult
-  public mutating func applyTransformationPasses(
-    _ module: Module.ID
-  ) async -> PhaseResult {
+  public mutating func applyTransformationPasses(_ module: Module.ID) async -> PhaseResult {
     let elapsed = ContinuousClock().measure {
       program.applyTransformationPasses(module)
     }
@@ -168,31 +175,31 @@ public struct Driver {
     let elapsed = try ContinuousClock().measure {
       let t = TargetMachine(
         target: target, optimization: optimization, relocation: relocation, codeModel: codeModel)
-      var m = try program.compileToLLVM(module, target: t)
+      var llvm = try program.compileToLLVM(module, target: t)
 
-      try verify(m)
-      m.runDefaultModulePasses(optimization: optimization)
-      try verify(m)
+      try verify(llvm)
+      llvm.runDefaultModulePasses(optimization: optimization)
+      try verify(llvm)
 
-      llvmModules[module] = LLVMModuleBox(consume m)
+      llvmModules[module] = LLVMModuleBox(consume llvm)
     }
     return .init(elapsed: elapsed, containsError: program[module].containsError)
   }
 
-  /// Applies LLVM IR verification passes to `module` iff this file has been compiled in debug
+  /// Applies LLVM IR verification passes to `m` iff this file has been compiled in debug
   /// mode; does nothing otherwise.
   ///
-  /// - Throws: if the contents of `module` failed verification.
-  private func verify(_ module: borrowing SwiftyLLVM.Module) throws {
+  /// - Throws: if the contents of `m` failed verification.
+  private func verify(_ m: borrowing SwiftyLLVM.Module) throws {
     do {
-      try module.verifyInDebugBuilds()
+      try m.verifyInDebugBuilds()
     } catch let e as LLVMError {
       throw Error(
         message: """
         LLVM verification failed with the following message: \(e.description)
 
         Module contents:
-        \(module.description)
+        \(m.description)
         """)
     }
   }
@@ -215,11 +222,11 @@ public struct Driver {
       if usesStandardLibrary {
         // FIXME: Enable this after we can lower the standard library
         // modulesToLink.append(program.modules[.standardLibrary]!.identity)
-        cSources.append(chosenStandardLibraryRoot.appending(component: cShimSource))
+        cSources.append(Driver.standardLibraryCShim)
       }
 
       try await FileManager.default.withUniqueTemporaryDirectory { (d) in
-        let hyloObjects = try writeObjectFiles(for: modulesToLink, into: d)
+        let hyloObjects = try emitObjectFiles(for: modulesToLink, into: d)
         var cObjects: [URL] = []
         for s in cSources {
           cObjects.append(try await compileCToObject(source: s, destinationDirectory: d))
@@ -235,11 +242,11 @@ public struct Driver {
     llvmModules[module]?.module.llCode()
   }
 
-  /// Returns the assembly of module `m`.
+  /// Returns the assembly of `module`.
   ///
-  /// - Requires: module `m` has been lowered to LLVM.
-  public func assembly(of m: Module.ID) throws -> String {
-    try llvmModules[m]!.module.compile(.assembly).utf8Decoded
+  /// - Requires: `module` has been lowered to LLVM.
+  public func assembly(of module: Module.ID) throws -> String {
+    try llvmModules[module]!.module.compile(.assembly).utf8Decoded
       .unwrapOrThrow(Error(message: "Failed to decode assembly as an UTF8 string."))
   }
 
@@ -248,14 +255,21 @@ public struct Driver {
   /// - Requires: each element of `modules` has been lowered to LLVM.
   /// - Throws: if `destinationDirectory` doesn't exist.
   @discardableResult
-  public func writeObjectFiles(
+  public func emitObjectFiles(
     for modules: [Module.ID], into destinationDirectory: URL
   ) throws -> [URL] {
     try modules.map { (m) in
       let o = destinationDirectory.appendingPathComponent(moduleName(m) + ".o", isDirectory: false)
-      try llvmModules[m]!.module.write(.objectFile, to: o.path)
+      try emitObjectFile(of: m, to: o)
       return o
     }
+  }
+
+  /// Writes the object file of `module` to `output`.
+  ///
+  /// - Requires: `module` has been already lowered to LLVM.
+  public func emitObjectFile(of module: Module.ID, to output: URL) throws {
+    try llvmModules[module]!.module.write(.objectFile, to: output.path)
   }
 
   /// Compiles `source` using `clang` to an object file.
@@ -279,39 +293,52 @@ public struct Driver {
   ///
   /// If `moduleCachePath` is set, the module is loaded from cache if an archive is found and its
   /// fingerprint matches the fingerprint of the source files in `root`. Otherwise, the module is
-  /// compiled from sources and an archive is stored at `moduleCachePath`. If `moduleCachePath` is
-  /// not set, the module is unconditionally compiled from sources and no archive is stored.
+  /// compiled from sources and an archive is stored at `moduleCachePath`; A cached archive that
+  /// cannot be parsed is compiled from sources and overwritten.
   public mutating func load(
     _ module: Module.Name, withSourcesAt root: URL, additionalSources: [SourceFile] = []
   ) async throws {
     let sources = additionalSources + (try sources(at: root))
 
-    // Attempt to load the module from disk.
-    do {
-      if cachingIsEnabled, let data = archive(of: module) {
-        let h = SourceFile.fingerprint(contentsOf: sources)
-        var a = ReadableArchive(data)
-        let (_, fingerprint) = try Module.header(&a)
-        if h == fingerprint {
-          a = ReadableArchive(data)
-          try program.load(module: module, from: &a)
-          return
-        }
+    // If caching is enabled, load the module from disk or compile and cache it if that failed.
+    if cachingIsEnabled {
+      if try !loadCached(module, matching: sources) {
+        let m = try await compile(module, from: sources)
+        try writeToCache(m)
       }
-    } catch ArchiveError.invalidInput {
-      let m = """
-        Failed to parse module archive of '\(module)' at '\(moduleCachePath, default: "nil")'.
-
-        Maybe the archive was compiled with a different version of the compiler. \
-        Try erasing the module cache.
-        """
-      throw Error(message: m)
     }
+    // Otherwise, ignore the cache entirely.
+    else {
+      _ = try await compile(module, from: sources)
+    }
+  }
 
-    // Compile the module from sources.
-    let m = program.demandModule(module)
+  /// Loads `module` from its cache and returns `true` iff its fingerprint matches that of `s`
+  private mutating func loadCached(_ module: Module.Name, matching s: [SourceFile]) throws -> Bool {
+    guard let data = cachedArchive(of: module) else { return false }
+    do {
+      var a = ReadableArchive(data)
+      let (name, fingerprint) = try Module.header(&a)
+      // Invalid name or fingerprint.
+      guard name == module, fingerprint == SourceFile.fingerprint(contentsOf: s) else {
+        return false
+      }
 
-    if usesStandardLibrary && module != Module.standardLibraryName {
+      a = ReadableArchive(data)
+      try program.load(module: module, from: &a)
+      return true
+    } catch is ArchiveError {
+      return false
+    }
+  }
+
+  /// Compiles the module named `n` from `sources` to refined IR, returning its identifier.
+  private mutating func compile(
+    _ n: Module.Name, from sources: [SourceFile]
+  ) async throws -> Module.ID {
+    let m = program.demandModule(n)
+
+    if usesStandardLibrary && (n != Module.standardLibraryName) {
       program[m].addDependency(Module.standardLibraryName)
     }
 
@@ -330,39 +357,135 @@ public struct Driver {
     await applyTransformationPasses(m)
     try throwIfContainsError(m)
 
-    if cachingIsEnabled {
-      let a = try program.archive(module: m)
-      let f = moduleCachePath!.appending(component: module + ".hylomodule")
-      try a.write(into: f)
-    }
+    return m
+  }
+
+  /// Writes `module` to the module cache.
+  ///
+  /// - Requires: `cachingIsEnabled` is `true`.
+  private func writeToCache(_ module: Module.ID) throws {
+    let f = moduleCachePath!.appending(component: program[module].name + ".hylomodule")
+    try writeArchive(of: module, to: f)
   }
 
   /// Loads the standard library with `load(_:withSourcesAt:)` and makes
   /// standard library a dependency of modules loaded thereafter.
-  /// 
+  ///
   /// Use the `USE_BUNDLED_STANDARD_LIBRARY` compiler flag to control whether the  bundled or local
   /// standard library is used. Defaults to local.
   public mutating func loadStandardLibrary() async throws {
     try await load(
-      Module.standardLibraryName, withSourcesAt: chosenStandardLibraryRoot,
+      Module.standardLibraryName, withSourcesAt: Driver.standardLibraryRoot,
       additionalSources: [SourceFile(contentsOf: generatedStandardLibrarySource)])
     usesStandardLibrary = true
   }
 
-  /// Searches for an archive of `module` in `librarySearchPaths`, returning it if found.
-  public func archive(of module: Module.Name) -> Data? {
-    if let prefix = moduleCachePath {
-      let path = prefix.appending(path: module + ".hylomodule")
-      return try? Data(contentsOf: path)
-    } else {
-      return nil
+  /// Returns the archive of `module` in the module cache if present, `nil` otherwise.
+  internal func cachedArchive(of module: Module.Name) -> Data? {
+    guard let c = moduleCachePath else { return nil }
+    return try? Data(contentsOf: c.appending(path: module + ".hylomodule"))
+  }
+
+  /// Returns the location and contents of `module`'s archive found in `moduleSearchPaths`, if any.
+  ///
+  /// - Throws if an archive exists but cannot be read.
+  private func importedArchive(of module: Module.Name) throws -> (url: URL, data: Data)? {
+    for prefix in moduleSearchPaths {
+      let u = prefix.appending(path: module + ".hylomodule")
+      if FileManager.default.fileExists(atPath: u.path) {
+        do {
+          return (u, try Data(contentsOf: u))
+        } catch {
+          throw Error(message: "cannot read module archive at '\(u.path)': \(error)")
+        }
+      }
+    }
+    return nil
+  }
+
+  /// Loads `module` and its dependencies from archives found in `moduleSearchPaths`, returning the
+  /// identity of `module`.
+  @discardableResult
+  public mutating func loadArchivedModule(_ module: Module.Name) throws -> Module.ID {
+    var ms = Set<Module.Name>()
+    return try loadArchivedModule(module, modulesStartedLoading: &ms)
+  }
+
+  /// Loads `module` and its dependencies from archives found in `moduleSearchPaths`, returning the
+  /// identity of `module` and using `modulesStartedLoading` to detect circular dependencies.
+  @discardableResult
+  private mutating func loadArchivedModule(
+    _ module: Module.Name, modulesStartedLoading: inout Set<Module.Name>
+  ) throws -> Module.ID {
+    // Nothing to do if the module is already loaded.
+    if let module = program.identity(module: module) { return module }
+
+    if !modulesStartedLoading.insert(module).inserted {
+      throw Error(message: "circular dependency detected while loading module '\(module)'")
+    }
+
+    guard let (source, data) = try importedArchive(of: module) else {
+      let ps = moduleSearchPaths.map(\.path).joined(separator: ", ")
+      throw Error(message: "no archive found for module '\(module)' in module search paths [\(ps)]")
+    }
+
+    // TODO: save fingerprint of dependencies and check that they match the version `module` was
+    // compiled against.
+
+    // Ensure the dependencies are loaded before we load `module`.
+    var a = ReadableArchive(data)
+
+    let h: (name: Module.Name, fingerprint: UInt64, dependencies: [Module.Name])
+    do { h = try Module.headerAndDependencies(&a) }
+    catch { throw cannotParseArchive(of: module, at: source) }
+
+    guard h.name == module else {
+      throw Error(message: """
+        archive for '\(module)' at '\(source.path)' declares a different module name '\(h.name)'
+        """)
+    }
+    for d in h.dependencies {
+      try loadArchivedModule(d, modulesStartedLoading: &modulesStartedLoading)
+    }
+
+    var body = ReadableArchive(data)
+    do {
+      return try program.load(module: module, from: &body).identity
+    } catch is ArchiveError {
+      throw cannotParseArchive(of: module, at: source)
     }
   }
 
-  /// Throws the diagnostics of `m` if those contain an error.
-  private func throwIfContainsError(_ m: Module.ID) throws {
-    if program[m].containsError {
-      throw CompilationError(diagnostics: .init(program[m].diagnostics))
+  /// Returns an error reporting that the archive of `module` at `location` could not be parsed.
+  private func cannotParseArchive(of module: Module.Name, at location: URL) -> Error {
+    Error(message: """
+      Failed to parse the module archive of '\(module)' at '\(location.path)'.
+
+      Maybe the archive was compiled with a different version of the compiler.
+      """)
+  }
+
+  /// Writes the archive of `module` to `output`.
+  ///
+  /// The parent directories of `output` are expected to exist.
+  public func writeArchive(of module: Module.ID, to output: URL) throws {
+    try program.archive(module: module).write(into: output)
+  }
+
+  /// Returns a hash of `module` that a build system can use as a rebuild key for its dependents.
+  public func moduleInterfaceHash(of module: Module.ID) throws -> UInt64 {
+    // TODO: Make the interface hash resilient to changes in function bodies.
+    // https://github.com/hylo-lang/hylo-new/issues/321
+
+    // Note: avoiding `.hashValue` for consistency across runs and host platforms.
+    let a = try program.archive(module: module)
+    return FNV1.hash(a, into: FNV1.u64()).state
+  }
+
+  /// Throws the diagnostics of `module` if those contain an error.
+  private func throwIfContainsError(_ module: Module.ID) throws {
+    if program[module].containsError {
+      throw CompilationError(diagnostics: .init(program[module].diagnostics))
     }
   }
 
@@ -403,8 +526,8 @@ public struct Driver {
     var module: LLVMModule
 
     /// Wraps `module` by consuming it.
-    init(_ module: consuming LLVMModule) {
-      self.module = module
+    init(_ m: consuming LLVMModule) {
+      self.module = m
     }
 
   }
--- a/Sources/FrontEnd/Module.swift
+++ b/Sources/FrontEnd/Module.swift
@@ -550,6 +550,7 @@ extension Module: Archivable {
       }
 
       var me = Self(name: name, identity: ctx.identities[name]!)
+      me.dependencies = dependencies
 
       // <source count> <identity> <contents> ...
       let count = try Int(archive.readUnsignedLEB128())
@@ -650,7 +651,7 @@ extension Module: Archivable {
         try archive.write(s.variableToBinding, in: &ctx, sortedBy: \.key)
         try archive.write(s.syntaxToType, in: &ctx, sortedBy: \.key)
         try archive.write(s.nameToDeclaration, in: &ctx, sortedBy: \.key)
-        try archive.write(s.witnessTables, in: &ctx)
+        try archive.write(s.witnessTables, in: &ctx, sortedBy: \.key)
       }
 
       // IR functions.
@@ -680,6 +681,22 @@ extension Module: Archivable {
     let name = try archive.read(Name.self)
     let hash = try archive.read(UInt64.self, endianness: .little)
     return (name, hash)
+  }
+
+  /// Returns the name, fingerprint, and dependencies of the module stored in `archive`.
+  public static func headerAndDependencies<A>(
+    _ archive: inout ReadableArchive<A>
+  ) throws -> (name: Module.Name, fingerprint: UInt64, dependencies: [Module.Name]) {
+    let (name, hash) = try header(&archive)
+
+    // <dependency count> [<dependency name> ...]
+    let dependencyCount = try archive.readUnsignedLEB128()
+    var dependencies: [Module.Name] = []
+    for _ in 0 ..< dependencyCount {
+      dependencies.append(try archive.read(Name.self))
+    }
+
+    return (name, hash, dependencies)
   }
 
 }
--- a/Sources/hc/CommandLine.swift
+++ b/Sources/hc/CommandLine.swift
@@ -14,14 +14,14 @@ private typealias Module = FrontEnd.Module
   /// Configuration for this command.
   public static let configuration = CommandConfiguration(commandName: "hc", version: hyloVersion)
 
-  /// The directories to pass to the linker as library search paths.
+  /// The linker's library search path.
   @Option(
     name: [.customShort("L")],
     help: ArgumentHelp(
       "Add a directory to the linker's search path.",
       valueName: "path"),
     transform: URL.init(fileURLWithPath:))
-  private var librarySearchPaths: [URL] = []
+  private var librarySearchPath: [URL] = []
 
   /// The paths at which imported module archives (`.hylomodule`) may be found.
   @Option(
@@ -30,7 +30,7 @@ private typealias Module = FrontEnd.Module
       "Add a directory to the module search path, where imported module archives are found.",
       valueName: "path"),
     transform: URL.init(fileURLWithPath:))
-  private var moduleSearchPaths: [URL] = []
+  private var moduleSearchPath: [URL] = []
 
   /// The path containing cached module data.
   @Option(
@@ -232,8 +232,8 @@ private typealias Module = FrontEnd.Module
       optimization: optimized ? .aggressive : .none,
       relocation: relocationModel ?? Driver.defaultRelocationModel,
       codeModel: codeModel ?? .default,
-      librarySearchPaths: Array(librarySearchPaths.uniqued()),
-      moduleSearchPaths: Array(moduleSearchPaths.uniqued()))
+      librarySearchPath: Array(librarySearchPath.uniqued()),
+      moduleSearchPath: Array(moduleSearchPath.uniqued()))
 
     do {
       // Load the imported modules.
--- a/Sources/hc/CommandLine.swift
+++ b/Sources/hc/CommandLine.swift
@@ -36,7 +36,10 @@ private typealias Module = FrontEnd.Module
   @Option(
     name: [.customLong("module-cache")],
     help: ArgumentHelp(
-      "Specify the module cache path (default: a 'hylo' directory in the user's caches directory).",
+      """
+      Specify the module cache path (default: a 'hylo' directory in \
+      $\(CommandLine.defaultCacheRootVariable) or the user's caches directory).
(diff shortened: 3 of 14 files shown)


Sources/Driver/Driver.swift
```diff
@@ -33,9 +33,14 @@ public struct Driver {
   /// The code model for code generation.
   public var codeModel: CodeModel
 
-  /// The list of directories in which to search for libraries to link.
+  /// The list of directories in which to search for native libraries to link.
   public var librarySearchPaths: [URL]
 
+  /// The list of directories in which to search for imported module archives (`.hylomodule`).
+  ///
+  /// This is distinct from `librarySearchPaths`, which locates native libraries for the linker.
```
tothambrus11:
```suggestion
```

Sources/Driver/Driver.swift
```diff
@@ -64,26 +69,36 @@ public struct Driver {
...
   #else
   /// The root folder of the standard library's sources.
-  private let chosenStandardLibraryRoot = localStandardLibrarySources
+  public static let standardLibraryRoot = localStandardLibrarySources
   #endif
 
+  /// The file within `standardLibraryRoot` implementing the standard library's
+  /// `@extern_c_indirect` declarations.
+  ///
+  /// Clients that link Hylo objects, must compile and link this file unless no-std.
```
tothambrus11:
```suggestion
  /// Build systems that link Hylo objects must compile and link this file unless no-std.
```

Sources/Driver/Driver.swift
```diff
@@ -278,41 +300,48 @@ public struct Driver {
   ///
   /// If `moduleCachePath` is set, the module is loaded from cache if an archive is found and its
   /// fingerprint matches the fingerprint of the source files in `root`. Otherwise, the module is
-  /// compiled from sources and an archive is stored at `moduleCachePath`. If `moduleCachePath` is
-  /// not set, the module is unconditionally compiled from sources and no archive is stored.
+  /// compiled from sources and an archive is stored at `moduleCachePath`; a cached archive that
+  /// cannot be parsed is treated as missing and overwritten.
```
tothambrus11:
```suggestion
  /// cannot be parsed is compiled from sources and overwritten.
```

Sources/Driver/Driver.swift
```diff
@@ -278,41 +300,48 @@ public struct Driver {
...
+      try program.load(module: m, from: &a)
+      return true
+    } catch is ArchiveError {
+      return false
     }
+  }
 
-    // Compile the module from sources.
-    let m = program.demandModule(module)
+  /// Compiles the module named `n` from `sources`, returning its identifier.
```
tothambrus11:
```suggestion
  /// Compiles the module named `n` from `sources` to transformed IR, returning its identifier.
```

Sources/Driver/Driver.swift
```diff
@@ -329,32 +358,129 @@ public struct Driver {
...
   }
 
-  /// Searches for an archive of `module` in `librarySearchPaths`, returning it if found.
-  public func archive(of module: Module.Name) -> Data? {
-    if let prefix = moduleCachePath {
-      let path = prefix.appending(path: module + ".hylomodule")
-      return try? Data(contentsOf: path)
-    } else {
-      return nil
+  /// Returns the archive of `module` in the module cache if present, or `nil` otherwise.
```
tothambrus11:
```suggestion
  /// Returns the archive of `module` in the module cache if present, `nil` otherwise.
```

Sources/Driver/Driver.swift
```diff
@@ -329,32 +358,129 @@ public struct Driver {
...
+  internal func cachedArchive(of module: Module.Name) -> Data? {
+    guard let c = moduleCachePath else { return nil }
+    return try? Data(contentsOf: c.appending(path: module + ".hylomodule"))
+  }
+
+  /// Returns the location and contents of `module`'s archive found in `moduleSearchPaths`, or `nil`
+  /// if no search path contains an archive for `module`.
```
tothambrus11:
```suggestion
  /// Returns the location and contents of `module`'s archive found in `moduleSearchPaths`, if any.
```

Sources/Driver/Driver.swift
```diff
@@ -329,32 +358,129 @@ public struct Driver {
...
+  ///
+  /// - Throws: `Driver.Error` if an archive exists but cannot be read.
```
tothambrus11:
```suggestion
  /// - Throws if an archive exists but cannot be read.
```

Sources/Driver/Driver.swift
```diff
@@ -329,32 +358,129 @@ public struct Driver {
...
+    var ms = Set<Module.Name>()
+    return try loadArchivedModule(module, modulesStartedLoading: &ms)
+  }
+
+  /// Loads `module` and its dependencies from archives found in `moduleSearchPaths`, returning the
+  /// identity of `module`.
+  /// 
+  /// Uses `modulesStartedLoading` to detect circular dependencies upon recursive calls.
+  @discardableResult
+  private mutating func loadArchivedModule(_ module: Module.Name, modulesStartedLoading: inout Set<Module.Name>) throws -> Module.ID {
```
tothambrus11:
```suggestion
  private mutating func loadArchivedModule(
    _ module: Module.Name, modulesStartedLoading: inout Set<Module.Name>
  ) throws -> Module.ID {
```

Sources/Driver/Driver.swift
```diff
@@ -329,32 +358,129 @@ public struct Driver {
...
+    
+    guard let (source, data) = try importedArchive(of: module) else {
+      let ps = moduleSearchPaths.map(\.path).joined(separator: ", ")
+      throw Error(message: "no archive found for module '\(module)' in module search paths [\(ps)]")
+    }
+
+    // TODO: save fingerprint of dependencies and check that they match the version `module` was
+    // compiled against.
+
+    // Ensure the dependencies are loaded before `module` itself.
```
tothambrus11:
```suggestion
    // Ensure the dependencies are loaded before we load `module`.
```

Sources/Driver/Driver.swift
```diff
@@ -329,32 +358,129 @@ public struct Driver {
...
+  /// Writes the archive of `module` to `output`.
+  /// 
+  /// The parent directories of `output` are expected to exist.
+  public func writeArchive(of module: Module.ID, to output: URL) throws {
+    try program.archive(module: module).write(into: output)
+  }
+
+  /// Returns a hash of `module` that a build system can use as a rebuild key for its dependents.
+  ///
+  /// The hash is currently the fingerprint of the module's *entire* archive.
```
tothambrus11:
```suggestion
  /// The hash is currently the fingerprint of the module's entire archive.
```

Sources/FrontEnd/Module.swift
```diff
@@ -650,7 +651,7 @@ extension Module: Archivable {
         try archive.write(s.variableToBinding, in: &ctx, sortedBy: \.key)
         try archive.write(s.syntaxToType, in: &ctx, sortedBy: \.key)
         try archive.write(s.nameToDeclaration, in: &ctx, sortedBy: \.key)
-        try archive.write(s.witnessTables, in: &ctx)
+        try archive.write(s.witnessTables, in: &ctx, sortedBy: \.key)
```
tothambrus11, line 654: @kyouko-taiga Is there a reason why they were not sorted before? That meant that archives were never generated bit-identically due to the randomization of hash seeds at process startup.

Sources/hc/CommandLine.swift
```diff
@@ -104,6 +119,48 @@ private typealias Module = FrontEnd.Module
...
+
+  /// The names of the modules that the compiled module depends on.
+  ///
+  /// Each is loaded from a `<name>.hylomodule` archive found in the module search paths.
+  @Option(
+    name: [.customLong("import")],
+    help: ArgumentHelp(
+      """
+      Make the given module visible to the module being compiled. Pass this option once for each
+      dependency.
```
```suggestion
      Make the given module visible to the module being compiled. Pass this option once for each \
      dependency.
```

Sources/hc/CommandLine.swift
```diff
@@ -104,6 +119,48 @@ private typealias Module = FrontEnd.Module
...
+    transform: URL.init(fileURLWithPath:))
+  private var moduleArchiveURL: URL?
+
+  /// The path at which the compiled module's interface hash should be written, if any.
+  @Option(
+    name: [.customLong("emit-module-interface-hash-to")],
+    help: ArgumentHelp(
+      """
+      Write a hash of the module's observable interface to <file>. Stable across body-only \
+      changes, so build systems can skip recompiling dependents when it is unchanged.
```
```suggestion
      Write a hash of the module's observable interface to <file>. When unchanged, \
      build systems may skip recompiling dependents.
```

Sources/hc/CommandLine.swift
```diff
@@ -130,32 +187,70 @@ private typealias Module = FrontEnd.Module
...
+    if noCaching && moduleCachePath != nil {
+      throw ValidationError("'--no-caching' and '--module-cache' are mutually exclusive")
+    }
+  }
+
   /// Executes the command.
   public mutating func run() async throws {
-    try configureSearchPaths()
+    // Answered before any other work: this query takes no inputs and must not have the side
+    // effects of setting up a module cache.
```
```suggestion
```

Sources/hc/CommandLine.swift
```diff
@@ -190,9 +285,13 @@ private typealias Module = FrontEnd.Module
         return
       }
 
-      // Write the module to the cache for future runs.
-      let a = try driver.program.archive(module: module)
-      note("module archive size: \(a.count)")
+      try emitInterfaceHash(of: module, from: driver)
```
tothambrus11: rename to emitInterfaceHashIfNeeded

Sources/hc/CommandLine.swift
```diff
@@ -190,9 +285,13 @@ private typealias Module = FrontEnd.Module
...
+
+      // Emit the module's archive so that other modules can import it.
+      if let u = moduleArchiveURL {
+        try driver.writeArchive(of: module, to: u)
+        note("written \(u.path)")
+      }
```
tothambrus11: refactor to writeArchiveIfNeeded

Sources/hc/CommandLine.swift
```diff
@@ -342,21 +440,31 @@ private typealias Module = FrontEnd.Module
...
 
-  /// Sets up the value of search paths for locating libraries and cached artifacts.
-  private mutating func configureSearchPaths() throws {
-    let m = FileManager.default
-    if let cache = moduleCachePath {
-      librarySearchPaths.append(cache)
+  /// Ensures that the latest interface hash of the currently compiled module is emitted to disk
+  /// if requested.
+  ///
+  /// Only rewrites the file when the content actually changes, as some build tools depend on
```
```suggestion
  /// Only rewrites the file when the content actually changes, as some build tools only track
```

Sources/hc/CommandLine.swift
```diff
@@ -447,6 +544,27 @@ private typealias Module = FrontEnd.Module
     /// Executable binary.
     case binary = "binary"
 
+    /// `true` iff the invocation may emit a module archive or interface hash; the tree and IR
+    /// dumps are pure dumps and stop before any other artifact is emitted.
```
```suggestion
    /// `true` iff the invocation may emit a module archive or interface hash. 
    /// IR lowering and mandatory transformations are required.
```

Sources/hc/CommandLine.swift
```diff
@@ -447,6 +544,27 @@ private typealias Module = FrontEnd.Module
...
+    var supportsModuleEmission: Bool {
+      switch self {
+      case .ast, .typedAST, .rawIR, .ir:
+        return false
+      case .llvm, .asm, .object, .binary:
+        return true
+      }
```
```suggestion
      case .ast, .typedAST, .rawIR:
        return false
      case .ir, .llvm, .asm, .object, .binary:
        return true
      }
```

Sources/hc/CommandLine.swift
```diff
@@ -473,12 +591,14 @@ private typealias Module = FrontEnd.Module
...
-  private func objectFilesDirectory() throws -> URL {
-    guard outputURL?.relativePath != "-" else {
+  /// Given the desired name of the compiler's product, returns the file to write when "object" is
+  /// selected as the output type.
+  private func objectFile(_ productName: Module.Name) throws -> URL {
+    if outputURL?.relativePath == "-" {
       throw ValidationError("object files cannot be written to the standard output")
+    } else {
+      return outputURL ?? URL(fileURLWithPath: "\(productName).o")
     }
```
tothambrus11: We should catch this in the validation phase.

Sources/Driver/Driver.swift
```diff
@@ -329,30 +356,128 @@ public struct Driver {
...
 
-  /// Searches for an archive of `module` in `librarySearchPaths`, returning it if found.
-  public func archive(of module: Module.Name) -> Data? {
-    if let prefix = moduleCachePath {
-      let path = prefix.appending(path: module + ".hylomodule")
-      return try? Data(contentsOf: path)
-    } else {
-      return nil
+  /// Returns the archive of `module` in the module cache if present, `nil` otherwise.
+  internal func cachedArchive(of module: Module.Name) -> Data? {
```
dabrahams, line 384: Repeating a thing's type in its name is not great.  Why is `module`  better than `m` here?  Remember that the comment will be read along with the signature.  See other instances in this file.

tothambrus11: I changed this pattern throughout the file.

tothambrus11:
It seems that different module archives are produced for the main module based on whether we loaded the standard library from cache or whether we compiled it from sources. This currently breaks the test in parallel test runs. @kyouko-taiga Do you see why that might happen?

kyouko-taiga:
If we build a module from source we may generate different type identities because running inference will reserve some space in the type store for unification variables. That should not be an issue outside of our test suite because the guarantee is that serialization always produces the same byte-wise artifacts only when the initial conditions are the same. Compiling dependencies from source or reusing serialized versions changes these initial conditions.

tothambrus11:
Thanks! CMake builds will always have the guarantee that modules are compiled separately, and dependencies are always loaded from archives. That won't necessarily be the case for our own SPM-like build system that can do more things in process and without having to re-read stuff from archives. We can worry about that later, until then it's just something to keep in mind might bite us.

dabrahams:
Nit: I don't think we want to bake in a suboptimal _guarantee-always_ to the CMake system.  Someday we might decide to modify CMake to get more efficient builds.

Sources/FrontEnd/Module.swift
kyouko-taiga: Looks like an oversight on my part.

Sources/Driver/Driver.swift
```diff
@@ -33,9 +33,12 @@ public struct Driver {
   /// The code model for code generation.
   public var codeModel: CodeModel
 
-  /// The list of directories in which to search for libraries to link.
+  /// The list of directories in which to search for native libraries to link.
```
kyouko-taiga: What's the distinction?

Sources/Driver/Driver.swift
```diff
@@ -33,9 +33,12 @@ public struct Driver {
...
   public var librarySearchPaths: [URL]
 
+  /// The list of directories in which to search for imported module archives (`.hylomodule`).
```
```suggestion
  /// The list of directories in imported module archives (i.e., `.hylomodule` files) are searched .
```

Sources/Driver/Driver.swift
```diff
@@ -64,26 +67,36 @@ public struct Driver {
...
   #else
   /// The root folder of the standard library's sources.
-  private let chosenStandardLibraryRoot = localStandardLibrarySources
+  public static let standardLibraryRoot = localStandardLibrarySources
   #endif
 
+  /// The file within `standardLibraryRoot` implementing the standard library's
+  /// `@extern_c_indirect` declarations.
+  ///
+  /// Build systems that link Hylo objects must compile and link this file unless no-std.
```
```suggestion
  /// Build systems that link Hylo objects must compile and link this file unless `no-std` is set.
```
I think that is not only `no-std` but also `freestanding`, which may be even more relevant. `no-std` is mostly for development purposes. Freestanding programs won't link to libc but they will still have a standard library.

Sources/Driver/Driver.swift
```diff
@@ -107,114 +118,107 @@ public struct Driver {
     return .init(elapsed: elapsed, containsError: program[module].containsError)
   }
 
-  /// Assigns the trees in `module` to their scopes.
+  /// Assigns the trees in `m` to their scopes.
   @discardableResult
-  public mutating func assignScopes(
-    of module: Module.ID
-  ) async -> PhaseResult {
+  public mutating func assignScopes(of m: Module.ID) async -> PhaseResult {
```
kyouko-taiga: Please keep the naming of the parameter consistent across all these driver functions. If you rename it here you should rename it in `parse(_:into:)`. FWIW I prefer the status quo and I'm not sure this change improves readability. I'm also pretty sure I used the single letter `m` in the implementation of some of the methods you have renamed.

Sources/Driver/Driver.swift
```diff
@@ -251,11 +255,18 @@ public struct Driver {
...
-      try llvmModules[m]!.module.write(.objectFile, to: o.path)
+      try writeObjectFile(for: m, to: o)
       return o
     }
   }
 
+  /// Writes the object file of `m` to `output`.
+  ///
+  /// - Requires: `m` has been already lowered to LLVM.
+  public func writeObjectFile(for m: Module.ID, to output: URL) throws {
```
kyouko-taiga: The documentation is inconsistent with the labels of the function.

Sources/Driver/Driver.swift
```diff
@@ -278,39 +289,47 @@ public struct Driver {
   ///
   /// If `moduleCachePath` is set, the module is loaded from cache if an archive is found and its
   /// fingerprint matches the fingerprint of the source files in `root`. Otherwise, the module is
-  /// compiled from sources and an archive is stored at `moduleCachePath`. If `moduleCachePath` is
-  /// not set, the module is unconditionally compiled from sources and no archive is stored.
+  /// compiled from sources and an archive is stored at `moduleCachePath`; a cached archive that
```
```suggestion
  /// compiled from sources and an archive is stored at `moduleCachePath`. A cached archive that
```

Sources/Driver/Driver.swift
```diff
@@ -278,39 +289,47 @@ public struct Driver {
...
+  /// cannot be parsed is compiled from sources and overwritten.
   public mutating func load(
     _ module: Module.Name, withSourcesAt root: URL, additionalSources: [SourceFile] = []
   ) async throws {
     let sources = additionalSources + (try sources(at: root))
 
-    // Attempt to load the module from disk.
+    if cachingIsEnabled, try loadCached(module, matching: sources) {
+      return
+    }
```
```suggestion
    if cachingIsEnabled, try loadCached(module, matching: sources) { return }
    
    // Compile the module from sources.
```

kyouko-taiga: I think the control flow can be expressed more simply.

```
// If caching is enabled, load the module from disk or compile and cache it if that failed.
if cachingIsEnabled {
  if try !loadCached(module, matching: sources) {
    let m = try await compile(module, from: sources)
    try writeArchiveToCache(of: m)
  }

// Otherwise, ignore the cache entirely.
} else {
  try await compile(module, from: sources)
}
```

Repeating the call to `compile` is warranted here because it makes the control flow easier to follow. Early returns are almost always hurting readability unless they are very simple guards.

Sources/Driver/Driver.swift
```diff
@@ -278,39 +289,47 @@ public struct Driver {
...
+    let m = try await compile(module, from: sources)
+    if cachingIsEnabled {
+      try writeArchiveToCache(of: m)
+    }
+  }
+
+  /// Loads `m` from the module cache and returns `true` iff `m`'s cached fingerprint matches the
+  /// fingerprint of `s`.
```
```suggestion
  /// Loads `m` from the module cache and returns `true` iff its fingerprint matches that of `s`
```

Sources/Driver/Driver.swift
```diff
@@ -278,39 +289,47 @@ public struct Driver {
...
+      try program.load(module: m, from: &a)
+      return true
+    } catch is ArchiveError {
+      return false
     }
+  }
 
-    // Compile the module from sources.
-    let m = program.demandModule(module)
+  /// Compiles the module named `n` from `sources` to transformed IR, returning its identifier.
```
```suggestion
  /// Compiles the module named `n` from `sources` to IR, returning its identifier.
```
Transformed IR is not a thing. If you want to qualify IR after mandatory passes, the adjective is "refined". But it should be understood that "IR" without qualification means "refined IR", at least outside of emitter.

Sources/Driver/Driver.swift
```diff
@@ -278,39 +289,47 @@ public struct Driver {
...
+  private mutating func compile(
+    _ n: Module.Name, from sources: [SourceFile]
+  ) async throws -> Module.ID {
+    let m = program.demandModule(n)
 
-    if usesStandardLibrary && module != Module.standardLibraryName {
+    if usesStandardLibrary && n != Module.standardLibraryName {
```
```suggestion
    if usesStandardLibrary && (n != Module.standardLibraryName) {
```

Sources/Driver/Driver.swift
```diff
@@ -329,35 +348,133 @@ public struct Driver {
...
 
-    if cachingIsEnabled {
-      let a = try program.archive(module: m)
-      let f = moduleCachePath!.appending(component: module + ".hylomodule")
-      try a.write(into: f)
-    }
+    return m
+  }
+
+  /// Writes the archive of `m` to the module cache.
```
```suggestion
  /// Writes `m` to the module cache.
```
The archive of `m` *is* `m`.

Sources/Driver/Driver.swift
```diff
@@ -329,35 +348,133 @@ public struct Driver {
...
+  ///
+  /// - Requires: `cachingIsEnabled` is `true`.
+  private func writeArchiveToCache(of m: Module.ID) throws {
```
```suggestion
  private func writeToCache(_ m: Module.ID) throws {
```

Sources/Driver/Driver.swift
```diff
@@ -329,35 +348,133 @@ public struct Driver {
...
+  @discardableResult
+  public mutating func loadArchivedModule(_ m: Module.Name) throws -> Module.ID {
+    var ms = Set<Module.Name>()
+    return try loadArchivedModule(m, modulesStartedLoading: &ms)
+  }
+
+  /// Loads `m` and its dependencies from archives found in `moduleSearchPaths`, returning the
+  /// identity of `m`.
+  ///
+  /// Uses `modulesStartedLoading` to detect circular dependencies upon recursive calls.
```
```suggestion
  /// Loads `m` and its dependencies from archives found in `moduleSearchPaths`, returning the
  /// identity of `m` and using `modulesStartedLoading` to detect circular dependencies.
```

Sources/Driver/Driver.swift
```diff
@@ -329,35 +348,133 @@ public struct Driver {
...
+      throw Error(message: "no archive found for module '\(m)' in module search paths [\(ps)]")
+    }
+
+    // TODO: save fingerprint of dependencies and check that they match the version `module` was
+    // compiled against.
+
+    // Ensure the dependencies are loaded before we load `m`.
+    var a = ReadableArchive(data)
+    let h = try (try? Module.headerAndDependencies(&a))
+      .unwrapOrThrow(cannotParseArchive(of: m, at: source))
```
kyouko-taiga: The double `try` is extremely jarring. Perhaps that can be expressed better with a `Result`?

Sources/Driver/Driver.swift
```diff
@@ -329,35 +348,133 @@ public struct Driver {
...
+    Error(message: """
+      Failed to parse the module archive of '\(m)' at '\(location.path)'.
+
+      Maybe the archive was compiled with a different version of the compiler.
+      """)
+  }
+
+  /// Writes the archive of `m` to `output`.
+  ///
+  /// The parent directories of `output` are expected to exist.
```
kyouko-taiga, line 470:
```suggestion
  /// - Requires: The parent directories of `output` exist.
```

Sources/Driver/Driver.swift
```diff
@@ -329,35 +348,133 @@ public struct Driver {
...
+  public func writeArchive(of m: Module.ID, to output: URL) throws {
+    try program.archive(module: m).write(into: output)
+  }
+
+  /// Returns a hash of `m` that a build system can use as a rebuild key for its dependents.
+  ///
+  /// The hash is currently the fingerprint of the module's entire archive.
```
```suggestion
  /// The hash is the fingerprint of the module's entire archive.
```
Everything is "currently" what it is.

Sources/Driver/Driver.swift
```diff
@@ -329,35 +348,133 @@ public struct Driver {
...
+  ///
+  /// - TODO: Optimize this to hash only `module`'s externally observable interface (e.g., function
+  ///   signatures, `@frozen` type layouts, inlinable/generic bodies), so that body-only changes
+  ///   leave the hash unchanged and importing modules don't need to be recompiled.
```
kyouko-taiga: I don't think that belongs here. Either write that in an issue or in the PR comments. There may be need for discussions around what goes into fingerprinting (what's suggested here is not enough) and we should not version the design process in the source code.

Sources/hc/CommandLine.swift
```diff
@@ -14,20 +14,29 @@ private typealias Module = FrontEnd.Module
   /// Configuration for this command.
   public static let configuration = CommandConfiguration(commandName: "hc", version: hyloVersion)
 
-  /// The paths at which libraries may be found.
+  /// The paths at which native libraries to link may be found.
```
kyouko-taiga: What's a native library if not just a library?

Sources/hc/CommandLine.swift
```diff
@@ -14,20 +14,29 @@ private typealias Module = FrontEnd.Module
...
     help: ArgumentHelp(
-      "Add a directory to the library search path.",
+      "Add a directory to the native library search path.",
       valueName: "path"),
     transform: URL.init(fileURLWithPath:))
   private var librarySearchPaths: [URL] = []
 
+  /// The paths at which imported module archives (`.hylomodule`) may be found.
+  @Option(
+    name: [.customShort("I"), .customLong("module-search-path")],
```
kyouko-taiga: We should keep `-I` for C search paths. It is very likely that we'll want to use `hc` with mixed source files.

Sources/hc/CommandLine.swift
```diff
@@ -14,20 +14,29 @@ private typealias Module = FrontEnd.Module
...
+      valueName: "path"),
+    transform: URL.init(fileURLWithPath:))
+  private var moduleSearchPaths: [URL] = []
+
   /// The path containing cached module data.
   @Option(
     name: [.customLong("module-cache")],
     help: ArgumentHelp(
-      "Specify the module cache path.",
+      "Specify the module cache path (default: a 'hylo' directory in the user's caches directory).",
```
kyouko-taiga: There should be an option to specify that when we compile `hc` and therefore I would not hardcode this default in the command line.

Sources/hc/CommandLine.swift
```diff
@@ -90,6 +99,12 @@ private typealias Module = FrontEnd.Module
     help: "Do not load the standard library")
   private var noStandardLibrary: Bool = false
 
+  /// `true` iff the compiler should print the standard library's root and exit.
+  @Flag(
+    name: [.customLong("print-stdlib-root")],
```
kyouko-taiga, line 107:
```suggestion
    name: [.customLong("show-stdlib-root")],
```

Sources/hc/CommandLine.swift
```diff
@@ -90,6 +99,12 @@ private typealias Module = FrontEnd.Module
...
+    help: "Print the path of the standard library's root directory and exit.")
+  private var printStandardLibraryRoot: Bool = false
```
kyouko-taiga, line 109:
```suggestion
  private var showStandardLibraryRoot: Bool = false
```

Sources/hc/CommandLine.swift
kyouko-taiga: I'd like to use the term "show" to show debug information.

Sources/hc/CommandLine.swift
```diff
@@ -104,6 +119,48 @@ private typealias Module = FrontEnd.Module
     help: "Trace type inference")
   private var lineTracingInference: LineLocator?
 
+  /// The name of the module being compiled.
+  @Option(
+    name: [.customLong("module-name")],
+    help: ArgumentHelp(
+      "The name of the module being compiled.",
+      valueName: "name"))
+  private var moduleName: String = "Main"
```
kyouko-taiga: Is that a good default? Does it have an impact on the binary name?

If we write `foo.hylo` and compile it with `hc -o foo foo.hylo` the module name is `Main`? I think that would be a little odd.

Sources/hc/CommandLine.swift
```diff
@@ -104,6 +119,48 @@ private typealias Module = FrontEnd.Module
...
+
+  /// The names of the modules that the compiled module depends on.
+  ///
+  /// Each is loaded from a `<name>.hylomodule` archive found in the module search paths.
```
```suggestion
  /// Each dependency is loaded from a `<name>.hylomodule` archive found in the module search paths.
```

Sources/hc/CommandLine.swift
```diff
@@ -104,6 +119,48 @@ private typealias Module = FrontEnd.Module
...
 
+  /// The name of the module being compiled.
+  @Option(
+    name: [.customLong("module-name")],
+    help: ArgumentHelp(
+      "The name of the module being compiled.",
+      valueName: "name"))
+  private var moduleName: String = "Main"
+
+  /// The names of the modules that the compiled module depends on.
```
```suggestion
  /// The names of the modules that are dependencies of the module being compiled.
```

Sources/hc/CommandLine.swift
```diff
@@ -104,6 +119,48 @@ private typealias Module = FrontEnd.Module
...
+      dependency.
+      """,
+      valueName: "module"))
+  private var imports: [String] = []
+
+  /// The path at which the archive of the compiled module should be written, if any.
+  @Option(
+    name: [.customLong("emit-module-to")],
+    help: ArgumentHelp(
+      "Write the compiled module's archive to <file>, so other modules can import it.",
```
```suggestion
      "Serializes the compiled module to <file>, so other modules can import it.",
```

Sources/hc/CommandLine.swift
```diff
@@ -130,32 +187,71 @@ private typealias Module = FrontEnd.Module
   /// Creates a new instance with default options.
   public init() {}
 
+  /// Checks that the parsed arguments form a consistent configuration.
+  public func validate() throws {
+    if moduleArchiveURL != nil && !outputType.supportsModuleEmission {
```
```suggestion
    if (moduleArchiveURL != nil) && !outputType.supportsModuleEmission {
```

Sources/hc/CommandLine.swift
```diff
@@ -130,32 +187,71 @@ private typealias Module = FrontEnd.Module
...
+      throw ValidationError(
+        "'--emit-module-to' cannot be used with '--emit \(outputType.rawValue)'")
+    }
+    if writeModuleInterfaceHashAt != nil && !outputType.supportsModuleEmission {
```
```suggestion
    if (writeModuleInterfaceHashAt != nil) && !outputType.supportsModuleEmission {
```

Sources/hc/CommandLine.swift
```diff
@@ -130,32 +187,71 @@ private typealias Module = FrontEnd.Module
...
+        use '--emit object' and link the objects yourself
+        """)
+    }
+    if imports.contains(moduleName) {
+      throw ValidationError("module '\(moduleName)' cannot import itself")
+    }
+    if noCaching && moduleCachePath != nil {
+      throw ValidationError("'--no-caching' and '--module-cache' are mutually exclusive")
+    }
+    if outputURL?.relativePath == "-" && !outputType.canBeWrittenToStandardOutput {
```
```suggestion
    if (outputURL?.relativePath == "-") && !outputType.canBeWrittenToStandardOutput {
```

Sources/hc/CommandLine.swift
```diff
@@ -130,32 +187,71 @@ private typealias Module = FrontEnd.Module
...
+    if !imports.isEmpty && outputType.linksDependencies {
+      throw ValidationError("""
+        '--import' is not yet supported with '--emit \(outputType.rawValue)'; \
+        use '--emit object' and link the objects yourself
+        """)
+    }
+    if imports.contains(moduleName) {
+      throw ValidationError("module '\(moduleName)' cannot import itself")
+    }
+    if noCaching && moduleCachePath != nil {
```
```suggestion
    if noCaching && (moduleCachePath != nil) {
```

Sources/hc/CommandLine.swift
```diff
@@ -130,32 +187,71 @@ private typealias Module = FrontEnd.Module
...
       }
 
+      for i in imports {
+        note("load imported module \(i)")
+        try driver.loadArchivedModule(.init(i))
+      }
+
       // Create a module for the product being compiled.
-      let product = productName(inputs)
+      let product = moduleName
```
kyouko-taiga: Not a fan. I want the product name to be `foo` if I compile `foo.hylo` as a single file.

Sources/hc/CommandLine.swift
```diff
@@ -185,15 +281,15 @@ private typealias Module = FrontEnd.Module
       await perform("normalization", for: module) {
         await driver.applyTransformationPasses(module)
       }
+
+      try emitInterfaceHashIfNeeded(of: module, from: driver)
+      try writeArchiveIfNeeded(of: module, from: driver)
```
kyouko-taiga: Why `write` and no `emit`?

Sources/hc/CommandLine.swift
```diff
@@ -342,21 +437,39 @@ private typealias Module = FrontEnd.Module
     }
   }
 
-  /// Sets up the value of search paths for locating libraries and cached artifacts.
-  private mutating func configureSearchPaths() throws {
-    let m = FileManager.default
-    if let cache = moduleCachePath {
-      librarySearchPaths.append(cache)
+  /// Ensures that the latest interface hash of the currently compiled module is emitted to disk
```
```suggestion
  /// Ensures that the latest interface hash of the currently compiled module is written to disk
```

Somehow "emitted to disk" sounds off.

Sources/hc/CommandLine.swift
```diff
@@ -342,21 +437,39 @@ private typealias Module = FrontEnd.Module
...
+  /// if requested.
+  ///
+  /// Only rewrites the file when the content actually changes, as some build tools only track
```
kyouko-taiga, line 451:
```suggestion
  /// Only rewrites the file when the content actually changes as some build tools only track
```

Sources/hc/CommandLine.swift
```diff
@@ -342,21 +437,39 @@ private typealias Module = FrontEnd.Module
...
-      let cache = m.temporaryDirectory.appending(path: ".hylocache")
-      try m.createDirectory(at: cache, withIntermediateDirectories: true)
-      note("module cache path: \(cache.path)")
-      librarySearchPaths.append(cache)
-      moduleCachePath = cache
+      note("interface hash unchanged, leaving \(u.path) untouched")
     }
+  }
+
+  /// Writes the archive of the currently compiled module to the requested location, if any, so
```
```suggestion
  /// Writes the archive of the module being compiled to the requested location, if any, so
```
Let's use this phrase consistently.

Sources/hc/CommandLine.swift
```diff
@@ -447,6 +549,35 @@ private typealias Module = FrontEnd.Module
     /// Executable binary.
     case binary = "binary"
 
+    /// `true` iff the invocation may emit a module archive or interface hash. 
+    /// IR lowering and mandatory transformations are required.
```
kyouko-taiga: The formatting is off or the sentence does not make sense. What is required and when? How does that relate to the value of this property?

Sources/hc/CommandLine.swift
```diff
@@ -447,6 +549,35 @@ private typealias Module = FrontEnd.Module
...
+    var supportsModuleEmission: Bool {
+      switch self {
+      case .ast, .typedAST, .rawIR:
+        return false
+      case .ir, .llvm, .asm, .object, .binary:
+        return true
+      }
+    }
+
+    /// `true` iff hc must link dependencies of the product into the output.
```
```suggestion
    /// `true` iff the compiler should link dependencies of the product into the output.
```

Tests/CommandLineTests/CommandLineEndToEndTests.swift
```diff
@@ -0,0 +1,271 @@
+import ArgumentParser
+import Foundation
+import FrontEnd
+import Utilities
+import XCTest
+
+/// End-to-end tests running the built `hc` executable as a subprocess.
```
kyouko-taiga:
```suggestion
/// End-to-end tests running `hc` as a subprocess.
```

Tests/CommandLineTests/CommandLineEndToEndTests.swift
```diff
@@ -0,0 +1,271 @@
...
+      a.append("--no-caching")
+    }
+    return try Process.execute(
+      Self.hcExecutable,
+      arguments: a,
+      workingDirectory: workingDirectory,
+      environment: environment.isEmpty
+        ? nil
+        : ProcessInfo.processInfo.environment
+          .merging(environment, uniquingKeysWith: { (_, new) in new }))
```
kyouko-taiga: Can that be better expressed with an `if` expression? If not, an abstraction is deserved.

kyouko-taiga (changes requested):
Impressive work!

Sources/Driver/Driver.swift
tothambrus11: Native libraries are things like `.dll`s or `.so` files (can be linked), while module search path is used to import Hylo libraries. I wanted to add a distinction because one might assume hylo libraries can also be linked (although with an intermediate step of compiling them first).

Sources/Driver/Driver.swift
tothambrus11: You can decide this with @dabrahams, as he suggested the opposite change. I applied this renaming to all functions except one where `m` was already reserved. I think `module` is often to read than `m`, because we don't have to have the contextual knowledge what `m`s usually refer to. I'm also fine with single-letter names if we have small scopes.

For the future: it would be great if we could spend less time on bikeshedding about local variable names, as these discussions have consumed far more attention than their relative importance to the success of the project.

Sources/Driver/Driver.swift
tothambrus11: It's not a precondition. The caller cannot uphold this condition for sure, just try. I think there should be a distinction about how we document these things. @dabrahams any thoughts?

Sources/hc/CommandLine.swift
tothambrus11: Yeah, I think I'll just remove `-I` for now.

For the record: `swiftc` and SPM mixes the two concerns. That can be good for CLI user experience but it's probably not worth the overhead of more search overall, and the extra possibility for confusion. Common CLI flags should behave the same as in clang/gcc.

Sources/hc/CommandLine.swift
tothambrus11: Why would you want to vary this at compile time of `hc`? As far as I see, if somebody is concerned with where caches are saved, they could just pass the flag explicitly. What cases do you see in which this is not intended?

We could also default this to `<user-cache>/hylo-distribution-version`, so that when the user has multiple compilers, they won't accidentally try to reuse the same cache.

Sources/hc/CommandLine.swift
tothambrus11: This is not debug information, it's a standard interface of the compiler to report something on stdout that the build system directly consumes.

I'm fine with either name, as long as we keep the CLI self-consistent. SPM is inconsistent in this regard, it has both `--show-bin-path` and `-print-target-info`. Whatever we choose, I'd like to use consistent naming for all these printing flags.

I think `print` is a little more concrete than `show`. Showing the standard library root sounds like it could just pop up an explorer or print some debug representation of the standard library root, e.g. its contents. We could still use `show` for genuenly debug-purpose flags (though I can't think of anything useful at the moment).

Do you still prefer `show` or should we keep `print`?

Sources/hc/CommandLine.swift
tothambrus11: Yeah, that's indeed suboptimal. But how would you name the module and binary name if multiple source files or a whole folder is given? Should we have this conditionally required?

Sources/hc/CommandLine.swift
tothambrus11: I meant to describe that lowering and mandatory transformations are required for module emission, so this property is true exactly then. I would rephrase it to:
```
    /// `true` iff the invocation may emit a module archive or interface hash, i.e. after mandatory transformations.
```

Sources/Driver/Driver.swift
kyouko-taiga: But you can link a Hylo library. I think the distinction is between modules and libraries, not between libraries and "native" libraries.

Sources/Driver/Driver.swift
kyouko-taiga: Following established convention is one way to avoid bike shedding. Presumably I spent some time bike shedding already, at least on my own, and I spent additional time going through the renaming in this patch wondering why the change (which is seemingly unrelated to the purpose of the patch) was necessary.

kyouko-taiga: There is no one size fits all solution to name variables. It is highly contextual, and mileage will vary depending on who you ask. I have my "system" and it produces outcomes that @dabrahams or anyone else is welcome to challenge.

*Changing* something places the burden of demonstrating why the change was necessary on the person making it. If you can't or don't want to carry this burden, don't make the change.

And no, I don't think spending time on variable names hurts the velocity of the project. Naming is hard, tedious, controversial, and incredibly important. What I think we should do in the future is refrain from suggesting renames that are unrelated to a PR's intent and/or propose that the renaming be done in another patch.

Sources/Driver/Driver.swift
tothambrus11: Yes, but we are not searching for Hylo libraries to link, we are searching for already compiled libraries (which I called native). I would rephrase it to `/// The native linker search paths.`

Sources/Driver/Driver.swift
kyouko-taiga: `module` is better than `m` here because it reads better in the docs and because `m` is used in the implementations of related and similarly documented functions throughout the file.

Sources/Driver/Driver.swift
kyouko-taiga: I think you mean to say the **system** linker.

My problem is with "native" and my point is that one could compile a Hylo program to a `.so` library and link it to another Hylo program. Hylo lets you write "native" programs and so the distinction is not helpful, if not misleading.

Sources/Driver/Driver.swift
tothambrus11: The alternatives I could think of looked more verbose:
```
    let h = try Result { try Module.headerAndDependencies(&a) }
                  .mapError(cannotParseArchive(of: module, at: source))
                  .get()
```
or
```
    let h: (name: Module.Name, fingerprint: UInt64, dependencies: [Module.Name])
    do { h = try Module.headerAndDependencies(&a) } 
    catch { throw cannotParseArchive(of: module, at: source) }
```

It would be nicer if we could use `do` as an expression. I would go with the last option, it's probably clearer what it does.

Sources/Driver/Driver.swift
tothambrus11: Yes, I see your point, native can be misleading if we think of Hylo programs directly as native libraries. I think both mental models are reasonable. (I thought of Hylo programs/libraries as separate abstract entities that we author, which we can compile down to a native library, or a WASM library, or interpret them.)

How about we just drop the qualifier altogether? `/// The linker library search paths.` or `/// The list of directories to pass to the linker as library search paths.`

Note: "system linker" may introduce an unintended distinction between mold/ld/lld/link.exe. lld can link native libraries on Windows while not being the system linker.

Sources/Driver/Driver.swift
tothambrus11: Removed the details from the docs and added a reference to https://github.com/hylo-lang/hylo-new/issues/321:

```
// TODO: Make the interface hash resilient to changes in function bodies.
// https://github.com/hylo-lang/hylo-new/issues/321
```

Sources/hc/CommandLine.swift
tothambrus11: Replaced with `The list of directories to pass to the linker as library search paths.`

Sources/hc/CommandLine.swift
tothambrus11: The two versions have slightly different meanings. According to https://www.magnumproofreading.com/post/blog-proper-punctuation-comma-before-as, we should use a comma when the meaning is `because`, and no comma when the meaning is `in the same way as`.

Sources/Driver/Driver.swift
dabrahams: That suggestion isn't grammatical though.  It parses (omitting the parenthesized part as “The (list (of (directories in imported module archives))) are searched .”  

There's an argreement problem with “the list are searched.”  The simplest way to make it grammatical would be “the lists are searched” but I don't think that makes sense or is what you mean.

```suggestion
  /// The directories searched for imported module archives (`.hylomodule` files), 
  /// in search order.
```

Sources/Driver/Driver.swift
dabrahams: I've seen the word “native” used liberally without a definition.  Is it a real distinction?  I think we almost always mean “binary.”

Sources/Driver/Driver.swift
dabrahams: I was going on the principle that putting type information in the name of a non-type almost always adds verbosity to no benefit, and doesn't serve the documentation better than a name describing a role or a single letter, and the guidelines say to choose parameter names to serve the documentation.  However, if the project has some other convention so be it; we can't change course now.

Sources/Driver/Driver.swift
tothambrus11: To me, searching "in search order" sounds like a recursive description. What about "in order" or leaving it implicit?

My current version is: ``  /// The directories in which imported module archives (`.hylomodule` files) are searched.``

Sources/Driver/Driver.swift
tothambrus11: Hylo archives are also in a binary representation.

In my experience, "native libraries" always used to refer to something like a shared library or static library, and definitiely not a Java *.jar library. Sometimes we refer to the library as a whole (with its source code, headers, etc) as a native library, which may be confusing in this case. Linkers are generally understood, so I removed the native terminology altogether.

Sources/hc/CommandLine.swift
tothambrus11: This is implemented by https://github.com/hylo-lang/hylo-new/pull/397/changes/9d6b2ab608c4c761a524ee7520668c26ec999fc7

We either use the given module name, the unique source file's base name, or Main.

Sources/hc/CommandLine.swift
tothambrus11: See: https://github.com/hylo-lang/hylo-new/pull/397/changes/9d6b2ab608c4c761a524ee7520668c26ec999fc7

Sources/hc/CommandLine.swift
dabrahams: That's not right either. A search path is a list of directories

Sources/hc/CommandLine.swift
dabrahams: I'm for print for the reasons you suggest: show could mean lots of things, including that something will not be hidden in some other output. It's the opposite of hide

Sources/hc/CommandLine.swift
tothambrus11: I think these are independent. Linkers define what their search path is, and we pass a list of directories to them. I don't think people will have trouble understanding what this means.

dabrahams: Why not just get the language right?

It's the “The linker's library search path.“ Nobody needs to be told it's a list, and so far in this project `URL` is just a proxy for directory.

Sources/hc/CommandLine.swift
kyouko-taiga: TL;DR: I want to be able to install `hc` so that it writes its cache in `~/koala/` *and* then be able to write `hc hello.hylo`.

It may not be possible to use the default, or the user may prefer to change the default when they install `hc` for whatever reason they choose. In general I strongly believe that one should have complete control over the locations of programs installed on their system, because applications cannot predict all possible configurations. For example, I find it infuriating when an application defaults to the locations where homebrew writes its stuff.

If it is possible to change the default location of Hylo cache, then it must be possible to use `hc` without specifying that cache location *again*.

tothambrus11: Ah yes, then I would use an environment variable for that rather than recompile `hc` to suit the user's need.

Sources/hc/CommandLine.swift
kyouko-taiga: Fair enough. I'm used to "show" being used to print stuff on my terminal. e.g.
- `git remote show origin`
-  `curl --show-headers google.com`

I will not die on this hill.

Sources/hc/CommandLine.swift
tothambrus11: I removed the explicit "list of" mentioning, as it's indeed redundant. I'd like to explicitly state the distinction between URLs that represent folders and files, and not rely on a project-level pattern that happened to be the case so far.

tothambrus11: The latest version is: `  /// The directories to pass to the linker as library search paths.`

dabrahams: Still wrong.  Why not use the right words that I specified above?

Sources/hc/CommandLine.swift
dabrahams: Ambrus is right here.

Sources/hc/CommandLine.swift
tothambrus11: "The linker's library search path" - it's clearly not a single path. We can argue that it's not confusing too much if we omit the plural, but I wouldn't say my version is "wrong" and your version is "right". I think my version is good enough, though may not be perfect in some sense.

dabrahams: You're getting confused by the ways “path” can be used.  But a “[search path](https://en.wikipedia.org/wiki/PATH_(variable)#History)“ is always a set of directories with some implied order of search. My version is right because it implies that the order of search is the order of the list; your version is wrong because it doesn't.  It also implies that each directory is a search path because you end with the word “paths.” A “list of directories” is not a search path unless you say it is.

tothambrus11: Thanks, I didn't know that this is such an established term. I could interpret this as the path along which we walk to look for libraries. I'm changing to your description and replace the variable name with the singular form.

Sources/hc/CommandLine.swift
tothambrus11: Done in https://github.com/hylo-lang/hylo-new/pull/397/changes/04f421245b48cdadfe2d7c6f278372f545609d70

Sources/Driver/Driver.swift
dabrahams: We must document the behavior because the client should know what's required for success.
Either:

1. Have the method create any missing directories and say that's what the method does. No need to document the other failure modes; they are implied by the fact that it's disk operations.
2. Don't change the functionality of the method, but say “Throws if the parent directory of `output` does not exist.”

The meaning of “expected to” is vague and should never be used in contract documentation.

Sources/FrontEnd/Module.swift
dabrahams: Why not have archivist try to downcast to known collections that aren't repeatably ordered and trap if it succeeds?  It's not perfect insurance but it could help.  You can get there by creating a marker protocol and making `Dictionary` and `Set` conform.

Sources/Driver/Driver.swift
tothambrus11: The vagueness was sort of intentional: I didn't want to promise that it will throw if the parent directory doesn't exist. Maybe it tries creating it, it will be just less efficient than if we immediately tried to save the document in an existing directory. All I want to imply is that the caller should its best to satisfy this condition, but it's not the end of the world if it doesn't hold. We can argue if this is a good idea, but at least that was my original intention.

Sources/Driver/Driver.swift
dabrahams: It may sound recursive to you but it isn't. You should make the change I suggested. The wording you gave is incomplete. Losing precision in the name of avoiding awkward language is not a win.

Sources/Driver/Driver.swift
dabrahams: Then you should choose option 1. But choose one of those two please.

Sources/Driver/Driver.swift
tothambrus11: If there is a specific folklore term "search order" that means specifically the linear search, then I was not aware of it, and I would rather say it explicitly that it's a linear search. 

Otherwise, the act of searching is already implied by "are searched", so the phrasing would give redundant information.

dabrahams: ??? “Search path” implies linear search; everybody knows that earlier elements take priority.

I suppose technically you could do it in parallel, but the semantics remain the same as a linear search.

Sources/FrontEnd/Module.swift
kyouko-taiga: Archivist is not opinionated about ordered collections. We could add an opt-out in the API but I since this design is not fail proof I am hesitant.

dabrahams: Yeah, I can understand that.

Actually I think a better solution is to create our own `Dictionary` and `Set` that do not conform to `Sequence`s, and then try to avoid using `[_:_]` style type ascriptions.

Sources/Driver/Driver.swift
```diff
@@ -34,10 +34,13 @@ public struct Driver {
   /// The code model for code generation.
   public var codeModel: CodeModel
 
-  /// The list of directories in which to search for libraries to link.
-  public var librarySearchPaths: [URL]
+  /// The linker's library search path.
+  public var librarySearchPath: [URL]
 
-  /// The names of the native libraries to link.
+  /// The directories in which imported module archives (`.hylomodule` files) are searched.
```
dabrahams, line 40:
```suggestion
  /// Where imported module archives (`.hylomodule` files) are found, in search order.
```

```suggestion
  /// The search path for imported module archives (`.hylomodule` files).
```

PR #407, opened 2026-08-04
Fix isExposed for extensions

### LINKED ISSUE #404: Extension members are seen as !isExposed even if the extendee type is exposed
At 79ffed4 if you try to work around #403 by making  `__hylo_offset_of_member` `@inline(always)` the first diagnostic incorrectly claims I am using a non-exposed function:

```
/Users/dave/src/hylo-new/Tests/CompilerTests/CompilerTests.swift:286: error: -[CompilerTests.CompilerTests test_positive_runtime_offset_of_member] : failed - 
StandardLibrary/Sources/Core/Runtime/OffsetOfMember.hylo:552.30-36: error: use of non-exposed function in inlined function
  sink let p = memberArray + Int(n)
                             ~~~~~~
```

tothambrus11:
This is a pretty ugly solution. We might want to make `ExtensionDeclaration` a `ModifiableDeclaration` (not necessarily allowing writing modifiers manually, just to have the capability to query its exposedness by forwarding that call to its extendee). It may be also useful to create a ModifiableDeclarationIdentity, as now I was forced to pass a type-erased ID instead of something that requires the type to conform to a protocol.

Fixes #404 

It might be good to rewrite this before merging, but I'm not sure how.

### DIFF
--- a/Sources/FrontEnd/Program.swift
+++ b/Sources/FrontEnd/Program.swift
@@ -470,19 +470,81 @@ public struct Program: Sendable {
   /// Returns `true` if the contents of `d` is visible in all modules.
   ///
   /// The result is `true` if `d` is annotated with `@exposed` and/or `d` and all the scopes
-  /// enclosing it are public.
+  /// enclosing it are public. An extension has no modifiers; its scope is public iff the
+  /// declaration of its extendee, if any, is exposed.
   public func isExposed<T: ModifiableDeclaration>(_ d: T.ID) -> Bool {
+    isExposed(DeclarationIdentity(d))
+  }
+
+  private func parentModifiableDeclaration<T: SyntaxIdentity>(of n: T) -> AnySyntaxIdentity? {
+    guard let s = parent(containing: n).node else { return nil }
+
+    if self[s] is (any ModifiableDeclaration) { 
+      return s.erased
+    }
+    if let e = cast(s, to: ExtensionDeclaration.self),
+      let x = nominalExtendee(of: e)
+    {
+      return self[x] is (any ModifiableDeclaration) ? x.erased : nil
+    }
+    return nil
+  }
+
+  /// Returns `true` if the contents of `d` is visible in all modules.
+  ///
+  /// - Requires: The module containing `d` is typed.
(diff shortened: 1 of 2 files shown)


kyouko-taiga:
> This is a pretty ugly solution. We might want to make ExtensionDeclaration a ModifiableDeclaration (not necessarily allowing writing modifiers manually, just to have the capability to query its exposedness by forwarding that call to its extendee). It may be also useful to create a ModifiableDeclarationIdentity, as now I was forced to pass a type-erased ID instead of something that requires the type to conform to a protocol.

So ... I would very much like to avoid having to write `public extension T {}` and that is why I didn't make extensions conform to `ModifiableDeclaration`. I think extensions should be simply transparent w.r.t. exposure.

Note that it does not matter to make an extension "private" if it extends a type that can't be named outside of the scope considered, since it won't be possible to (statically knowingly) have a term of that type anyway.

Sources/FrontEnd/Program.swift
```diff
@@ -470,19 +470,81 @@ public struct Program: Sendable {
   /// Returns `true` if the contents of `d` is visible in all modules.
   ///
   /// The result is `true` if `d` is annotated with `@exposed` and/or `d` and all the scopes
-  /// enclosing it are public.
+  /// enclosing it are public. An extension has no modifiers; its scope is public iff the
+  /// declaration of its extendee, if any, is exposed.
```
kyouko-taiga, line 474: That is a bad idea to tie the exposure of an extension with the exposure of the declaration of the extendee because many extendees have no declaration. Also it makes things tricky if we consider aliases and type parameters. What is the exposure of `S<T>` if `S` is public but `T` isn't? What if `T` is a private typealias for a public type `U`?

I think the "Swift" model of extension is a bad one to think about Hylo's. Extensions are not really attached to types and their declarations. They merely introduce new functions whose first argument can be left of a dot. I don't think the extendee should have any other role in describing the properties of an extension scope.

So, I think a simpler fix is just to say that extensions are transparent with respect to exposure.

kyouko-taiga (changes requested):
See inline comments.

Issue #333, opened 2026-07-21
Some unreachable blocks may not be removed when inlining is involved


tothambrus11:
Removing the unreachable blocks currently runs before inlining. We may inline a function that has a `Builtin.trap()` (returning `Never`) that trips up the invariant that the program doesn't contain unreachable blocks after the mandatory passes. This trips up the serialization that assumes this.

```swift
    // Write the number of basic blocks in the function. Note that the function cannot contain any
    // unreachable block at this point.
    archive.write(unsignedLEB128: blocks.count)

   ...

    assert(blocks.count == table.blocks.count) // assertion fails
```

Reproducer:

```hylo
@inline(always)
public fun crash() {
  Builtin.trap()
}

public fun main() {
  var x = true
  crash()
  &x = false
}
```

kyouko-taiga:
I think the simplest fix is to run `removeCodeAfterNeverReturningCalls` and `removeUnreachableBlocks` after inlining.

tothambrus11:
It could be a nice solution for now.

For the future: what are the tradeoffs? I would guess removing code before inlining could reduce compile times. On the other hand, inlining may reveal more optimization opportunities where we could remove unreachable blocks before LLVM lowering.

kyouko-taiga:
There is no shame running some passes more than once in a standard optimization pipeline. DCE is a massive speedup for many passes and so it ought to come early. It is not an issue to run it after inlining a second time.

PR #408, opened 2026-08-04
Implement recursive inlining

kyouko-taiga:
This patch fixes the hack necessary to inline functions recursively, as well as the bug observed in #333.


Sources/FrontEnd/Program.swift
```diff
@@ -1557,28 +1555,26 @@ public struct Program: Sendable {
...
     // Is `f` defined in `m`?
     if self[m].ir[i].isDefined {
-      return (m, self[m].ir[i])
+      return (m, i)
     }
 
     // Is `f` defined in a dependency?
     for d in self[m].dependencies {
       guard let n = self.identity(module: d) else { continue }
       if let j = self[n].ir.identity(function: f), self[n].ir[j].isDefined {
```
tothambrus11, line 1586:
```suggestion
      if let n = self.identity(module: d),
        let j = self[n].ir.identity(function: f), self[n].ir[j].isDefined
      {
```

Sources/FrontEnd/IR/IRFunction.swift
```diff
@@ -954,6 +960,7 @@ extension IRFunction: Archivable {
     self.output = try archive.read(Output.self, in: &context)
     self.typeParameters = try archive.read([GenericParameter.ID].self, in: &context)
     self.termParameters = try archive.read([IRParameter].self, in: &context)
+    self.passedMandatoryInlining = try archive.read(Bool.self)
```
tothambrus11, line 1051:
```suggestion
    self.passedMandatoryInlining = true
```
Do we allow archiving modules that haven't passed all mandatory transformation passes without errors? In #397 I gated archiving to be only possible after IR transformations. Otherwise we might want to also save in the archive what stage of compilation it's in, so subsequent processing can complete the missing steps.

Sources/FrontEnd/IR/Transforms/IRFunction+Inlining.swift
```diff
@@ -22,10 +37,15 @@ extension IRFunction {
   }
 
   /// Inlines the contents of the callees in `self` that were resolved statically to a declaration
-  /// annotated with `@inline(always)`, and whose definition is visible from `m`.
-  internal mutating func inlineSimpleCallees(emittingInto m: Module.ID, using typer: inout Typer) {
-    var work = Array(blocks)
-    while let b = work.popLast() {
+  /// that should be inlined given `context`.
```
tothambrus11, line 54: It could be good to describe here how `context` determines whether to inline something.

Sources/FrontEnd/IR/Transforms/IRFunction+Inlining.swift
```diff
@@ -22,10 +37,15 @@ extension IRFunction {
...
+  internal mutating func inlineSimpleCallees(
+    emittingInto m: Module.ID, using typer: inout Typer,
+    in context: inout InliningContext
+  ) {
+    // Nothing to do if the function already went through mandatory inlining.
+    if passedMandatoryInlining { return }
```
tothambrus11, line 60: I might have missed it, but I didn't see any place where this was set to true.

Sources/FrontEnd/Program.swift
kyouko-taiga: I don't like this suggestion. The two lines have different purposes. The first is about finding a module, the second is about finding a function.

I also think newlines in conditions are harmful to legibility.

Sources/FrontEnd/IR/Transforms/IRFunction+Inlining.swift
kyouko-taiga: I deliberately kept that abstract here because I plan on `context` to hold much more data in the future, including information necessary to compute heuristics. At that point the algorithm will be declared elsewhere.

Sources/FrontEnd/IR/IRFunction.swift
kyouko-taiga: I think we should only archive modules that have gone through mandatory transformations, but we should probably create another PR to encode what passes of the pipeline a function went through. This Boolean is just a proto-version of that setup.

tothambrus11: Isn't it enough to save this at a module level? Who would use this information after opening the archive?

Sources/FrontEnd/IR/Transforms/IRFunction+Inlining.swift
kyouko-taiga: Good catch, I must have missed one commit during my rebase.

Sources/FrontEnd/IR/IRFunction.swift
kyouko-taiga: We need a per-function bit about inlining at least.

Tests/CompilerTests/negative/recursive-inlining.diagnostics.expected
```diff
@@ -0,0 +1,6 @@
+negative/recursive-inlining.hylo:2.11-14: error: cannot inline recursive function
+fun f() { f() }
+          ~~~
+negative/recursive-inlining.hylo:8.11-14: error: cannot inline recursive function
+fun h() { g() }
+          ~~~
```
tothambrus11: Can we report all sites of errors? Both the body of `g` and `h` is wrong, but we only diagnose one of them arbitrarily. I would like to see everywhere locally if my function is wrong. Maybe I'm in the middle of implementing `g()` and I mistakenly called `h` instead of `f`. I want to see that error locally rather than only in `h`. Maybe I also have other problems at the same time, and I only want to look at issues elsewhere once I fixed the code I'm working on.

Sources/FrontEnd/IR/Transforms/IRFunction+Inlining.swift
```diff
@@ -37,38 +57,83 @@ extension IRFunction {
...
+
+    // Update the inlining context.
+    context.stack.append(callee)
+    defer { context.stack.removeLast() }
+
+    // Locate the definition of `f`.
+    guard let (n, f) = typer.program.definition(of: callee, visibleFrom: m) else {
+      // Either the callee is not defined, in which case an error has been reported elsewhere, or
+      // it is being inlined, in which case we must complain about recursive inlining.
+      if context.stack.contains(callee) {
```
tothambrus11, line 116: We just pushed `callee` to the stack above, so this will be vacuously true.

Tests/CompilerTests/negative/recursive-inlining.diagnostics.expected
kyouko-taiga: That'd be nice but quite difficult to do. Not in scope for this PR.

kyouko-taiga: I improved the diagnostic a bit but reporting all cases of recursion requires the construction of a call graph.

At least we know that diagnostics will only be reported in a function marked `@inline` so there is no risk of getting an error in an ordinary function just because it happens to be called by a `@inline` function.

tothambrus11:
I added some more positive test cases in https://github.com/hylo-lang/hylo-new/pull/427. I think we can merge now.

PR #409, opened 2026-08-04
Fix the implementation of 'Program.isExposed(_:)' in the presence of extensions

### LINKED ISSUE #404: Extension members are seen as !isExposed even if the extendee type is exposed

kyouko-taiga:
An alternative to #407 to fix #404 (`Program.isExposed(_:)` involving extensions).

### DIFF
--- a/Sources/FrontEnd/Program.swift
+++ b/Sources/FrontEnd/Program.swift
@@ -469,8 +469,8 @@ public struct Program: Sendable {
 
   /// Returns `true` if the contents of `d` is visible in all modules.
   ///
-  /// The result is `true` if `d` is annotated with `@exposed` and/or `d` and all the scopes
-  /// enclosing it are public.
+  /// The result is `true` if `d` is annotated with `@exposed` or `d` and all the scopes enclosing
+  /// it are public. Note that the scope of an extension is always public.
   public func isExposed<T: ModifiableDeclaration>(_ d: T.ID) -> Bool {
     // Is `d` explicitly exposed?
     if (annotation("exposed", appliedTo: d) != nil) { return true }
@@ -479,8 +479,13 @@ public struct Program: Sendable {
     if !self[d].is(.public) { return false }
     var p = parent(containing: d)
     while let a = p.node {
-      guard let b = self[a] as? (any ModifiableDeclaration), b.is(.public) else { return false }
-      p = parent(containing: a)
+      if tag(of: a) == ExtensionDeclaration.self {
+        p = parent(containing: a)
+      } else if let b = self[a] as? (any ModifiableDeclaration), b.is(.public) {
+        p = parent(containing: a)
+      } else {
+        return false
+      }
     }
     return p.isFile
   }
--- a/Tests/CompilerTests/positive/inlining-with-extensions.hylo
+++ b/Tests/CompilerTests/positive/inlining-with-extensions.hylo
@@ -1,0 +1,17 @@
+//! exit-status:1
+
+extension Int32 {
+
+  @inline(always)
+  public fun incremented() -> Int32 { self + (1 as Int32) }
+
+}
+
+@inline(always)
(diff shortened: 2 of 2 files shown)


tothambrus11 (approved):

dabrahams:
@kyouko-taiga This is much uglier than the fix in my branch (4d07d5e5c2b6cb20ce39614a0d0421f60c7a7f7c). Is there a reason it needs to be?

kyouko-taiga:
"much" is a bit strong I think.

In any case, your version is not implementing the same logic.
Mine says:
- if it's an extension, go to parent scope; otherwise
- if it's a public modifiable declaration, go to parent scope; otherwise
- exit with `false`.

Yours says:
- if it's a non-public modifiable declaration, exit with `false`; otherwise 
- go to parent scope

IIUC, in the following program, my version returns `false` for the function in the body of the conditional while yours returns `true`.

```
public fun main() {
  if true { public fun is_this_exposed() {} }
}
```

The correct result is `false`.

tothambrus11:
Should the type-checker already reject programs where an inner declaration tries to be more public than its enclosing scope?

kyouko-taiga:
> Should the type-checker already reject programs where an inner declaration tries to be more public than its enclosing scope?

Don't think it's necessary but I also don't have strong arguments against it.

dabrahams:
But isnt rhat the correct logic?presumably a non-modifiable declaration could have no effect on access control.

kyouko-taiga:
We're talking about exposure, not access control. If a declaration is declared in a local scope, it is not exposed. Otherwise we're making local functions and lambdas part of ABIs, which I think is counter-intuitive and probably useless since we can't name local functions outside of their scopes.

dabrahams:
> Should the type-checker already reject programs where an inner declaration tries to be more public than its enclosing scope?

No. Should be possible for a private struct to have members declared public as a way of marking its encapsulation boundary.

kyouko-taiga:
> No. Should be possible for a private struct to have members declared public as a way of marking its encapsulation boundary.

That was my first reaction but when I tried to forge an example I realized that if our lowest level is private to the file, then there is no struct that you can write that would require its members to be more public than the struct.

dabrahams:
>  If a declaration is declared in a local scope, it is not exposed. Otherwise we're making local functions and lambdas part of ABIs

I see. Thanks for explaining.

Regarding the access control on struct members, I realize it has no impact on what you can do with those members, but how else am I going to mark the intended API boundary?  Also valuable to be able to make a private struct public without re-evaluating access on all its members.

Sources/FrontEnd/Program.swift
```diff
@@ -479,8 +479,13 @@ public struct Program: Sendable {
     if !self[d].is(.public) { return false }
     var p = parent(containing: d)
     while let a = p.node {
-      guard let b = self[a] as? (any ModifiableDeclaration), b.is(.public) else { return false }
-      p = parent(containing: a)
+      if tag(of: a) == ExtensionDeclaration.self {
+        p = parent(containing: a)
+      } else if let b = self[a] as? (any ModifiableDeclaration), b.is(.public) {
+        p = parent(containing: a)
```
dabrahams, line 485: Consider lifting this line and its duplicate out to the level of the loop, after the condition.

PR #415, opened 2026-08-07
Fix the lowering of mutating calls

### LINKED ISSUE #310: inout subscript bundle chain fails to pass access checking
```
subscript s(x: auto Int) auto -> Int {
  let {

    yield 0//x.s0[]
  }
  inout {
    yield &0
  }
}

subscript t(x: auto Int) auto -> Int {
  let {
    yield s[x]
  }
  inout {
    inout w = s[&x] // error: illegal 'inout' access
    let z = w + 1 // remove this and it compiles
    yield &0
  }
}
```

kyouko-taiga:
This patch fixes #310


tothambrus11:
#310 also had a similar reproducer which doesn't pass access checking yet:

```
struct Q {

  subscript s0() inout -> Int {
    var x = 0
    yield &x
  }

  subscript s1() auto -> Int {
    let {
      yield self.s0[]
    }
    inout {
      yield &self.s0[] // error: illegal 'inout' access
    }
  }

}
```
Does it make sense to push this already, or should we address all the similar known issues in this PR?

Tests/CompilerTests/positive/bundle-application-1.hylo
```diff
@@ -0,0 +1,14 @@
+//! stage:lowering
```
tothambrus11, line 1: We should add the chained subscript application test cases from #310

kyouko-taiga: I think this other test case has been solved elsewhere. I can no longer reproduce on `main` and I remember having worked on it.

tothambrus11: I reproduce the problem with this:

```
struct Q {

  subscript s0() inout -> Int {
    var x = 0
    yield &x
  }

  subscript s1() auto -> Int {
    let {
      yield self.s0[]
    }
    inout {
      yield &self.s0[] // error: illegal 'inout' access
    }
  }

}

```

kyouko-taiga: Are you sure? This program is illegal because of the call to `s0` in the `let` variant.

kyouko-taiga: If you fix that legitimate error the program compiles.

PR #416, opened 2026-08-07
Implement lowering for uses of the 'Self' alias

### LINKED ISSUE #308: Using `Self` as a value: FrontEnd/IREmitter.swift:1396: Fatal error: Lowering not implemented for $<builtin alias>
Actually that message only came up  because I added some diagnostic output to the `fatalError`.  Reproduce building the standard library with the `lvalue-lowering-fatalError` branch.  

As noted you need to build the stdlib so there's no single file involved.  The last commit on that branch if it helps is https://github.com/hylo-lang/hylo-new/commit/c183f9b8c185747b166c25f75c7c5ac277171d82

(Please delete the branch when this is fixed).

kyouko-taiga:
This patch fixes #308

### DIFF
--- a/Sources/FrontEnd/Syntax/Builtins/BuiltinEntity.swift
+++ b/Sources/FrontEnd/Syntax/Builtins/BuiltinEntity.swift
@@ -1,6 +1,9 @@
 import Archivist
 
 /// An entity built in the compiler.
+///
+/// Such entities may be referred to from source (e.g. `Builtin.add_i8(_:_:)`) but don't have a
+/// corresponding declaration.
 @Archivable
 public enum BuiltinEntity: Hashable, Sendable {
 
@@ -25,7 +28,7 @@ public enum BuiltinEntity: Hashable, Sendable {
   /// A built-in type.
   case type
 
-  /// A built-in entity.
+  /// A built-in function.
   case function(BuiltinFunction)
 
 }
--- a/Sources/FrontEnd/Syntax/Builtins/BuiltinEntity.swift
+++ b/Sources/FrontEnd/Syntax/Builtins/BuiltinEntity.swift
@@ -4,11 +4,26 @@ import Archivist
 @Archivable
 public enum BuiltinEntity: Hashable, Sendable {
 
-  /// A built-in type alias.
-  case alias
+  /// The `Self` alias.
+  case selfAlias
+
+  /// The `Metatype` alias.
+  case metatypeAlias
+
+  /// The `Never` alias.
+  case neverAlias
+
+  /// The `Void` alias.
+  case voidAlias
+
+  /// The built-in module.
+  case module
 
   /// The witness of a coercion.
   case coercion
+
+  /// A built-in type.
+  case type
 
   /// A built-in entity.
   case function(BuiltinFunction)
--- a/Sources/FrontEnd/IR/IREmitter.swift
+++ b/Sources/FrontEnd/IR/IREmitter.swift
@@ -2477,16 +2477,21 @@ internal struct IREmitter {
     _return()
   }
 
-  /// Generates the IR for storing the type witness expressed by `e` into a temporary alloca and
-  /// returns that alloca.
+  /// Generates IR for storing the type witness expressed by `e` into a fresh alloca and returns
+  /// that alloca.
   ///
   /// - Requires: The evaluation of `e` has no side effects.
   private mutating func _emitTypeWitness(expressedBy e: ExpressionIdentity) -> IRValue {
     let t = program.type(assignedTo: e, assuming: Metatype.self)
-    let u = program.types.dealiased(program.types[t].inhabitant)
-    let v = program.types.demand(TypeWitness())
-    let x = _alloca(v.erased)
-    _emitInitialize(x, with: .type(u, v))
+    return _emitTypeWitness(of: program.types[t].inhabitant)
+  }
+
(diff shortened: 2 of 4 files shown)


Sources/FrontEnd/Syntax/Builtins/BuiltinEntity.swift
```diff
@@ -4,12 +4,27 @@ import Archivist
 @Archivable
 public enum BuiltinEntity: Hashable, Sendable {
 
-  /// A built-in type alias.
-  case alias
+  /// The `Self` alias.
+  case selfAlias
```
tothambrus11, line 11: Could you add some more context in the description of this enum? The current docs don't tell much more than what the names already imply, so without looking at its usages, I'm not sure what the role of this type is.

kyouko-taiga: What's not clear about the case describing the "`Self` alias" of "a entity built in the compiler"?

Sources/FrontEnd/Syntax/Builtins/BuiltinEntity.swift
```diff
@@ -4,12 +4,27 @@ import Archivist
...
+  /// The built-in module.
+  case module
 
   /// The witness of a coercion.
   case coercion
 
+  /// A built-in type.
+  case type
+
   /// A built-in entity.
```
tothambrus11:
```suggestion
  /// A built-in function.
```

Sources/FrontEnd/Syntax/Builtins/BuiltinEntity.swift
tothambrus11: I think "entity" was not clear enough. I didn't know what kind of thing an entity can be. Upon reading its usages, I saw it's used for referring to entities from source that are built in and don't have a declaration. I think adding this would be helpful:

```
/// An entity built in the compiler.
///
/// Such entities may be referred to from source (e.g. `Builtin.add_i8(_:_:)`) but don't 
/// have a corresponding declaration.
```

Sources/FrontEnd/Syntax/Builtins/BuiltinEntity.swift
```diff
@@ -4,12 +4,27 @@ import Archivist
...
+  case metatypeAlias
+
+  /// The `Never` alias.
+  case neverAlias
+
+  /// The `Void` alias.
+  case voidAlias
+
+  /// The built-in module.
+  case module
```
tothambrus11, line 23:
```suggestion
  case builtinModule
```
... or `builtinNamespace`.

This makes it clear that it's not just _some_ module, but the Builtin module.  It reads consistently with `*Alias`, which are composed of a name + their type.

Sources/FrontEnd/Syntax/Builtins/BuiltinEntity.swift
```diff
@@ -4,12 +4,27 @@ import Archivist
...
 
   /// The witness of a coercion.
   case coercion
 
+  /// A built-in type.
+  case type
```
tothambrus11, line 29: How do we know which built-in type and coercion we are referring to? What makes `builtinFunction` require a selection while these 2 can look seemingly abstract?

If their concrete payload is implied from some other source (e.g. the type paired with this DeclarationReference), we could add a note here mentioning that `/// The actual type of this entity can be acquired from the corresponding `DeclarationReference`'s `type` property.

tothambrus11:
I left some comments about docs. Otherwise LGTM.

Sources/FrontEnd/Syntax/Builtins/BuiltinEntity.swift
kyouko-taiga: Happy to add the sentence but please note that some vocabulary is part a general lexicon that I hope we don't have to redefine in sources all the time. An entity is a concept defined in the specification (even if it's partially obsolete).

Sources/FrontEnd/Syntax/Builtins/BuiltinEntity.swift
kyouko-taiga: I disagree. At use sites it reads `.builtin(.module)`

Sources/FrontEnd/Syntax/Builtins/BuiltinEntity.swift
kyouko-taiga: A built-in entity is only ever used to construct a declaration reference, which already has a type.

I don't think it is worth documenting all of these small details. Documentation should not specify how the value will be used when it becomes part of some other value.

Sources/FrontEnd/Syntax/Builtins/BuiltinEntity.swift
tothambrus11: If all use sites will be wrapped in a `.builtin(...)`, then it's perfect.

Sources/FrontEnd/Syntax/Builtins/BuiltinEntity.swift
tothambrus11: > I don't think it is worth documenting all of these small details. Documentation should not specify how the value will be used when it becomes part of some other value.

I think the motivation behind this is usually to aid modularity, less coupling. However, it is already very much coupled implicitly, and such coupling are hard to understand without an explanation, e.g. on the type. New readers won't know that it's actually only used in a particular way.

tothambrus11 (approved):

PR #425, opened 2026-08-08
Eliminate dead code from the emitter

kyouko-taiga:


Sources/FrontEnd/IR/IREmitter.swift
```diff
@@ -685,10 +685,6 @@ internal struct IREmitter {
       lower(store: program.castUnchecked(e, to: If.self), to: target)
     case InoutExpression.self:
       lower(store: program.castUnchecked(e, to: InoutExpression.self), to: target)
-    case IntegerLiteral.self:
-      lower(store: program.castUnchecked(e, to: IntegerLiteral.self), to: target)
-    case FloatingPointLiteral.self:
-      lower(store: program.castUnchecked(e, to: FloatingPointLiteral.self), to: target)
```
tothambrus11, line 691: It would be great if we could tell from cases that are simply unimplemented from cases that are intended to be unexpected. I would list all tags that we know are not supposed to end up here as explicitly `unexpected`, and leave the default for `unimplemented`, whose set may implicitly change over time as we add more expression types and forget to update switches.

kyouko-taiga: It is a distinction without a difference. They will end up as an issue saying that some assertion occurred with a backtrace anyway.

tothambrus11: I've got bitten by this exact kind of problem multiple times. I tried diagnosing a crash, and I was not sure if it's due to a broken invariant (bug) or an unimplemented feature.  Declaring our intent would be very helpful.

PR #427, opened 2026-08-08
Add positive test cases to recursive inlining

tothambrus11:
Adds some more tests to inlining to cover some more tricky cases.

### DIFF
--- a/Tests/CompilerTests/positive/inlining-broken-cycle.hylo
+++ b/Tests/CompilerTests/positive/inlining-broken-cycle.hylo
@@ -1,0 +1,9 @@
+//! no-std stage:lowering
+
+// A call cycle that leaves the set of inlined functions is not a recursive inlining.
+@inline(always)
+fun f() { g() }
+
+fun g() { f() }
+
+public fun main() { f() }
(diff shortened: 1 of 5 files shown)


Tests/CompilerTests/positive/inlining-broken-cycle.hylo
```diff
@@ -0,0 +1,10 @@
+//! no-std stage:lowering
+
+// A call cycle that leaves the set of inlined functions is not a recursive inlining.
+@inline(always)
+fun f() { g() }
+
+@inline(never)
```
kyouko-taiga: `@inline(never)` is a very strong requirement. It won't run any other code at the moment but I think the test would be stronger with no annotation at all. That way, heuristic inlining will have to figure out that it can't inline `g` in `f`.

kyouko-taiga (approved):
Very nice, thanks!

Tests/CompilerTests/positive/inlining-broken-cycle.hylo
tothambrus11:
```suggestion
```

PR #434, opened 2026-08-08
Interpreter remove unused properties

RishabhRD:
Interpreter code brings lines (for completeness purpose) which are not currently being used as they will be useful in later phase. But that is leading to breaking coverage of main. Thus the PR removes those lines and will add them when will really be used and thus tested.


dabrahams (approved):
You should leave the `CustomStringConvertible` conformances and `description`, because they don't impact correctness and are just too useful for debugging.  Otherwise, LGTM

PR #441, opened 2026-08-09
Interpreter: roundUp(toNearestMultipleOf:) to have UnsignedInteger as domain

RishabhRD:


Sources/Interpreter/TypeLayout.swift
```diff
@@ -61,12 +61,12 @@ struct TypeLayout: Regular {
...
 }
 
-extension BinaryInteger {
+extension UnsignedInteger {
 
   /// Returns `self` rounded up to the nearest multiple of `n`.
   ///
   /// - Precondition: `n > 0`.
-  fileprivate func roundedUp(toNearestMultipleOf n: Self) -> Self {
+  internal func roundedUp(toNearestMultipleOf n: Self) -> Self {
```
tothambrus11, line 69:
```suggestion
  internal func roundedUp(toAMultipleOf n: Self) -> Self {
```

tothambrus11:
What's the motivation for narrowing the domain? Was it incorrect/less efficient before? I think we should keep algorithms as generic as possible.

It seems now we also need to convert between Int/UInt, which looks like a small regression.

RishabhRD:
> What's the motivation for narrowing the domain? 

This comes from [this comment](https://github.com/hylo-lang/hylo-new/pull/318#discussion_r3739690932) of yours. The reason being negative `self` is problematic as any `negative_number % n` is negative, that is not expected. Either I need to complicate the contract to explain the postcondition, or I can narrow the domain for simpler contracts. 

Note: this extension is there to support generic case when `n` is not a power of 2. If `n` is power of 2, there is existing way to implement this operation without loss of efficiency.

Sources/Interpreter/TypeLayout.swift
RishabhRD: I find the word `Nearest` to be useful in understanding the function and not needless. I know definition of rounding implies it, but nearest word helps to not think of definition and instantly understand what function does.

tothambrus11:
Thanks, that makes sense.

Sources/Interpreter/TypeLayout.swift
dabrahams, line 69:
```suggestion
  internal func roundedUp(toAMultipleOf n: Self) -> Self {
```

dabrahams (commented):
LGTM

Issue #337, opened 2026-07-21
Unsupported dynamically sized alloca


tothambrus11:
```hylo
//! exit-status:1234
struct X is Deinitializable {
  public var x: Int32
  public memberwise init

  fun take_value(from source: sink Self) auto { 
    set {
      &self = X(x: source.x)
    }
    inout {
      &self.x = source.x
    }
  }

}

given X is Movable {}

fun pipe<T is Movable>(x: sink T) -> T {
  var r: T = x // Dynamically sized alloca

  return r
}

fun main() -> Int32 {
  sink let a = X(x: 1234 as Int32)
  
  let b = pipe(a)
  return b.x.copy()
}
```

```
BackEnd/Program+LLVM.swift:238: Assertion failed: unsupported dynamically sized alloca

  0                  0x00007c21fd4f8d91 _assertionFailure(_:_:file:line:flags:) + 225 in libswiftCore.so
  1 [ra]             0x00005fa2fa201281 Program.incorporate(_:in:) + 768 in HyloPackageTests.xctest at /home/ambrus/hylo-new4/Sources/BackEnd/Program+LLVM.swift:238:5
  2 [ra]             0x00005fa2fa1ff719 Program.incorporate(_:in:) + 888 in HyloPackageTests.xctest at /home/ambrus/hylo-new4/Sources/BackEnd/Program+LLVM.swift:171:14
  3 [ra]             0x00005fa2fa1ff378 Program.incorporate(from:in:) + 503 in HyloPackageTests.xctest at /home/ambrus/hylo-new4/Sources/BackEnd/Program+LLVM.swift:157:14
  4 [ra]             0x00005fa2fa1ea569 Program.incorporate(contentsOf:into:in:) + 2488 in HyloPackageTests.xctest at /home/ambrus/hylo-new4/Sources/BackEnd/Program+LLVM.swift:136:7
  5 [ra]             0x00005fa2fa1e90ec Program.incorporate(_:in:) + 667 in HyloPackageTests.xctest at /home/ambrus/hylo-new4/Sources/BackEnd/Program+LLVM.swift:98:5
  6 [ra]             0x00005fa2fa1feb2b Program.compileToLLVM(_:target:) + 1162 in HyloPackageTests.xctest at /home/ambrus/hylo-new4/Sources/BackEnd/Program+LLVM.swift:35:7
[...]

PR #449, opened 2026-08-10
Implement support for allocating existentialized types

kyouko-taiga:
Fixes part of #337.

This patch fixes the handling of calls to trait implementations. Specifically, it ensures that the abstract type of interfaces (i.e., the functions actually stored in witness tables) are not forgotten during lowering, which is necessary to let code generation handle parameters properly.

The fix introduces a type inconsistency in the IR. Because we keep the types of these interfaces abstract, a call to a function obtained via a `property` instruction may appear ill-typed (e.g., passing an instance of `Int` to a function expecting a skolem). This problem can be solved by inserting `place_cast` instructions. However, this change is non-trivial to implement and thus out of scope for this PR.

This patch also implements support for allocating existentialized types. See the tests for a concrete example.


Sources/BackEnd/Program+LLVM.swift
```diff
@@ -235,14 +235,31 @@ extension Program {
...
     let s = ctx.ir.at(i)
-    assert(s.witness == nil, "unsupported dynamically sized alloca")
+    let v = IRValue.register(i.erased)
 
-    let t = metadata(of: s.storage, in: &ctx.module)
-    let x = ctx.module.llvm.insertAlloca(t.llvm, atEntryOf: ctx.llvm)
-    ctx.module.llvm.setAlignment(t.layout.alignment, for: x)
+    // Is the alloca dynamically sized?
+    if let t = s.witness {
+      let x0 = ctx.module.llvm.insertLoad(
```
tothambrus11: Here it would be very helpful to add meaningful variable names. It's a lot of x-es and by the time I read a line I forget the previous one.

Sources/BackEnd/ModuleGenerationContext.swift
```diff
@@ -95,7 +95,7 @@ internal struct ModuleGenerationContext: ~Copyable {
     self.plateau = self.llvm.functionType(from: [ptr, ptr, fun, ptr], to: i32)
     self.nestedProject = self.llvm.structType([ptr, fun, ptr])
     self.plateauCallback = self.llvm.structType([fun, ptr])
-    self.typeWitnessHeader = [iptr, i32, i16, i16]
+    self.typeWitnessHeader = self.llvm.structType(named: "$htwh", [iptr, i32, i16, i16])
```
tothambrus11, line 98: Since struct type names never get into an object file (apart from possibly debug symbols), we could potentially rename this to `$htypewitnessheader` to be less cryptic.

Either way, we need to adjust the demangler accordingly because it probably expects an operator at the beginning and would produce an error if it didn't find one.

I realize we don't yet have a design doc for the mangling scheme, but once we have it, we need to describe these kinds of exceptions there so we can keep reasoning about the scheme without looking too much at the code.

Sources/BackEnd/Program+LLVM.swift
kyouko-taiga: It's counterproductive to try to name each inserted instruction because really what documents them is the statement. That's an artifact of lowering/codegen being extremely imperative and administrative.

In fact we're basically writing ANF, which is why we have to declare so many local variables. If we could get more declarative we would write things like `load(getelementptr(p0, 0, 0))` and there would be no need for naming subexpressions.

Sources/BackEnd/ModuleGenerationContext.swift
kyouko-taiga: Our demangler should be resilient to errors so if it can't recognize a symbol it should just leave it there. Operators must have a capital letter so a lowercase name can never clash with a mangled symbol.

kyouko-taiga:
In fact it's a full fix for #337.

Sources/Utilities/SortedDictionary.swift
```diff
@@ -96,6 +96,56 @@ public struct SortedDictionary<Key: Comparable, Value> {
...
 
+  /// Returns a dictionary with the contents of `self` merged with the contents of `other`, using
+  /// `combine` to determine the values of duplicate keys.
+  ///
+  /// Complexity: O(n) where n is the number of key/value pairs in the largest dictionary.
+  public func merging(
+    _ other: Self, uniquingKeysWith combine: (Value, Value) throws -> Value
+  ) rethrows -> Self {
+    // Trivial if either side is empty.
+    if other.isEmpty { return self } else if other.isEmpty { return other }
```
tothambrus11:
```suggestion
    if other.isEmpty { return self } else if self.isEmpty { return other }
```

Sources/Utilities/SortedDictionary.swift
```diff
@@ -96,6 +96,56 @@ public struct SortedDictionary<Key: Comparable, Value> {
...
+
+    // Otherwise, merge the contents of the RHS into the LHS.
+    var keys = ContiguousArray<Key>()
+    keys.reserveCapacity(self.count)
+    var values = ContiguousArray<Value>()
+    values.reserveCapacity(self.count)
```
tothambrus11:
```suggestion
    let c = Swift.max(self.count, other.count)
    keys.reserveCapacity(c)
    var values = ContiguousArray<Value>()
    values.reserveCapacity(c)
```

Sources/BackEnd/Program+LLVM.swift
```diff
@@ -235,14 +235,31 @@ extension Program {
...
+        ctx.module.llvm.ptr, from: ctx.value[t]!, at: ctx.insertionPoint!)
+      let x1 = ctx.module.llvm.insertGetStructElementPointer(
+        of: x0, typed: ctx.module.typeWitnessHeader, index: 1, at: ctx.insertionPoint!)
+      let x2 = ctx.module.llvm.insertLoad(
+        ctx.module.llvm.i32, from: x1, at: ctx.insertionPoint!)
+
+      let x3 = ctx.module.llvm.insertAlloca(
+        arrayOf: x2, ctx.module.llvm.i8, at: ctx.insertionPoint!)
+      ctx.module.llvm.setAlignment(ctx.module.dynamicAllocationAlignment(), for: x3)
+      ctx.value[v] = x3.v
```
```suggestion
      let witness = ctx.module.llvm.insertLoad(
        ctx.module.llvm.ptr, from: ctx.value[t]!, at: ctx.insertionPoint!)
      let sizeField = ctx.module.llvm.insertGetStructElementPointer(
        of: witness, typed: ctx.module.typeWitnessHeader, index: 1, at: ctx.insertionPoint!)
      let size = ctx.module.llvm.insertLoad(
        ctx.module.llvm.i32, from: sizeField, at: ctx.insertionPoint!)

      let allocation = ctx.module.llvm.insertAlloca(
        arrayOf: size, ctx.module.llvm.i8, at: ctx.insertionPoint!)
      ctx.module.llvm.setAlignment(ctx.module.dynamicAllocationAlignment(), for: allocation)
      ctx.value[v] = allocation.v
```
I think "imperative nature" is not a good enough reason that it should take me 20 minutes to figure out the correct names for these variables. When I look at a line containing only single-letter variable names, and the lines above are also mostly using single-letter names, it becomes computationally hard to solve the puzzle, and eventually forces top to bottom very careful reading.

Sources/Utilities/SortedDictionary.swift
kyouko-taiga: I considered `self.count + other.count` as capacity, based on the hypothesis that most calls to `merge` are with disjoint dictionaries. But this strategy over allocates whenever there is an overlap. That's why I opted for `self.count`. It is likely someone writes `xs.merge(ys ...)`, which `ys` being smaller and partially overlapping. Of course it's just an educated guess.

`max(self.count, other.count)` is a sour spot I think. On one hand of the spectrum we have `LHS` covers all keys from `RHS` and so we need just `self.count` (same for RHS, by symmetry). On the other side of the spectrum we have `self.count + other.count`. This suggestion seems to pick something relatively close to the first end. But then that means if the two dictionaries share few keys then we're allocating just as much.

That being said, the cost of paying for `max` is probably insignificant.

Sources/BackEnd/Program+LLVM.swift
kyouko-taiga: Your suggestions do not help me read the code better. Besides, adopting the practice of naming each individual register will require endless nitpicking about what's the right way to distinguish between a pointer to a value and the value itself.

We have applied the `xn` convention at large in the compiler. I'm happy to add a few lines before blocks of code to clarify if necessary, but I won't start naming each register.

Sources/Utilities/SortedDictionary.swift
tothambrus11: IIUC the final count of a merge is in `[max(a.count, b.count),  a.count + b.count]`, so choosing the lower end guarantees no overallocation while ensuring we allocate as much as minimally required. If for some reason `rhs` was greater, we would pay `~log(m)` reallocations without this. On the other hand, based on your predicted statistics, choosing `a.count + b.count` may be even faster, since that guarantees no reallocations while probably wasting very little extra space.

kyouko-taiga: I think we need benchmarks to decide, ideally in real situations. I'll keep `self.count` for the moment because I'm a little scared of overallocation.

Sources/BackEnd/Program+LLVM.swift
tothambrus11: I understand that it won't help you read the code better, and it may take extra effort to come up with descriptive variable names. At the same time, it was much harder for me, as I didn't write the code myself. To understand this block of code, I had to know the type layout of a type witness and type witness header, then match that with the way we actually did the llvm instruction insertion calls.

`insertAlloca(arroyOf: x0, i8, ...)` confused me for a while. I thought `x0` is some array. Reading `count` makes that much clearer.

> endless nitpicking about what's the right way to distinguish between a pointer to a value and the value itself

I would be more than happy to settle on a compromise, whatever wording we choose. When we don't have type information helping us, we need to distinguish between them. It's the same reason I would write more verbose names in untyped Python than in Hylo. I would much rather see a descriptive variable name than some comment that I need to match to code to understand what it's doing.

I don't think we need to name each register everywhere, but I do think code should tell a story, and when we need inline comments to clarify what's going on, we should reconsider how we write that code.

Much of my opinions come from this talk: https://www.youtube.com/watch?v=aiy5TrU-Hwc. I do value 1-2-letter variables in short and simple code, and when its usage is close to its declaration, but as code grows, I found that it hurts readability.

Sources/Utilities/SortedDictionary.swift
tothambrus11: `max(a.count, b.count)` would also guarantee no overallocation, and at the cost of a single `max`, it reduces the risk of reallocation.

kyouko-taiga: The night convinced myself that your suggestion may be the right bet.

Sources/BackEnd/Program+LLVM.swift
kyouko-taiga: If you want to clarify sequences of codegen statements, the tool is to write an abstraction, not to name each register. I'll elaborate.

We don't name subexpressions because their position in the expression is documenting "the story". When we `open("xyz", mode: "a")` we don't need a name for each argument because we know what they mean from context (you could argue that labels are name, which does not contradict my point).

Sadly, we have to write codegen in ANF because Swift has questionable rules about order of operation. So we can't write `open("xyz", mode: "a")` and instead we have to write `let x0 = "xyz" ; let x1 = "a" ; let x2 = open(x0, mode: x1)`. Insisting on finding descriptive names for the subexpressions is increasing cognitive overload, because that was never necessary in the original form (in fact we would have probably asked someone to refactor `let path = "xyz" ; let openMode = "a" ; open(path, mode: openMode)`). Those names are only appealing because of a technical limitation. Working around this limitation means finding ways to express the original, declarative intent of the expression (i.e., `open("xyz", mode: "a")`).

tothambrus11: Ah I see, thanks for the explanation. This document suggest there is left to right evaluation order: https://github.com/swiftlang/swift-evolution/blob/main/proposals/0352-implicit-open-existentials.md#order-of-evaluation-restrictions

I would still either split the count as a separate variable or rename the SwiftyLLVM API to include arrayOfLength

kyouko-taiga: I think `m.insertAlloca(arrayOf: n, i8)` reads better but I can live with another label if you insist.

PTAL at my latest changes before you open a PR to Swifty LLVM.

Sources/BackEnd/ModuleGenerationContext.swift
tothambrus11: https://github.com/hylo-lang/hylo-new/issues/455

Sources/BackEnd/Program+LLVM.swift
tothambrus11: The refactoring into a separate function and using the name `n` was very helpful. No need to change the label.

PR #457, opened 2026-08-11
Wait until exit after draining inputs

tothambrus11:


dabrahams:
Please see https://github.com/dabrahams/SPMBuildToolSupport/issues/13#issuecomment-3679175247

tothambrus11:
Thanks for the thread, that's interesting. The recent comments admit that it wasn't fully working though IIUC. We might also have the different problem that we are actually interested in reading all stout and stderr, which would be prevented by closing the streams early.

dabrahams:
I think this is superseded by https://github.com/hylo-lang/hylo-new/commit/5a66b3c942f91d8c3b273ec19a3ffd07d7da63fe

dabrahams:
My understanding from that thread is that closing the streams is not incompatible with reading from the pipes.  However, the code in that library doesn't read stderr and is much too complicated to cannibalize.  I implemented the method used in gyb-swift.

tothambrus11:
Avoiding new threads is pretty nice.

dabrahams:
@tothambrus11 Gonna close this; feel free to reopen if I'm wrong to do so

PR #460, opened 2026-08-11
Include the list of test cases as a build system dependency

### LINKED ISSUE #414: Generated test suite doesn't get regenerated when adding/removing hylo files
CompilerTestsPlugin declares `inputFiles: []`, so SPM doesn't regenerate the test list when you add or remove a .hylo file. I had to delete the cached CompilerTests+GeneratedTests.swift under `.build/plugins/outputs/` or touch `Sources/hc-tests/CommandLine.swift` to trigger generation. We should make it so that when the folder contents change, the build plugin reruns.

tothambrus11:
Fixes #414

### DIFF
--- a/Plugins/CompilerTestsPlugin/CompilerTestsPlugin.swift
+++ b/Plugins/CompilerTestsPlugin/CompilerTestsPlugin.swift
@@ -11,14 +11,35 @@ struct CompilerTestsPlugin: BuildToolPlugin {
     let output = context.pluginWorkDirectoryURL
       .appending(component: "CompilerTests+GeneratedTests.swift")
 
+    let snapshot = context.pluginWorkDirectoryURL.appending(component: "test-cases.txt")
+    let tests = testCases(in: target.directoryURL)
+      .map({ (u) in u.path(percentEncoded: false) })
+      .sorted()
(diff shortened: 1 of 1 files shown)


Plugins/CompilerTestsPlugin/CompilerTestsPlugin.swift
```diff
@@ -11,14 +11,35 @@ struct CompilerTestsPlugin: BuildToolPlugin {
     let output = context.pluginWorkDirectoryURL
       .appending(component: "CompilerTests+GeneratedTests.swift")
 
+    let snapshot = context.pluginWorkDirectoryURL.appending(component: "test-cases.txt")
```
kyouko-taiga, line 14: The idea is that the plugin updates this snapshot file every time it runs when the list doesn't match?

tothambrus11: Yes. The alternative was to supply the list of files directly as input, but that triggers a regeneration even if you just modify the contents of a test file. This should only regenerate if the set of test cases changed.

kyouko-taiga (approved):
LGTM

PR #463, opened 2026-08-12
Diagnose illegal mutations of 'sink let' bindings

### LINKED ISSUE #344: Enforce immutability in `sink let` bindings
Currently, `sink let` works more or less like a `var` binding. We should enforce that after the binding actually happens (first assignment), we cannot mutate the value except by a `sink` access. We should rule out reassignment (set access) and in-place mutation (inout access).

This problem seems to appear in these two places:

```hylo
struct S is Deinitializable {
  let x: Int32

  public init(x: Int32) {
    if true {
      &self.x = x.copy() // first assignment should be legal (works)
      &self.x = 23  // reassignment should be illegal
      &self.x += 2  // inout access should be illegal
    } else {
      &self.x = 3 as Int32
    }
  }
}
```

```hylo
public fun main() -> Int32 {
  sink let x: Int32

  if true {
    &x = 2  // First assignment should be legal (works)
  } else {
    &x = 3  // First assignment should be legal (works)
    &x = 4  // reassignment should be illegal
    &x += 1  // inout access should be illegal
  }

  return x  // sink access should be legal (works)
}
```

kyouko-taiga:
This patch fixes #344

### DIFF
--- a/Sources/FrontEnd/IR/Transforms/IRFunction+LifetimeNormalization.swift
+++ b/Sources/FrontEnd/IR/Transforms/IRFunction+LifetimeNormalization.swift
@@ -694,28 +694,27 @@ private struct Transfer: AbstractTransferFunction {
     }
   }
 
-  /// Returns the pattern binding declaration introducing the entity referred to by `v`, which is
-  /// the source of an access in `f`, or `nil` if such a declaration cannot be identified.
-  private func binding(declaring v: IRValue, in f: IRFunction) -> BindingDeclaration.ID? {
-    let b: VariableDeclaration.ID? = if let d = f.declaration(v) {
-      program.cast(d, to: VariableDeclaration.self)
+  /// Returns the declaration binding `v`, which computes the source of an access in `f`, or `nil`
+  /// if such a declaration cannot be identified.
+  private func binding(declaring v: IRValue, in f: IRFunction) -> VariableDeclaration.ID? {
+    if let d = f.declaration(v) {
+      return program.cast(d, to: VariableDeclaration.self)
     } else if let r = v.register, let s = f.at(r) as? IRSubfield, let d = s.declaration {
-      program.cast(d, to: VariableDeclaration.self)
+      return program.cast(d, to: VariableDeclaration.self)
     } else {
-      nil
+      return nil
     }
-    return b.flatMap(program.bindingDeclaration(containing:))
   }
 
-  /// Returns the introducer of the binding associated with `v`, which is the source of an access
-  /// in `f`, along with a value defining the place referred to by `v`.
+  /// Returns the introducer of the binding associated with `v`, which computes the source of an
+  /// access in `f`, along with a value defining the place referred to by `v`.
   private func introducer(
     binding v: IRValue, in f: IRFunction
   ) -> (BindingPattern.Introducer?, IRValue) {
     var s = v
     while true {
       // Is `s` attached to a binding declaration?
-      if let b = binding(declaring: s, in: f) {
+      if let b = binding(declaring: s, in: f).flatMap(program.bindingDeclaration(containing:)) {
         let k = program[program[b].pattern].introducer.value
         if (k == .let) && (!program.isLocal(b) || program.isMember(b)) {
           return (.sinklet, s)
--- a/Sources/FrontEnd/IR/Transforms/IRFunction+LifetimeNormalization.swift
+++ b/Sources/FrontEnd/IR/Transforms/IRFunction+LifetimeNormalization.swift
@@ -13,6 +13,9 @@ extension IRFunction {
   /// deallocated or flows into a `set` access. This situation may occur when deinitialization was
   /// left implicit during IR lowering. These new instructions are emitted into `m`, using `typer`
   /// to resolve implementations.
+  ///
+  /// Lifetime normalization is part of the mandatory transformation pipeline and is expected to
+  /// run after access reification and last use analysis.
   internal mutating func normalizeLifetimes(
     emittingInto m: Module.ID, using typer: inout Typer
   ) -> Bool {
--- a/Sources/FrontEnd/IR/Transforms/IRFunction+LifetimeNormalization.swift
+++ b/Sources/FrontEnd/IR/Transforms/IRFunction+LifetimeNormalization.swift
@@ -235,7 +235,7 @@ private struct Transfer: AbstractTransferFunction {
 
     // Check if the access is violating immutability. If it is, then report an illegal access and
     // skip further changes to the context to avoid cascading diagnostics.
-    let isLegal = isUpholdingImmutability(i, in: f)
+    let isLegal = upholdsBindingImmutability(i, in: f)
     if !isLegal {
       report(program.illegalAccess(k, at: access.anchor))
     }
@@ -739,7 +739,7 @@ private struct Transfer: AbstractTransferFunction {
   /// Control flow is taken into account when the place is referred to by a `sink let` binding. In
   /// this case, a `set` access is legal only if the source is fully uninitialized, and a `sink`
   /// access is legal only if the source is fully initialized.
-  private mutating func isUpholdingImmutability(_ i: IRAccess.ID, in f: IRFunction) -> Bool {
+  private mutating func upholdsBindingImmutability(_ i: IRAccess.ID, in f: IRFunction) -> Bool {
     let s = f.at(i)
     let k = s.capabilities.uniqueElement!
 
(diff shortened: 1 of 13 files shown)


Sources/FrontEnd/IR/Transforms/IRFunction+LifetimeNormalization.swift
```diff
@@ -694,6 +694,88 @@ private struct Transfer: AbstractTransferFunction {
...
+        switch f.tag(of: r) {
+        case IRCase.self:
+          s = (f.at(r) as! IRCase).source
+        case IRPlaceCast.self:
+          s = (f.at(r) as! IRPlaceCast).source
+        case IRSubfield.self:
+          s = (f.at(r) as! IRSubfield).base
+        default:
+          return (nil, s)
+        }
```
tothambrus11: What is the condition for adding these cases? If in the future we introduce a new instruction, how are we going to detect a bug if we forget to update this switch? I would use an exhaustive pattern matching here, listing all currently known instructions that don't apply here, and forcing the decisions on the person introducing a new instruction.

Sources/FrontEnd/IR/Transforms/IRFunction+LifetimeNormalization.swift
```diff
@@ -694,6 +694,88 @@ private struct Transfer: AbstractTransferFunction {
...
+
+      // No binding declaration.
+      else { return (nil, s) }
+    }
+  }
+
+  /// Returns `true` iff `i`, which is in `f`, is not a mutable access on an immutable binding.
+  private mutating func isUpholdingImmutability(_ i: IRAccess.ID, in f: IRFunction) -> Bool {
+    let s = f.at(i)
+    let k = s.capabilities.uniqueElement!
```
tothambrus11, line 746: How do we know that an IRAccess instruction has a single capability here? Is this something that holds the possible capabilities before lifetime normalization but upon normalization we decide on a single one?

Maybe we should add a precondition that this function is run after lifetime normalization of `f`.

kyouko-taiga: This function runs _during_ lifetime normalization. Its call site even has a comment answering your question.

Sources/FrontEnd/IR/Transforms/IRFunction+LifetimeNormalization.swift
```diff
@@ -694,6 +694,88 @@ private struct Transfer: AbstractTransferFunction {
...
+        }
+      }
+
+      // No binding declaration.
+      else { return (nil, s) }
+    }
+  }
+
+  /// Returns `true` iff `i`, which is in `f`, is not a mutable access on an immutable binding.
+  private mutating func isUpholdingImmutability(_ i: IRAccess.ID, in f: IRFunction) -> Bool {
```
tothambrus11: We are in an extension of IRFunction, can't we use `self` instead of `f`?

Sources/FrontEnd/IR/Transforms/IRFunction+LifetimeNormalization.swift
kyouko-taiga: A case is necessary if the instruction presents a place whose source is on a path that may be attached to a declaration.

Listing all cases is not really a great idea. It will add a lot of noise, which will show up in source but also in coverage reports by creating code paths that are simply unreachable. There is nothing mechanically guaranteeing exclusivity for a switch over IR instructions so the risk of breaking consistency is very high. And on top of it, it won't even make it easier to know if an update to the switch is necessary when we introduce a new instruction.

To detect bugs we'll have to run tests and debug unexpected situations when we make changes to the IR.

Sources/FrontEnd/IR/Transforms/IRFunction+LifetimeNormalization.swift
kyouko-taiga: This is a method of `Transfer`.

Sources/FrontEnd/IR/Transforms/IRFunction+LifetimeNormalization.swift
tothambrus11: Ok, then we should add the precondition: "Access of `i` is reified." or "Accesses in `f` are reified".

kyouko-taiga: That is a precondition of the whole pass. We don't need to add a precondition to each function internal to the pass explicitly.

Sources/FrontEnd/IR/Transforms/IRFunction+LifetimeNormalization.swift
tothambrus11: Oh I see, it's not an enum. It would be great if we could convert to one. I'm concerned that this pattern creates a lot of loose ends and lots of tiny subtle decisions that are usually not exercised when adding a new instruction, until the new instruction is used in the exact complicated way where the distinction would surface as a test failure/crash. I think this has a pretty high risk. I've been previously bitten by cases where the failure to correctly add a case for an instruction resulted in a quite distance assertion failure, and it was very hard to track down. The assertion only came up when I actually tried to use that instruction in my program in the presence of subscripts, not when that instruction was added.

Btw I don't think multiple case conditions show up in coverage, since all take the same branch. Unfortunately (fortunately in this case), Swift doesn't support measuring condition coverage or expression-level coverage, only line and function coverage.

Sources/FrontEnd/IR/Transforms/IRFunction+LifetimeNormalization.swift
tothambrus11: If it's a precondition of the whole pass, it's not written down currently. Even if it was, this is still a regular function and nothing prompts me to read the detailed docs of the pass apart from being a function in a file in `IR/Transforms/`.

kyouko-taiga: By this logic we need to add `the program is type checked` as a precondition of all functions in the IR pipeline.

Sources/FrontEnd/IR/Transforms/IRFunction+LifetimeNormalization.swift
kyouko-taiga: PTAL at the latest version.

Sources/FrontEnd/IR/Transforms/IRFunction+LifetimeNormalization.swift
tothambrus11: Fair enough. Then we just need to write it in the pass documentation.

Sources/FrontEnd/IR/Transforms/IRFunction+LifetimeNormalization.swift
tothambrus11: Nice, the best is to eliminate the switch altogether :D

Tests/CompilerTests/negative/illegal-sink-let.hylo
```diff
@@ -0,0 +1,44 @@
...
+    &self.y = 0       // OK (initialization)
+    &self.y = 1       // OK (assignment)
+  }
+
+  fun f0() inout {
+    &self.x = 0       // error
+    &self.y = 1       // OK (assignment)
+  }
+
+  fun f0() sink -> Int {
```
tothambrus11:
```suggestion
  fun f1() sink -> Int {
```

Sources/FrontEnd/IR/Transforms/IRFunction+LifetimeNormalization.swift
```diff
@@ -694,6 +697,82 @@ private struct Transfer: AbstractTransferFunction {
...
+
+  /// Returns the introducer of the binding associated with `v`, which computes the source of an
+  /// access in `f`, along with a value defining the place referred to by `v`.
+  private func introducer(
+    binding v: IRValue, in f: IRFunction
+  ) -> (BindingPattern.Introducer?, IRValue) {
+    var s = v
+    while true {
+      // Is `s` attached to a binding declaration?
+      if let b = binding(declaring: s, in: f).flatMap(program.bindingDeclaration(containing:)) {
```
tothambrus11, line 720: The binding we find this way may be a binding of a `var` struct member, even though we have `sink let` on the outer storage.

This program is currently accepted
```
struct S is Deinitializable {

  public var y: Int

  public init() {
    &self.y = 0
  }

}

public fun main() {
  sink let s = S()
  &s.y = 1 // illegal set access
  &s.y += 1   // illegal inout access
}
```

tothambrus11 (changes requested):

Sources/FrontEnd/IR/Transforms/IRFunction+LifetimeNormalization.swift
kyouko-taiga: I have thought about this problem but we need a different way to fix it. Won't be in this PR.

In a nutshell we have to modify how we lower uses of stored properties so that we insert an access that will be the source of `subfield`.

tothambrus11 (approved):

PR #469, opened 2026-08-12
Add a note about possibly unmatched block comment delimiters

kyouko-taiga:


tothambrus11 (approved):
I think part of the problem was also in a position where we tried to look up */ in a position that didn't have an explicit () expression.

kyouko-taiga:
> I think part of the problem was also in a position where we tried to look up */ in a position that didn't have an explicit () expression.

I don't understand what you mean. Can you show an example?

For information, in the absence of a leading `/*`, `*/` will be parse as an operator in the following cases:
- right before an expression (e.g., `*/a`), as a prefix operator;
- right after an expression (e.g. `a*/`), as a postfix operator;
- in between two expressions (e.g., `a */ b`), as an infix operator.

Note that there may be a newline after any infix operator. The following is syntactically legal and unambiguous:

```hylo
()
*/
()
```

tothambrus11:
Oh, I see. There was a statement afterwards that returned Void, so we could match that as an operand for `*/`.

tothambrus11:
What do you think about ruling out the `*/` and `/*` names as reserved keywords? They shouldn't be possible to write except with `` `*/` `` and `` `/*` ``.

kyouko-taiga:
I think the note is fine. Block comments are rare and the note makes the problem obvious. Let's not overthink this issue.

PR #473, opened 2026-08-13
Fix the priority inversion issue.

### LINKED ISSUE #467: Runtime warnings about priority inversion
Running tests on macOS triggers many runtime warnings complaining about priority inversion. For example:

> Thread running at User-initiated quality-of-service class waiting on a lower QoS thread running at Default quality-of-service class. Investigate ways to avoid priority inversions

The issue seems to be triggered by the call `group.wait()`, in `readPipeInBackground()`, in `Utilities/Process+Extensions.swift`.

This [SO post](https://stackoverflow.com/questions/74918932/dispatchqueue-global-default-qos-is-userinitiated) suggests we may be doing things sub-optimally.

Version: 8f809caaa1b93ba45057e8ab432c3bef79e2a2f1

dabrahams:
fixes #467

### DIFF
--- a/Sources/Utilities/Process+Extensions.swift
+++ b/Sources/Utilities/Process+Extensions.swift
@@ -80,8 +80,8 @@ extension Process {
     process.waitUntilExit()
 
     // Retrieve the data (blocks until background reads complete)
-    let output = stdoutData().decodedAsRepairedUTF8()
-    let error = stderrData().decodedAsRepairedUTF8()
+    let output = try stdoutData().decodedAsRepairedUTF8()
+    let error = try stderrData().decodedAsRepairedUTF8()
 
     return .init(
       standardOutput: output,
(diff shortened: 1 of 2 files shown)


Sources/Utilities/Process+Extensions.swift
```diff
@@ -119,33 +119,84 @@ extension Process {
...
     }
   }
 
   return {
-    group.wait()
-    return box.data
+    Task {
+      await output.finished.condition.wait()
+    }
+    return output.contents
```
tothambrus11: The created `Task { await ... }` exits immediately (never awaited) and `output.contents` is undeterministically truncated. We should make this whole function and all its callees async, then await the task.

tothambrus11:
Thanks for the work! I'm closing this in favor of https://github.com/hylo-lang/hylo-new/pull/497.

PR #474, opened 2026-08-13
Refactor the typing of conditional statements

### LINKED ISSUE #424: Simplify `inferredType(of b: Block.ID, occurringAsStatement isStatement: Bool)`
The function is only ever used with isStatement=true.

See comments here:

Image

kyouko-taiga:
Just a minor simplification of type inference, removing needless parameters.

Closes https://github.com/hylo-lang/hylo-new/issues/424

### DIFF
--- a/Sources/FrontEnd/Typer/Typer.swift
+++ b/Sources/FrontEnd/Typer/Typer.swift
@@ -1165,6 +1165,10 @@ public struct Typer {
   }
 
   /// Type checks `s`.
+  ///
+  /// `s` occurs as the else branch of a conditional statement occurring as a statement. If the
+  /// branch is another conditional expression (i.e., `... if else ...`), then that expression is
+  /// also checked as a statement.
   private mutating func check(_ s: If.ElseIdentity) {
     if let n = program.cast(s, to: If.self) {
       checkAsStatement(n)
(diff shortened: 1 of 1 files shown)


Sources/FrontEnd/Typer/Typer.swift
```diff
@@ -1152,12 +1152,29 @@ public struct Typer {
...
   /// Type checks `s`.
   private mutating func check(_ s: Discard.ID) {
     check(program[s].value)
     program[s.module].setType(.void, for: s)
   }
 
+  /// Type checks `s`.
+  private mutating func check(_ s: If.ElseIdentity) {
+    if let n = program.cast(s, to: If.self) {
+      checkAsStatement(n)
```
tothambrus11, line 1174: It is counterintuitive why an `else` node that's actually another `if` would be a statement. Some explanation or precondition would be nice to add.

```hylo
let x = if a { 1 } else if b { 2 } else { 3 }
```
I expect the second `if` to be also an expression. Does that somehow contain a statement that is wrapped again into a statement? How does this work?

kyouko-taiga: A conditional expression can appear either as an expression or as a statement. We use `Program.branches(of:)` to get the two trees representing both branches of the conditional. If this method returns `nil`, then its argument is considered as statement and we'll eventually get in `checkAsStatement(_:)`.

I added some documentation to clarify.

tothambrus11 (approved):
Nice simplification!

PR #475, opened 2026-08-14
Interpreter: support statically sized allocations for alloca

RishabhRD:
Now we can write:
```
let x = 2
```

The PR doesn't cover dynamic sized allocations for alloca, that would need handling witness. Before that I would need to implement function calls, then only we can write programs that actually does dynamically sized alloca.


Sources/Interpreter/Interpreter.swift
```diff
@@ -52,8 +52,38 @@ extension Program {
...
+  init(_ x: Any) {
+    storage = x
+  }
+
+  /// The memory location pointed to by `self`, if any.
+  public var location: Memory.TypedAddress? {
+    if let a = storage as? Access<Memory.TypedAddress> {
+      a.location
+    } else if let a = storage as? Memory.TypedAddress {
+      a
```
RishabhRD, line 69: Answering [this comment](https://github.com/hylo-lang/hylo/pull/1882/changes#r3409916040) from OFE interpreter.

> It's surprising to see the “if it's an access use its place; else if it's a place use that directly” logic.

This is because at IR level, we don't differentiate between place and access. In-fact from `IRAccess`:
```swift
  /// The type of the storage accessed by this instruction.
  public var type: IRType {
    .same(as: source)
  }
```
where type is the type of Instructions's result. And thus there would be instructions trying to get subfield from result of an access.

But at the Interpreter level, we need to differentiate between access and place for runtime safety and thus we need this logic.

Sources/Interpreter/Access.swift
```diff
@@ -0,0 +1,27 @@
...
+import FrontEnd
+import Utilities
+
+/// An access to a `Region` occurring during execution.
+public struct Access<Region: Regular>: Regular {
+
+  /// A unique `Access` identifier.
+  public typealias ID = UUID
+
+  /// A unique `Access` identifier.
```
dabrahams: This is like saying:

```swift
/// A string.
let x: String
```

What _role_ does this value play?

I have a hard time describing that in fact, because these instances are copyable, so it's not “the identity of this instance.”  Maybe it identifies something like the particular instruction in a particular thread that initiated the access?

Sources/Interpreter/Access.swift
```diff
@@ -0,0 +1,27 @@
...
+  public let id: ID
+
+  /// The associated permissions and obligations.
+  public let effect: AccessEffect
+
+  /// The location to which access applies.
```
dabrahams:
```suggestion
  /// The part of memory being accessed.
```

Sources/Interpreter/Access.swift
```diff
@@ -0,0 +1,27 @@
...
+  public let location: Region
+
+  /// Creates an instance of access to `r` having effect `e`.
```
dabrahams:
```suggestion
  /// Creates an instance accessing `r` with effect `e`.
```

Sources/Interpreter/Interpreter.swift
```diff
@@ -287,5 +397,15 @@ extension Program {
   }
 }
 
+extension IRAccess {
```
dabrahams, line 410:
```suggestion
extension IRAccess {

```

dabrahams (commented):
left some comments

Sources/Interpreter/Access.swift
RishabhRD: How about
```
  /// The identity of an access initiated by a particular instruction during execution.
```

Sources/Interpreter/Interpreter.swift
```diff
@@ -287,5 +397,17 @@ extension Program {
...
+
+  /// The associated permissions and obligations.
+  var finalCapability: AccessEffect {
+    // Because IR analysis should ensure single effect.
+    // See: Sources/FrontEnd/IR/Instructions/IRAccess.swift.
+    precondition(capabilities.count == 1)
+    return capabilities.first!
```
tothambrus11:
```suggestion
    capabilities.uniqueElement!
```
We could use this extension from Utilities.

Sources/Interpreter/RuntimeValue.swift
```diff
@@ -0,0 +1,61 @@
...
+  /// Creates an instance from its bytes representation `bytes`, with alignment `a`.
+  public init(bytes: [UInt8], havingAlignment a: Int) {
+    storage = .init(repeating: 0, count: bytes.count)
+
+    // If we didn't get suitably-aligned storage, allocate enough to
+    // ensure we can find a suitably-aligned region of the right
+    // size.
+    if storage.withUnsafeBytes({ UInt(bitPattern: $0.baseAddress) % UInt(a) != 0 }) {
+      storage = .init(repeating: 0, count: bytes.count + a - 1)
+    }
```
tothambrus11: Is there a chance that the inline buffer is used which now happens to be aligned, but would break if this instance got moved/copied?

This section looks tricky enough to deserve its own function, possibly its own data structure.

Sources/Interpreter/RuntimeValue.swift
```diff
@@ -0,0 +1,61 @@
...
+///
+/// - Precondition: `n` must be representable as a signed integer in `b` bytes.
+func twosComplementRepresentation(_ n: BigInt, size b: Int) -> BigUInt {
+  var unsignedRepresentation = n.magnitude
+  if n.sign == .minus {
+    let maximumUnsignedValue = BigUInt(1) << (8 * b)
+    unsignedRepresentation = maximumUnsignedValue - unsignedRepresentation
+  }
+  return unsignedRepresentation
+}
```
tothambrus11, line 81: We could add some unit tests for this. I like shifting testing closer to unit tests, as the integration tests are unlikely to catch all edge cases/bugs of lower levels.

Sources/Interpreter/RuntimeValue.swift
RishabhRD: Not if storage is immutable. Updated the code to make sure storage is immutable.

Sources/Interpreter/RuntimeValue.swift
RishabhRD: Nice! It should have been unit tested. Thanks for pointing this out :-)

Sources/Interpreter/Access.swift
dabrahams: Yeah, I guess that's OK. I was worried about identifying just with the instruction because an instruction is executed multiple times, even simultaneously, which is why I mentioned threads. But everybody knows that an access is temporal so I think that's enough.

Sources/Interpreter/Interpreter.swift
```diff
@@ -52,8 +52,38 @@ extension Program {
...
+    } else if let a = storage as? Memory.TypedAddress {
+      a
+    } else {
+      nil
+    }
+  }
+
+  /// The location pointed to by `self` together with its associated permissions
+  /// and obligations, if any.
+  public var accessedLocation: Access<Memory.TypedAddress>? {
```
```suggestion
  /// `self` if it is an access or `nil` otherwise
  public var asAccess: Access<Memory.TypedAddress>? {
```

But maybe you also want to dispatch to this:

```swift
  /// `self` if it is a `T`, or `nil` otherwise
  public func as<T>(_: T.type) -> T? { storage as? T }
```

Sources/Interpreter/Interpreter.swift
```diff
@@ -183,7 +213,13 @@ public struct Interpreter {
   /// - Precondition: `p.entry != nil`
   public init(_ p: Program) {
     memory = Memory(forRunning: p, on: UnrealABI())
-    callStack.enter(p.entry, definedIn: p, withParameters: [])
+    let l = memory.allocate(.init(.void))
```
dabrahams: I think a comment to explain what all of this is about would be useful

Sources/Interpreter/Interpreter.swift
```diff
@@ -183,7 +213,13 @@ public struct Interpreter {
...
+    let a = Access(
+      to: Memory.TypedAddress(
+        allocation: l.allocation,
+        offset: l.offset, type: .init(.void)),
```
```suggestion
     to: Memory.TypedAddress(
        allocation: l.allocation, offset: l.offset, type: .init(.void)),
```

Much more readable, to my eye.  Canʻt you even write this?

```suggestion
     to: .init(allocation: l.allocation, offset: l.offset, type: .init(.void)),
```

Sources/Interpreter/Interpreter.swift
dabrahams: Actually not surprising at all.  Accesses and places both have a `location`.

Sources/Interpreter/Interpreter.swift
```diff
@@ -200,22 +236,28 @@ public struct Interpreter {
   /// Applies the `Memory` and I/O effects of the current instruction and returns its epilogue.
   private mutating func applyCurrentInstruction() throws -> InstructionEpilogue {
     switch currentInstruction {
-    case is IRAccess:
+    case let x as IRAccess:
       // TODO: add a real implementation, validating new access in memory and
       // storing the access into register.
-      return .initializeRegister(to: .init(storage: ()))
+      let p = asTypedAddress(x.source)
```
dabrahams: This might be more readable, if achievable
```suggestion
      let p = x.source.asTypedAddress(in: self)
```

Sources/Interpreter/Interpreter.swift
```diff
@@ -200,22 +236,28 @@ public struct Interpreter {
...
+      let a = Access<Memory.TypedAddress>(to: p, effect: x.finalCapability)
+      return .initializeRegister(to: .init(a))
     case is IRRegionEnd<IRAccess>:
       // TODO: add a real implementation, validating if it is safe to end the access.
-      return .initializeRegister(to: .init(storage: ()))
+      return .initializeRegister(to: .init(()))
```
dabrahams: This is weird to see.  What is the default value of an `Access`?

Sources/Interpreter/Interpreter.swift
dabrahams: maybe if we had


```suggestion
    let l = memory.allocate(storageFor: .void)
```

It would be clearer?  WDYT about this pattern?

Sources/Interpreter/Interpreter.swift
```diff
@@ -200,22 +236,28 @@ public struct Interpreter {
...
     case let x as IRAlloca:
-      _ = x
+      if x.witness != nil {
+        unimplemented("dynamically sized stack allocation is not supported yet.")
+      }
+      let p = allocateOnStack(.init(x.storage))
```
dabrahams: Why is this called `allocateOnStack` while memory just has `allocate`?  Do we expect the interpreter to know about the heap or just to call functions that return heap memory?

Regardless Iʻd like to read:

```suggestion
      let p = allocate(stackSpaceFor: x.storage)
```

Sources/Interpreter/Interpreter.swift
```diff
@@ -200,22 +236,28 @@ public struct Interpreter {
...
+      return .initializeRegister(to: .init(p))
```
dabrahams: Iʻd like to read:
```suggestion
      return .initializeRegister(to: p)
```

Sources/Interpreter/Interpreter.swift
```diff
@@ -270,6 +317,69 @@ public struct Interpreter {
...
+    }
+  }
+
+  /// Returns the value occurring in the interpreted program that is carried by `v`.
+  private mutating func asRuntimeValue(_ v: IRValue) -> RuntimeValue {
+    switch v {
+    case .register(let r):
+      return topOfStack.registers[r]!.rvalue!
+    case .integer(let n, let t):
+      let l = memory.layout(.init(t.erased))
```
dabrahams: Iʻd like to read:
```suggestion
      let l = memory.layout(t)
```

Sources/Interpreter/Interpreter.swift
```diff
@@ -239,11 +281,16 @@ public struct Interpreter {
...
+        try memory.deallocate(a)
+      }
       return .return
     case let x as IRStore:
-      _ = x
+      try store(x.value, at: x.target)
+      return .initializeRegister(to: .init(()))
     case let x as IRSubfield:
-      _ = x
+      let l = location(x.path, in: x.base)
```
dabrahams: Maybe this would read better?  WDYT?
```suggestion
      let l = x.base.location(ofPart: x.path, in: self)
```

Sources/Interpreter/Memory.swift
```diff
@@ -257,8 +252,41 @@ extension UnsafeRawBufferPointer {
   /// the nearest address aligned to `a`.
   ///
   /// If `self.baseAddress == nil`, returns `0`.
-  fileprivate func firstOffsetAligned(to a: Int) -> Int {
+  internal func firstOffsetAligned(to a: Int) -> Int {
     return baseAddress?.offsetToAlignment(a) ?? 0
```
dabrahams: Shouldnʻt this be


```suggestion
    return baseAddress?.roundedUp(toNearestMultipleOf: a) ?? 0
```

For consistency?  I donʻt love the name `offsetToAlignment` (I know I came up with it)!

Is it really OK for baseAddress to be nil here or should it be `baseAddress!.`…?

dabrahams (approved):
Itʻs getting there; I think we could do some things for readabiity though.  LMK what you think!

Sources/Interpreter/Interpreter.swift
RishabhRD: Yes, explicit `Memory.TypedAddress` was there to infer `Region` type parameter of `Access<Region>`.
But I think then we can make it more readable with:
```swift
    let a = Access<Memory.TypedAddress>(
      to: .init(allocation: l.allocation, offset: l.offset, type: .init(.void)),
      effect: .set
    )
```

Sources/Interpreter/Interpreter.swift
RishabhRD: I like this pattern, just that we would need `` `as` `` because of the keyword thing.

Sources/Interpreter/Interpreter.swift
RishabhRD: But this is `end_access` and we are storing `Void` to signify, the instruction produces `.nothing`.

Sources/Interpreter/Interpreter.swift
RishabhRD: No, we are not expecting interpreter to know about heap. It should have been named `allocate(storageFor: )`.

Sources/Interpreter/Interpreter.swift
RishabhRD: I am aligned with this.

Sources/Interpreter/Memory.swift
RishabhRD: > Is it really OK for baseAddress to be nil here or should it be baseAddress!.…?

It can be nil, when we create allocation of zero-sized types like `.void`.

RishabhRD:
> I think we could do some things for readabiity though. LMK what you think!

@dabrahams I am mostly aligned. Made the changes to achieve similar (if not exact) code for readability as you suggested. I think you should skim through it once, and if you feel everything is in place, we can merge the PR then.

Sources/Interpreter/Interpreter.swift
```diff
@@ -183,7 +204,15 @@ public struct Interpreter {
   /// - Precondition: `p.entry != nil`
   public init(_ p: Program) {
     memory = Memory(forRunning: p, on: UnrealABI())
-    callStack.enter(p.entry, definedIn: p, withParameters: [])
+
+    // `main` takes a `set` access to a `Void` value, so create the
+    // corresponding storage and access.
+    let l = memory.allocate(storageFor: .void)
+    let a = Access<Memory.TypedAddress>(
```
```suggestion
    let a = Access(
```

Sources/Interpreter/Interpreter.swift
```diff
@@ -200,22 +229,28 @@ public struct Interpreter {
   /// Applies the `Memory` and I/O effects of the current instruction and returns its epilogue.
   private mutating func applyCurrentInstruction() throws -> InstructionEpilogue {
     switch currentInstruction {
-    case is IRAccess:
+    case let x as IRAccess:
       // TODO: add a real implementation, validating new access in memory and
       // storing the access into register.
-      return .initializeRegister(to: .init(storage: ()))
+      let p = x.source.asTypedAddress(in: self)
+      let a = Access<Memory.TypedAddress>(to: p, effect: x.finalCapability)
```
```suggestion
      let a = Access(to: p, effect: x.finalCapability)
```

Sources/Interpreter/Interpreter.swift
```diff
@@ -270,6 +310,86 @@ public struct Interpreter {
...
+  /// Returns the memory location pointed to by `self` together with its
+  /// associated permissions and obligations, in the context of `executor`.
+  ///
+  /// - Precondition: `v` contains a place computed by `access` instruction.
+  fileprivate func asAccess(in executor: Interpreter) -> Access<Memory.TypedAddress> {
+    switch self {
+    case .parameter(let i):
+      executor.topOfStack.parameters[i]
+    case .register(let r):
+      executor.topOfStack.registers[r]!.`as`(Access<Memory.TypedAddress>.self)!
```
dabrahams: Really don't love the backticks at the use-site.  If they are really needed we should maybe consider `as_`?  Any other ideas?

Sources/Interpreter/Interpreter.swift
```diff
@@ -270,6 +310,86 @@ public struct Interpreter {
...
+    default:
+      preconditionFailure("\(executor.program.show(self)) is not a Access<Memory.TypedAddress>.")
```
```suggestion
      preconditionFailure("\(executor.program.show(self)) is not an Access<Memory.TypedAddress>.")
```

Sources/Interpreter/Interpreter.swift
```diff
@@ -270,6 +310,86 @@ public struct Interpreter {
...
+    }
+  }
+
+  /// Returns the value occurring in the interpreted program that is carried by
+  /// `self` in the context of `executor`.
```
dabrahams: I don't understand this doc.  In what way is a `Program` carried by an `IRValue`?

Sources/Interpreter/Interpreter.swift
dabrahams: Maybe use `__callAsFunction__` to make 


```suggestion
      executor.topOfStack.registers[r]!(as: Access<Memory.TypedAddress>.self)!
```

work?  Maybe extend `Optional<Whatever>` so that this works?

```suggestion
      executor.topOfStack.registers[r](as: Access<Memory.TypedAddress>.self)!
```

?

Sources/Interpreter/Interpreter.swift
RishabhRD: Yes, I think this is misleading. How about:
```
  /// Returns the value in the interpreted program corresponding to `self` in the
  /// current execution state of `executor`.
```

Sources/Interpreter/Interpreter.swift
```diff
@@ -270,6 +307,87 @@ public struct Interpreter {
...
+  /// and returns its address.
+  ///
+  /// - Precondition: `t` is a monomorphic type.
+  private mutating func allocate(storageFor t: AnyTypeIdentity) -> Memory.TypedAddress {
+    let a = memory.allocate(storageFor: t)
+    topOfStack.allocations.append(a)
+    return a.asTypedAddress(t)
+  }
+
+  /// Stores the value carried by `v` at the location pointed by the address `p`.
```
```suggestion
  /// Stores `v` at the address `p`.
```

Either it's a value in which case that's fine wording, or it's not a value and the type should be renamed.

Sources/Interpreter/Interpreter.swift
```diff
@@ -270,6 +307,87 @@ public struct Interpreter {
...
+      executor.topOfStack.parameters[i]
+    case .register(let r):
+      executor.topOfStack.registers[r]!(as: Access<Memory.TypedAddress>.self)!
+    default:
+      preconditionFailure("\(executor.program.show(self)) is not an Access<Memory.TypedAddress>.")
+    }
+  }
+
+  /// Returns the value in the interpreted program corresponding to `self` in the
+  /// current execution state of `executor`.
```
```suggestion
  /// Returns`self` as represented in `executor`.
```

Sources/Interpreter/Interpreter.swift
```diff
@@ -270,6 +307,87 @@ public struct Interpreter {
...
+  /// - Precondition: `t` is a monomorphic type.
+  private mutating func allocate(storageFor t: AnyTypeIdentity) -> Memory.TypedAddress {
+    let a = memory.allocate(storageFor: t)
+    topOfStack.allocations.append(a)
+    return a.asTypedAddress(t)
+  }
+
+  /// Stores the value carried by `v` at the location pointed by the address `p`.
+  private mutating func store(_ v: IRValue, at p: IRValue) throws {
+    try memory.store(v.asRuntimeValue(in: &self), at: p.asAccess(in: self))
```
dabrahams: Now that I see the correct description, this no longer seems like a cast.  An `IRValue` is a notional index of a value in some execution context, right?  I think this is either `v.value(in: &self)` or `self[v]`.  I like the latter because it hides the `&` as there's no notional mutation here; it's just needed for cache updates (right?)

Sources/Interpreter/RuntimeValue.swift
```diff
@@ -0,0 +1,83 @@
+import BigInt
+import FrontEnd
+import Utilities
+
+/// A value occurring during program execution.
+struct RuntimeValue {
+
+  /// The bytes of value, preceded by zero or more bytes of padding to satisfy
```
dabrahams:
```suggestion
  /// The bytes of the value preceded by zero or more bytes of padding to satisfy
```

Sources/Interpreter/RuntimeValue.swift
```diff
@@ -0,0 +1,83 @@
...
+  /// its alignment.
+  private let storage: [UInt8]
+
+  /// The number of bytes before the value logically begins.
+  private let baseOffset: Int
+
+  /// Creates an instance having bytes representation `bs`, such that the
+  /// address of the first byte is aligned to `a`.
+  public init(bytes bs: [UInt8], havingAlignment a: Int) {
```
dabrahams:
```suggestion
  /// Creates an instance having bytes `bs` and alignment `a`.
  public init(bytes bs: [UInt8], alignment a: Int) {
```

Sources/Interpreter/RuntimeValue.swift
```diff
@@ -0,0 +1,83 @@
...
+  }
+
+  /// Raw bytes of the value.
+  public var bytes: ArraySlice<UInt8> { storage[baseOffset...] }
+}
+
+extension RuntimeValue {
+
+  /// Creates an instance for an integer of size `b` bytes having value `n` and
+  /// alignment `a`.
```
dabrahams: If you measured integer width in bits instead you could write


```suggestion
  /// Creates a `w`-bit integer having value `n` and
  /// alignment `a`.
```

Sources/Interpreter/RuntimeValue.swift
```diff
@@ -0,0 +1,83 @@
...
+
+  default:
+    unreachable()
+  }
+}
+
+/// Returns the unsigned representation of `n` using `b` bytes in two's-complement form.
+///
+/// - Precondition: `n` must be representable as a signed integer in `b` bytes.
+internal func twosComplementRepresentation(_ n: BigInt, size b: Int) -> BigUInt {
```
dabrahams: I wonder if it wouldn't be better to return UInt128 here?  Do we need more width than that?  Would that allow us to drop the `b` parameter?

dabrahams (commented):
comments inline

Sources/Interpreter/Interpreter.swift
RishabhRD: > An IRValue is a notional index of a value in some execution context, right? 

Yes, if you mean any value (like place, access, runtime value) manipulated by IR.

> there's no notional mutation here; it's just needed for cache updates (right?)

Yes.

RishabhRD: Based on this, I’m wondering how `asTypedAddress` is different. An `IRValue` is also a notional index of a `TypedAddress` in an execution context, so perhaps we should move this to Interpreter as well and expose it as a subscript, e.g. `self[location: v]`. WDYT?

dabrahams: All I can say is, "see How the use sites look".  What makes the code more readable?

FWIW IRvalue is like a variable name. I get that as an index to its value. I think of the address of that variable as ancillary information and that's how we usually program: the variable name "decays" to its value and you have to do something extra to get the address.

dabrahams: Suggests `x.address(of: v)`

RishabhRD:
@dabrahams I have resolved all the comments. Do you think, it's fine to move forward with the PR?

dabrahams:
Of course!  Didn't I approve it at some point?

PR #477, opened 2026-08-14
Implement support for keyword identifiers

kyouko-taiga:


Tests/CompilerTests/positive/keyword-identifiers.hylo
```diff
@@ -0,0 +1,13 @@
+//! stage:typing no-std
+
+struct S {
```
tothambrus11, line 3: It would be nice to add a test case asserting that unqualified member accesses cannot be unquoted keywords.

kyouko-taiga: Such a test would boil down to checking that the parser handles an incomplete statement. We have many of them already.

PR #479, opened 2026-08-14
Don't inline functions that failed a mandatory IR pass

kyouko-taiga:
This patch fixes an assertion violation that was due to the inliner attempting to process an ill-formed function.

Inlining is defined for refined IR, meaning that it cannot be applied to a function that failed a mandatory pass. The fix in this patch consists of removing the definitions of ill-formed functions from the program, so that any subsequent pass will simply skip it.

### DIFF
--- a/Sources/FrontEnd/Program.swift
+++ b/Sources/FrontEnd/Program.swift
@@ -111,10 +111,18 @@ public struct Program: Sendable {
     self = emitter.release()
   }
 
-  /// Applies mandatory transformation passes on the IR of `m`.
+  /// Applies mandatory verification and transformation passes on the IR of `m`.
+  ///
+  /// This method applies the checks and transformation necessary to turn *raw* IR produced by the
+  /// emitter into *refined* IR. Functions that successfully go through mandatory passes can be
+  /// considered well-formed and ready to be interpreted or compiled.
+  ///
+  /// If a pass fails, an error diagnostic is reported and the definition of the function in which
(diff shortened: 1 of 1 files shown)


Sources/FrontEnd/Program.swift
```diff
@@ -143,13 +151,9 @@ public struct Program: Sendable {
...
         work[i].function.existentializeIfExposed(emittingInto: m, using: &typer)
-      }
 
-      // Move all functions back.
-      modify(&typer.program[m].ir) { (ir) in
-        while let (i, f) = work.popLast() {
-          ir[i].take(definition: f)
-        }
+        // No error occurred; move the function back.
+        typer.program[m].ir[work[i].id].take(definition: work[i].function)
```
tothambrus11, line 156: Can we add a negative test case that would produce an assertion violation if a function body was present?

tothambrus11 (approved):

Sources/FrontEnd/Program.swift
kyouko-taiga: It is a bit tricky because there is no guarantee that the assertion will fire. Giving an ill-formed function to the inliner is basically a violation of a precondition for which there is currently no run-time test.

tothambrus11: There needs to be no guarantee. It's just a safeguard that shows the bug is fixed, and if a similar bug is introduced again, the test would have some reason to crash.

PR #480, opened 2026-08-14
Use new thread with explicit QoS

tothambrus11:
Sets the QoS of the work we are waiting on to be [userInitiated](https://developer.apple.com/documentation/dispatch/dispatchqos), which is supposedly not lower than our current main thread, so there is no problem of priority inversion. The only QoS higher than `userInitiated` would be `userInteractive` but I don't think that applies in the compiler.


tothambrus11:
This appears to also hang on Windows, I'll reboot and try fixing it there.

PR #458, opened 2026-08-11
Fix stream reading deadlock in Process execution

tothambrus11:
This PR changes the reading of stderr/stdout to happen concurrently to avoid deadlocks.

Fixes #454

PR #482, opened 2026-08-14
Use thread-based stream draining for process output

tothambrus11:
This PR re-applies the thread-based implementation from #458 on top of #473, keeping the test case and the resilient utf-8 decoding.


tothambrus11:
This suffered from the same issue, it just wasn't shown without your test before. I'll diagnose it tomorrow on Windows.

PR #485, opened 2026-08-15
Fix priority inversion 2

tothambrus11:
The solution was a combination of waiting synchronously on process output in dedicated threads, then for everything to finish in the caller in an asynchronous way, without blocking the main thread.

I kept the test case (slightly adjusted to use `/c` on Windows instead of `-c`), and added a test timeout, so test is failed after 3 seconds of running. Process cancellation would be much more tricky to implement, so if we want that, we should adopt https://github.com/swiftlang/swift-subprocess

### DIFF
--- a/Tests/UtilitiesTests/HostTests.swift
+++ b/Tests/UtilitiesTests/HostTests.swift
@@ -21,24 +21,28 @@ final class HostTests: XCTestCase {
   }
 
   #if os(Windows)
-    func testFindBinaryExecutableFindsAndExecutesWhereExe() throws {
+    func testFindBinaryExecutableFindsAndExecutesWhereExe() async throws {
       let whereExe = try Host.findBinaryExecutable(invokedAs: "where")
       XCTAssertEqual(whereExe.lastPathComponent.lowercased(), "where.exe")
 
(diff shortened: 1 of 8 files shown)


Tests/UtilitiesTests/HostTests.swift
```diff
@@ -47,22 +51,21 @@ final class HostTests: XCTestCase {
...
     #endif
 
-    XCTAssertThrowsError(
-      try Process.executionOutput(executable, arguments: arguments), "",
-      { (error) in
-        if let e = error as? Process.NonzeroExit {
-          XCTAssertEqual(e.exitCode, 42)
-          XCTAssertEqual(e.executable, executable)
-          XCTAssertEqual(e.arguments, arguments)
+    do {
```
tothambrus11, line 54: XCTAssertThrowsError doesn't support async on all platforms.

tothambrus11:
@kyouko-taiga Could you verify if this gets rid of the warning on MacOS?

dabrahams:
I get no warning from this on macOS

tothambrus11:
Closing this in favor of https://github.com/hylo-lang/hylo-new/pull/497.

Issue #438, opened 2026-08-08
Research & design int-to-pointer and pointer-to-int semantics


tothambrus11:
Pointers are not always just integer addresses on all platforms. See https://rocm.docs.amd.com/projects/llvm-project/en/latest/LLVM/llvm/html/LangRef.html#non-integral-pointer-type

We should also figure out how we deal with provenance in the presence of integer to pointer conversions. See this article: https://www.ralfj.de/blog/2020/12/14/provenance.html

PR #492, opened 2026-08-16
Stand up support for static strings

kyouko-taiga:
WIP porting code from OFE.


Sources/FrontEnd/Syntax/Builtins/BuiltinFunction.swift
```diff
@@ -1210,14 +1210,13 @@ extension BuiltinFunction {
...
-    //    case "ptrtoint":
-    //      guard let t = machineType(&tokens) else { return nil }
-    //      self = .ptrtoint(t)
-    //
+    case "inttoptr":
+      guard let t = machineType(&tokens) else { return nil }
+      self = .inttoptr(s.demand(t))
+    case "ptrtoint":
+      guard let t = machineType(&tokens) else { return nil }
+      self = .ptrtoint(s.demand(t))
```
tothambrus11, line 1218: Pointer-to-int and int-to-pointer conversions are non-obvious when pointer is not just an address. See https://github.com/hylo-lang/hylo-new/issues/438 These considerations are important for embedded and GPU contexts.

Maybe what we ought to have is something that gets the integer address from a given pointer. I'm also not yet sure how provenance works when going through integers and integer based pointer manipulations.

kyouko-taiga: We're just using LLVM here. We have no control over the semantics of these instructions. Of course we can write better abstractions on top of them but that's not what `Builtin.xxx` ought to be.

tothambrus11: _Maybe_ if we narrow the allowed use cases of an instruction, the interpreter could have an easier time emulating the behavior and protecting against UB. But we can use this, and refactor later if needed.

kyouko-taiga: Meh. I don't want to change the semantics of LLVM by trying to cleverly intercept behaviors. The built-ins have been designed with the intention that they would have *exactly* the semantics of LLVM.

We don't really want to specify 100+ low-level instructions and make sure that our specification stays up to date with whatever LLVM decides. But if we start making changes to some instructions because we think they are confusing (and many are), then that's exactly what we'll have to do. People working on the standard library will need one authoritative source of truth and today that is the LLVM language reference.

Tests/UtilitiesTests/SubstringTests.swift
```diff
@@ -27,4 +27,8 @@ final class SubstringTests: XCTestCase {
     XCTAssertEqual(t, "vdlm")
   }
 
+  func testUnescaped() {
+    XCTAssertEqual("one\\ntwo".unescaped, "one\ntwo")
```
tothambrus11, line 31: I'd recommend using raw string literals on the LHS to make the test more readable.
We should add more test cases, with 1, 2, 3 number of `\`-es in the raw string, and testing different characters such as `\t` and invalid things to unescape, such as `\*`. Maybe unescaping is ought to produce diagnostics on failure.

tothambrus11:
This PR needs to move code to the standard library. I'll try to do that over the weekend.

PR #495, opened 2026-08-16
Fix the detection of missing yields after loops

kyouko-taiga:


Tests/CompilerTests/negative/illegal-yield.hylo
```diff
@@ -21,3 +21,20 @@ subscript f3(x: Bool) -> Bool {
...
   if x { yield true }
 }
+
+// A yield in a loop.
+subscript f4(x: Bool) -> Bool {
+  while x { yield true }
+}
+
+// A yield before a loop is fine.
+susbcript g0(x: Bool) -> Bool {
```
tothambrus11:
```suggestion
subscript g0(x: Bool) -> Bool {
```

Tests/CompilerTests/negative/illegal-yield.hylo
```diff
@@ -21,3 +21,20 @@ subscript f3(x: Bool) -> Bool {
...
+subscript g0(x: Bool) -> Bool {
+  yield true
+  while x {}
+}
+
+// A yield after a loop is fine.
+subscript g1(x: Bool) -> Bool {
+  while x {}
+  yield true
+}
```
tothambrus11, line 41: Should we move these to a positive test? Do we have a guarantee that if there is an error in a function/subscript body, it would have been reported for sure? Even if we want to maintain such guarantee, a test case probably shouldn't rely on that, but test it.

kyouko-taiga: It is relevant to test this behavior here as code paths are adjacent. As I said before, it is not possible to perfectly categorize every test case. And yes, there is a guarantee that if something goes wrong a diagnostic must be reported.

tothambrus11 (approved):

PR #497, opened 2026-08-16
Adopt Swift Subprocess

tothambrus11:
This PR replaces our attempts at making `Process` work with the robust library https://github.com/swiftlang/swift-subprocess.


Sources/Driver/Driver.swift
```diff
@@ -259,18 +261,17 @@ public struct Driver {
...
-    let o = destinationDirectory.appendingPathComponent("\(uniquePrefix)-\(fileName))",
+    let o = destinationDirectory.appendingPathComponent("\(uniquePrefix)-\(fileName)",
       isDirectory: false)
 
     var a = ["-c", source.path, "-o", o.path]
     if let r = relocation.asClangArgument { a.append(r) }
 
-    _ = try Process.executionOutput(
-      try Host.findBinaryExecutable(invokedAs: "clang"), arguments: a)
+    _ = try await subprocessOutput(of: .name("clang"), arguments: a)
```
tothambrus11, line 274: This is particularly amazing. Swift Subprocess implemented the logic to invoke clang.bat / clang.exe / clang.cmd, whatever is available first on PATH.

.github/workflows/build-and-test.yml
```diff
@@ -20,7 +20,10 @@ jobs:
       - uses: actions/checkout@v7
 
       - name: Check max line length of 100 characters
-        run: "! grep -rnE --include=*.swift --exclude-dir=cmake.* --exclude-dir=.build '.{101,}' ."
+        # Lines ending with "[allow long]" are exempt.
```
dabrahams, line 23: This new policy/convention should be discussed I think

Sources/Driver/Driver.swift
dabrahams: I hope they got it right; that isn't easy to [do well](https://github.com/hylo-lang/gyb-swift/blob/5fe38a2ec40911c3e9a951ed405dbe7ee706ebe0/Sources/gyb-swift/ProcessUtilities.swift#L67-L86).

Sources/Driver/Driver.swift
```diff
@@ -368,15 +369,15 @@ public struct Driver {
...
     arguments += librarySearchPaths.map({ "-L\($0.path)" })
     arguments += librariesToLink.map({ "-l\($0)" })
     arguments += objectFiles.map(\.path)
 
     #if os(macOS)
-    let xcrun = try Host.findBinaryExecutable(invokedAs: "xcrun")
-    let sdk = try Process.executionOutput(xcrun, arguments: ["--sdk", "macosx", "--show-sdk-path"])
+    let sdk = try await subprocessOutput(
+      of: .name("xcrun"), arguments: ["--sdk", "macosx", "--show-sdk-path"])
       .trimmingCharacters(in: .whitespacesAndNewlines)
```
dabrahams, line 381: What a silly name!  Newlines are whitespace aren't they?

Sources/Utilities/Host.swift
```diff
@@ -53,43 +53,7 @@ public enum Host: Sendable {
...
-    Host.environment["PATH", default: ""]
-      .split(separator: Host.searchPathSeparator)
-      .map(String.init)
-  }
-
-  /// The separator between individual locations in executable search path.
-  public static let searchPathSeparator: Character = Host.operatingSystem == .windows ? ";" : ":"
-
   /// The suffix of binary executables.
   public static let binaryExecutableSuffix = operatingSystem == .windows ? ".exe" : ""
```
dabrahams, line 57: Do we still need this?

dabrahams (approved):
LGTM!

Sources/Driver/Driver.swift
tothambrus11: Maybe they are not white but transparent...

Sources/Utilities/Host.swift
tothambrus11: Yes, when producing binaries ourselves.

Sources/Driver/Driver.swift
tothambrus11: They reimplemented the essential behavior of where.exe: 
- https://github.com/swiftlang/swift-subprocess/blob/main/Sources/Subprocess/Platforms/Subprocess%2BWindows.swift#L816
- https://github.com/swiftlang/swift-subprocess/blob/main/Sources/Subprocess/Platforms/Subprocess%2BWindows.swift#L848

dabrahams: ...which is why I say I hope they got it right...

Issue #502, opened 2026-08-17
Storing the result of a sink subscript


tothambrus11:
This should work, but currently a wrong diagnostic is emitted:
```hylo
//! exit-status:42

/// A `sink` subscript projects an owned value that the caller may move into its own storage.
subscript id(x: sink Int32) sink -> Int32 {
  yield &x
}

fun main() -> Int32 {
  var y: Int32 = 42
  var a = id[y]
  return a
}
```
```
positive/sink-subscript.hylo:11.10-11: error: illegal move
  return a
         ^
positive/sink-subscript.hylo:11.10-11: error: use of uninitialized object
  return a
         ^
```

PR #504, opened 2026-08-17
Expose size and alignment computation

tothambrus11:
The size is computed by sorting the members by decreasing alignment, then calling the existing `size_of_record_having_ordered_members` function. Alternatively, if we can assume that all members' offsets were queried, we can know that the array is already sorted, but I assume that's not the case in general.

In the future we may want to optimize cases when we access multiple members - then we could do a single insertion sort pass rather than multiple. Once we accessed the size of the record, we also know that the array is already sorted, so we could query the offsets without resorting.

The alignment is simply the max alignment of all members. 

I extended the test case and generator to have an expected alignment and size.

TODO: convert to use LCM


Tests/CompilerTests/positive/runtime-offset-of-member.package/runtime-offset-of-member.hylo
```diff
@@ -87,19 +97,25 @@ extension TypeWitnessHeader {
...
 type H = TypeWitnessHeader
 
+/// A witness that header tuples can be destroyed, so that they may be bound to locals.
+given TypeWitnessHeader is Deinitializable {}
+
+/// Returns an array of pointers to the elements of `headers`.
+fun witnesses(of headers: H[10]) -> Pointer<TypeWitnessHeader>[10] {
+  let p = Pointer<H>(value: Builtin.address(of: headers))
+  return (p + 0, p + 1, p + 2, p + 3, p + 4, p + 5, p + 6, p + 7, p + 8, p + 9)
+}
```
tothambrus11, line 107: This should really be a sink subscript after #502  is fixed. Also, we can't find a `Movable` conformance for this many tuple elements at the moment.

StandardLibrary/Sources/Core/Runtime/OffsetOfMember.hylo
```diff
@@ -548,3 +554,51 @@ public fun __hylo_offset_of_member(
...
+@inline(always)
+public fun __hylo_alignment_of_record(
+  in member_array: Pointer<RecordMemberType>, of_length member_count: UInt32
+) -> Alignment {
+  var largest = 1 as Alignment
+  var p = member_array.copy()
+  let end = member_array + member_count
+  while p < end {
+    sink let alignment = p.pointee().alignment()
+    if alignment > largest { &largest = alignment }
```
dabrahams, line 600: This is wrong unless you are guaranteed alignments are powers of some other number.  Otherwise you need the LCM.

dabrahams: This tells me your Swift tests are not sufficient.  For example, mine checks to see that each element is properly aligned with respect to the beginning of the record.  You should check that they are also properly aligned with respect to the beginning of the record +/- the record's alignment.

tothambrus11: Yeah, I just copied the behavior of the LLVM lowering, which makes the assumption that alignments are power of two. I can change both to LCM and strengthen the test suite.

dabrahams:
> if we can assume that all members' offsets were queried, we can know that the array is already sorted

No we can't.  The offset calculation destroys the contents of the array rather than simply leaving it sorted.  A new array is initialized for each offset calculation.

PR #505, opened 2026-08-17
Fix storing the result of a sink subscript

### LINKED ISSUE #502: Storing the result of a sink subscript

tothambrus11:
Fixes #502. The problem was that subscripts don't store into their `target` (result) place, and instead just yield the value through the register that's returned by `lower(call:output:)`. The fix involves emitting a non-reified move from the yielded place to the store `target` (the move can be later reified into either set/inout move).

### DIFF
--- a/Sources/FrontEnd/IR/IREmitter.swift
+++ b/Sources/FrontEnd/IR/IREmitter.swift
@@ -747,10 +747,11 @@ internal struct IREmitter {
     }
 
     // Are we lowering a subscript application?
-    else if program[e].style == .bracketed {
-      // Move the projected result into `target`.
+    else {
       let y = lower(lvalue: e)
-      lowering(e, { $0._emitMove([.inout, .set], y, to: target) })
+      if program[e].style == .bracketed {
+        lowering(e, { $0._emitMove([.inout, .set], y, to: target) })
+      }
     }
 
     // Are we lowering an ordinary call?
--- a/Sources/FrontEnd/IR/IREmitter.swift
+++ b/Sources/FrontEnd/IR/IREmitter.swift
@@ -746,6 +746,13 @@ internal struct IREmitter {
       lower(memberwiseInitialization: e, of: target)
     }
 
+    // Are we lowering a subscript application?
+    else if program[e].style == .bracketed {
+      // Move the projected result into `target`.
+      let y = lower(lvalue: e)
+      lowering(e, { $0._emitMove([.inout, .set], y, to: target) })
+    }
+
     // Are we lowering an ordinary call?
     else {
       lower(call: e, output: target)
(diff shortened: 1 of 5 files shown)


Sources/FrontEnd/IR/IREmitter.swift
```diff
@@ -746,6 +746,13 @@ internal struct IREmitter {
       lower(memberwiseInitialization: e, of: target)
     }
 
+    // Are we lowering a subscript application?
+    else if program[e].style == .bracketed {
+      // Move the projected result into `target`.
+      let y = lower(lvalue: e)
+      lowering(e, { $0._emitMove([.inout, .set], y, to: target) })
+    }
```
kyouko-taiga, line 753:
```suggestion
    else {
      let y = lower(lvalue: e)
      if program[e].style == .bracketed {
        lowering(e, { $0._emitMove([.inout, .set], y, to: target) })
      }
    }
```
Then you don't need the last else branch.

Tests/CompilerTests/positive/sink-subscript.hylo
```diff
@@ -0,0 +1,32 @@
+/// A `sink` subscript projects an owned value that the caller *may* move into its own storage.
```
kyouko-taiga: The comment is misleading. The caller may not move into different storage but it has to consume the value, no matter what.

kyouko-taiga (approved):
Nice work.

Don't know what's causing the CI to fail; it looks a bit odd.

Tests/CompilerTests/positive/sink-subscript.hylo
tothambrus11: Thanks, that's a cleaner way to think about it. What do you think about the following comment? I tried to make it a bit explanatory about how things work, and what is tested.

```
/// A `sink` subscript projects a value that the caller must consume,
/// e.g. by move assignment to fresh storage or by implicit deinitialization.
```

Sources/FrontEnd/IR/IREmitter.swift
tothambrus11: Oh that doesn't work because we need to pass `target` to the lowering of a regular function call. That was also the problem with my last-minute change that made the CI fail.

Tests/CompilerTests/positive/sink-subscript.hylo
tothambrus11:
```suggestion
/// A `sink` subscript projects a value that the caller must consume,
/// e.g. by move assignment to fresh storage or by implicit deinitialization.
```

Sources/FrontEnd/IR/IREmitter.swift
kyouko-taiga: Ah, I see. Sorry for the buggy suggestion.

Tests/CompilerTests/positive/sink-subscript.hylo
kyouko-taiga: WFM

PR #510, opened 2026-08-19
Interpreter: support conditionals

RishabhRD:


Sources/Interpreter/Interpreter.swift
```diff
@@ -249,13 +249,19 @@ public struct Interpreter {
       // TODO: add a real implementation, updating state of composed regions.
       return initializeRegister(to: ())
     case let x as IRBranch:
-      _ = x
+      return .jump(to: entryPoint(x.target))
```
dabrahams: Is  entryPoint the right word here?  @kyouko-taiga I thought an entry point was always a function..

Sources/Interpreter/Program+Extensions.swift
```diff
@@ -0,0 +1,30 @@
+import FrontEnd
+
+extension Program {
+
+  /// Returns the tag of `t`.
```
dabrahams, line 5: Curious what the tag of a type is?

Sources/Interpreter/Program+Extensions.swift
```diff
@@ -0,0 +1,30 @@
...
+  /// Returns the underlying type of `t`, defined in `p`, after unwrapping
+  /// type applications and aliases.
+  internal func underlyingType(_ t: AnyTypeIdentity) -> AnyTypeIdentity {
+    let u = tag(t)
+    if u == TypeApplication.self {
+      let a = type(t, as: TypeApplication.self)
+      return underlyingType(a.abstraction)
+    } else if u == TypeAlias.self {
+      let a = type(t, as: TypeAlias.self)
+      return underlyingType(a.aliasee)
```
dabrahams: Would an alias to an alias need further unwrappings?

Tests/InterpreterTests/InterpreterTestPrograms/Conditionals.hylo
```diff
@@ -0,0 +1,15 @@
...
+    &r = 1
+  } else {
+    &r = 2
+  }
+
+  if false {
+    &r = 3
+  } else {
+    &r = 4
+  }
```
dabrahams, line 17: Not a very good test for execution because the 2nd conditional unconditionally masks whether the first one worked or not and there's no way to observe that the 2nd one worked either.  Are result codes in these tests something we don't have the infrastructure in place for yet?

dabrahams (approved):

Tests/InterpreterTests/InterpreterTestPrograms/Conditionals.hylo
RishabhRD: We don't have infrastructure for it yet. We need support for function calls and builtin function calls (the basics not all: like trap and equality comparison), to support checking the operation result. This would allow us to call `precondition` function.

I have planned it to support these immediately after this as then we can write more confident interpreter run tests.

RishabhRD: I will update these tests to call `precondition` after that.

Sources/Interpreter/Interpreter.swift
RishabhRD: I remember we use `entry` specifically to denote basic blocks. I renamed it to `start(x.target)`. Does this read better?

Sources/Interpreter/Program+Extensions.swift
RishabhRD: It is the runtime information about the "type" of `t` in the frontend. For example, `t` may be a `MachineType` or an `Arrow`.

Tests/InterpreterTests/InterpreterTestPrograms/Conditionals.hylo
tothambrus11: Preconditions will be nice addition. At some point we could also add infrastructure to call functions directly from Swift and inspect their values:

```
var p = InterpretedProgram("""
  fun f(x: Int, label y: Int) {
    x + y
  }
  """

let result = p.lvalue_call("f", [(nil, .std.int(2)), ("label", .std.int(3))])
XCTAssertEqual(result.value, .std.int(3))
```

RishabhRD: @tothambrus11 This takes some effort, but I am noting it for future.

RishabhRD: Added todos for the weak tests for now.

Sources/Interpreter/Program+Extensions.swift
RishabhRD: Realized I can use `dealised` from typer, that is a more robust implementation and allows us to get rid of recursion too.

RishabhRD:
Merging the PR, as all comments have been resolved.

PR #520, opened 2026-08-22
Interpreter: Support Builtin.trap() function

RishabhRD:
Because trap never completes successfully, I separated the program execution tests into success and failure cases.


RishabhRD:
IDK what I can do for test coverage CI here. The missing lines are:
1. `Trap.description`
2. `return initializeRegister(to: v)` for `ApplyBuiltin` instruction, which doesn't happen for trap and other builtin functions are not supported yet.

dabrahams:
I guess don't worry about it?

Sources/Interpreter/Interpreter.swift
```diff
@@ -244,7 +244,8 @@ public struct Interpreter {
     case let x as IRApply:
       _ = x
     case let x as IRApplyBuiltin:
-      _ = x
+      let v = try call(x.callee, with: x.arguments)
```
```suggestion
      let v = try call(x.callee, passing: x.arguments)
```

Sources/Interpreter/Interpreter.swift
```diff
@@ -395,6 +396,14 @@ public struct Interpreter {
     return .init(i, in: programCounter.container)
   }
 
+  /// Returns the result of calling `f` with `arguments`.
+  private func call(_ f: BuiltinFunction, with arguments: [IRValue]) throws -> RuntimeValue {
```
dabrahams: “with” is a weak argument label; has too many interpretations.

```suggestion
  private func call(_ f: BuiltinFunction, passing arguments: [IRValue]) throws -> RuntimeValue {
```

Sources/Interpreter/Interpreter.swift
```diff
@@ -439,5 +448,20 @@ extension IRAccess {
 
 }
 
+extension Interpreter {
+
+  /// A trap occurred during program execution.
+  public struct Trap: Error, Regular, CustomStringConvertible {
```
dabrahams: You might want a protocol for all interpreter exceptions

Sources/Interpreter/Interpreter.swift
```diff
@@ -439,5 +448,20 @@ extension IRAccess {
...
+
+    public init() {}
+
+    public var description: String {
+      "Encountered a trap during program execution."
```
dabrahams: Seems a bit wordy.  All errors are “during program execution”  Can we do better than `Intepreter.Trap` which I think would be printed by default?

Tests/InterpreterTests/Utilities.swift
```diff
@@ -43,3 +43,18 @@ public func check<E: Error & Equatable, R>(
     XCTAssertEqual($0 as? E, expectedError, "\($0)", file: file, line: line)
   }
 }
+
+/// Executes `action` and reports test failure if it does not throw `error`.
+public func check<E: Error & Equatable, R>(
```
dabrahams, line 48: Isn't [this](https://developer.apple.com/documentation/xctest/error-assertions) already what you need?

dabrahams (approved):
comments inline

Tests/InterpreterTests/Utilities.swift
RishabhRD: I couldn't make it work for async `action` using `XCTAssertThrowsError`. I was getting the error:
```
'async' call in an autoclosure that does not support concurrency
```

Sources/Interpreter/Interpreter.swift
RishabhRD: > Can we do better than Intepreter.Trap which I think would be printed by default?

I don't think we can. I think I can thus remove `CustomStringConvertible` here.

Tests/InterpreterTests/Utilities.swift
tothambrus11: This is coming up quite often. This is the core issue: https://github.com/swiftlang/swift-corelibs-xctest/issues/543

RishabhRD:
I think this can be merged as `XCTAssertThrowsError` async overload is not there, that resolves all the comments.

Tests/InterpreterTests/Utilities.swift
dabrahams: Maybe we should name our thing like the XCTest facility.

PR #524, opened 2026-08-23
Interpreter: make enum layout consistent with llvm

### LINKED ISSUE #514: Misordered fields in LLVM backend's enum layout
The interpreter calls `storageLayoutOfRecord([payload, discriminator])` ([TypeLayoutCache.swift:103-105](https://github.com/hylo-lang/hylo-new/blob/main/Sources/Interpreter/TypeLayoutCache.swift#L103-L105)), which sorts by decreasing alignment and therefore most likely puts the discriminator last. This is inconsistent with LLVM lowering that requires the discriminator to be the first element. Not sure if this is a problem, but it would be probably better to have it consistent.

RishabhRD:
Fixes #514 

Previously, we were computing enum layout similar to runtime layout algorithm. But #514 suggests that's not the case with enums while LLVM lowering.

### DIFF
--- a/Sources/Interpreter/TypeLayoutCache.swift
+++ b/Sources/Interpreter/TypeLayoutCache.swift
@@ -94,22 +94,23 @@ struct TypeLayoutCache {
       layout(.init(c), in: &p)
     }
 
-    let d = layout(abi.enumDiscriminator(count: cases.count, in: &p), in: &p)
+    let discriminator = abi.enumDiscriminator(count: cases.count, in: &p)
+    let discriminatorLayout = layout(discriminator, in: &p).whole
 
-    let payload = TypeLayout.Bytes(
+    let payloadLayout = TypeLayout.Bytes(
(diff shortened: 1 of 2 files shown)


tothambrus11 (approved):
Thanks, LGTM!

dabrahams (changes requested):
I think the issue is wrong and LLVM lowering should be changed

tothambrus11:
Thanks for the work! According to the discussions in #514 , I think it's worth to in fact change LLVM lowering to have a consistent behavior with what's been done on the interpreter side. @RishabhRD Can we close this PR?

dabrahams:
I am not clear on whether what I've been saying about using the layout algorithm has been ignored or not.  It would be good to know it's been properly considered before we close this.

tothambrus11:
From what I understood, we agree that we could use the record layout algorithm to lay out the payload and discriminator

RishabhRD:
Closing the PR, as we agreed that LLVM lowering side should be consistent with runtime layout algorithm.

PR #527, opened 2026-08-24
Interpreter: Implement integer comparisons

RishabhRD:
This PR adds a step in direction so that we can write `precondition(condition)` kind of expression in program running tests, and thus tests would not be just some weak tests.


Sources/Interpreter/IntegerPredicate+Extensions.swift
```diff
@@ -0,0 +1,105 @@
+import FrontEnd
+
+extension IntegerPredicate {
+
+  /// Returns whether `lhs` satisfies `self` with respect to `rhs`.
+  public func callAsFunction(_ lhs: UInt8, _ rhs: UInt8) -> Bool {
```
dabrahams: Write generic functions over one of the integer protocols instead.

Sources/Interpreter/Interpreter.swift
```diff
@@ -420,6 +449,19 @@ extension IRValue {
     return executor.memory.location(p, in: a)
   }
 
+  /// Returns the address of field `f` in the address pointed by `self`, in
```
```suggestion
  /// Returns the address of field `f` in `self`, in
```

Sources/Interpreter/Memory.swift
```diff
@@ -298,13 +298,25 @@ extension Memory {
...
+  }
+
   /// Returns the address of `subPart` in `whole`.
   public mutating func location(_ subPart: IndexPath, in whole: TypedAddress) -> TypedAddress {
     let (t, o) =
       typeLayouts.typeAndOffset(subPart, within: whole.type, definedIn: &program)
     return .init(allocation: whole.allocation, offset: o + whole.offset, type: t)
   }
 
+  /// Returns the address of `f` in `whole`.
```
dabrahams: Make up for weak type information.

```suggestion
  /// Returns the address of the field `f` in `whole`.
```

Sources/Interpreter/RuntimeValue.swift
```diff
@@ -39,6 +39,14 @@ extension RuntimeValue {
     self.init(bytes: byteRepresentation(r, bitWidth: w), havingAlignment: a)
   }
 
+  /// Creates an instance of `MachineType.i(1)` having value `b` and layout `l`.
+  public init(bool b: Bool, havingLayout l: TypeLayout.Bytes) {
```
dabrahams, line 43: It's very weird to say we know its an `i1` and at the same time get a dynamic layout parameter.  Are you sure that's necessary?

Sources/Interpreter/RuntimeValue.swift
```diff
@@ -59,6 +67,43 @@ extension RuntimeValue {
     withUnsafePointer(to: UInt8.self) { $0.pointee != 0 }
   }
 
+  /// The 8-bit unsigned value.
+  ///
+  /// - Precondition: `self` is an instance of `MachineType.i(8)`.
+  public var i8: UInt8 {
```
dabrahams: I think it needs `as`.  We could have `self.asI8` or we could have `self.as(UInt8.self)`.  Former probably better.

Sources/Interpreter/TypeLayoutCache.swift
```diff
@@ -209,4 +209,17 @@ extension TypeLayoutCache {
       c.type = p.type
     }
   }
+
+  /// Returns the type and byte offset of `f` within `t`,
```
dabrahams:
```suggestion
  /// Returns the type and byte offset of  the field `f` within `t`,
```

Sources/Interpreter/TypeLayoutCache.swift
```diff
@@ -209,4 +209,17 @@ extension TypeLayoutCache {
...
+  /// where `t` is defined in `p`.
+  ///
+  /// - Precondition: There exists a field `f` in `t`.
```
dabrahams: Arguably unneeded.  Doesn't hurt.

Tests/InterpreterTests/TypeLayoutTests.swift
```diff
@@ -486,4 +486,29 @@ final class TypeLayoutTests: XCTestCase {
...
+          let y: Builtin.i16
+        }
+        """
+    )
+    let i8 = id(MachineType.i(8))
+    let i16 = id(MachineType.i(16))
+
+    var m = Memory(forRunning: p, on: UnrealABI())
+    let a = m.allocate(storageFor: t).asTypedAddress(t)
+
```
dabrahams, line 505: You should assert that i16 is more aligned than i8 or this test is invalid.

dabrahams (approved):

Sources/Interpreter/RuntimeValue.swift
RishabhRD: Actually I didn't want to. I used dynamic layout parameter because our current `TargetABI` protocol allows types such as `MachineType.i(1)` to have a target-dependent layout:

```swift
/// Types that describe the ABI for which we might interpret code.
protocol TargetABI {

  /// Returns the layout of `t`.
  func layout(_ t: MachineType) -> TypeLayout.Bytes

}
```

However, I believe this abstraction is broader than necessary. Most machine types have a well-known platform-independent size.

IMO a better abstraction would be:
```swift
/// Types that describe the ABI for which we might interpret code.
protocol TargetABI {

  /// Returns the ABI-required alignment of `type`.
  ///
  /// The alignment of `.i(1)` and `.i(8)` is required to be one byte.
  func alignment(_ type: MachineType) -> Int

  /// The size of a pointer.
  var pointerSize: Int { get }

}
```

This would allow us to write:
```
public init(bool b: Bool) {
```

tothambrus11:
```
  /// Returns the layout of `t`.
  func layout(_ t: MachineType) -> TypeLayout.Bytes
```
FWIW this function returns bytes while promising to return the layout. I'd expect it to be called `sizeOf(_:)`. 

I think it's good if the layout-related queries are abstracted away. Hardcoding the size of a type here would seem like doing something at the wrong level of abstraction. I would either keep it as it is or add a call to a generic size calculating function.

RishabhRD: > FWIW this function returns bytes while promising to return the layout. 

No, that's not true. `TypeLayout.Bytes` have byte size + alignment. So, it is a layout. Maybe the confusion was due to inaccuracy in return types of my suggested code (copy-pasted TypeLayout.Bytes from existing code :sweat_smile:). Fixed that above.

> Hardcoding the size of a type here would seem like doing something at the wrong level of abstraction.

Is it regarding the code suggestion I gave above?

tothambrus11: > Is it regarding the code suggestion I gave above?

Yeah, partially.

We should also decide if the TargetABI should really be a protocol or it's just a data structure, on which the layout algorithm operates uniformly. For now I would defer this to another PR, as it's outside of the scope here.

RishabhRD: Merging #532 solves this comment.

RishabhRD:
Merging the PR as all the comments are now resolved.

PR #530, opened 2026-08-28
Implement generic witness tables

kyouko-taiga:
This patch implements generic witness tables.

It is probably too large to be reviewed; my apologies. Instead I propose that we walk through the implementation during a team meeting.

### DIFF
--- a/Docs/Depolmorphization.md
+++ b/Docs/Depolmorphization.md
@@ -8,7 +8,7 @@ Both monomorphization and existentialization essentially consist of creating a c
 In the former case, type arguments are simply substituted for their corresponding parameters.
 In the latter case, the type parameters are transformed into term parameters accepting type witnesses at run-time.
 
-This document describes how Hylo implements depolymorphization.
+The rest of this document describes how Hylo implements existentialization.
 
 > Disclaimer:
 > The syntax of Hylo IR is far from final.
@@ -16,7 +16,7 @@ This document describes how Hylo implements depolymorphization.
 
 ## Existentialization
 
-In a nutshell, existentialization consists of replacing type parameters with term parameters while abstracting over any information that cannot be determined at compile-time.
+Existentialization replaces type parameters with term parameters that provide any information that cannot be determined at compile-time.
 Consider the following example to illustrate this concept:
 
 ```hylo
@@ -26,8 +26,7 @@ fun swap<T is Movable>(a: inout T, b: inout T) {
 ```
 
 The first step to compile this function is to lower it to Hylo IR.
-Recall a conformance constraint such as `T is Movable` is always compiled to term parameter accepting a conformance witness.
-As a result, the signature of the lowered form of `swap(_:_:)` is as follows:
+Because the IR cannot represent constraints directly, Hylo replaces each constraint such as `T is Movable` with a conformance witness parameter:
 
 ```hylo
 fun swap(_:_:)<T>(
@@ -36,11 +35,11 @@ fun swap(_:_:)<T>(
 ```
 
 The function is still polymorphic at this stage, taking a type parameter `T`.
-`%p0` accepts a witness of `T`'s conformance to movable.
+`%p0` accepts a witness of `T`'s conformance to `Movable`.
 `%p1` and `%p2`, correspond to `a` and `b`, respectively.
 Finally, `%p3` denotes the output register of the function, i.e., the place to which its result is written.
 
-Existentialization will produce a copy of this function in which the type parameter `T` is replaced with a term parameters:
+Existentialization then replaces the unconstrained type parameter `T` with a term parameter, `T`'s **type witness**:
 
 ```hylo
 fun swap(_:_:).existentialized(
@@ -48,17 +47,16 @@ fun swap(_:_:).existentialized(
 )
 ```
 
-Rather than taking a type parameter, this function accepts a *type witness* as its first parameter `%p0`.
-This type witness will be used to obtain information about the layout of `T`, as we shall see later.
+The type witness supplies information about `T` not given by conformances, such as its layout.
 The following parameters are the same as those of the polymorphic version, just renamed.
 Finally, occurrences of `T` have been replaced with `?0`, which denotes a *skolem* representing some type unknown at compile time.
 
-> From a more formal perspective, existentialization relies on the observation that a term of type `∀⍺.τ`, where `⍺` is a type variable that may occur in `τ`, can be transformed into a term of type `∃⍺.ω → τ` in which the extra parameter witnesses `⍺`'s properties at runtime.
-> One can then [skolemize](https://en.wikipedia.org/wiki/Skolem_normal_form) existentially quantified variable to get rid of the quantifier, resulting in a term of type `ω → τ` in which each occurrence of `⍺` has been replaced with a skolem (aka rigid variable).
+> Formally, existentialization transforms a term of type `∀⍺.τ`, where `⍺` is a type variable that may occur in `τ`, into a term of type `∃⍺.ω → τ` in which the extra parameter witnesses `⍺`'s properties at runtime.
+> It then [skolemizes](https://en.wikipedia.org/wiki/Skolem_normal_form) the existentially quantified variable to get rid of the quantifier, resulting in a term of type `ω → τ` in which each occurrence of `⍺` has been replaced with a skolem (aka rigid variable).
 >
-> For example, given a function of type `∀⍺.⍺ → ⍺`, we first construct a function `∃⍺.ω → ⍺ → ⍺` before applying skolemization to replace occurrences of `⍺` by some unique skolem `κ` and end up with a term of type `ω → κ → κ`, which is a monomorphic function.
+> For example, given a function of type `∀⍺.⍺ → ⍺`, we construct a function `∃⍺.ω → ⍺ → ⍺`, we then skolemize it, replacing occurrences of `⍺` by with a unique skolem `κ`, and end up with a monomorphic function of type `ω → κ → κ`.
 >
-> Note that nothing in the type system that links `ω` to neither `⍺` nor `κ`.
+> Note that nothing in the type system links `ω` to either `⍺` or `κ`.
 > As a consequence, the compiler cannot verify the type safety of a function after it has been existentialized.
 > This issue is inconvenient but does not invalidate the safety of the source language, since existentialization preserves semantics.
 > Hence, a showing the well-typedness of a polymorphic function provides guarantees about the behavior of its existentialized form, at least in principle.
--- a/Docs/Depolmorphization.md
+++ b/Docs/Depolmorphization.md
@@ -1,4 +1,4 @@
-# Deploymorphization
+# Depolymorphization
 
 A function (or type) is polymorphic iff it accepts type parameters and monomorphic otherwise.
 The implementation (aka definition) of a monomorphic function may feature uses of polymorphic constructs that have to be either *monomorphized* or *existentialized* before IR can be compiled to machine code.
--- a/Docs/Depolmorphization.md
+++ b/Docs/Depolmorphization.md
@@ -1,9 +1,395 @@
 # Deploymorphization
 
 A function (or type) is polymorphic iff it accepts type parameters and monomorphic otherwise.
-The implementation (aka definition) of a monomorphic function may feature uses of polymorphic constructs that have to be either monomorphized or existentialized before IR can be compiled
-to machine code.
-This transformation is referred to as *depolymorphization*.
+The implementation (aka definition) of a monomorphic function may feature uses of polymorphic constructs that have to be either *monomorphized* or *existentialized* before IR can be compiled to machine code.
+These transformations are referred to as *depolymorphization*.
 
 Both monomorphization and existentialization essentially consist of creating a copy of the polymorphic function.
-In the former case, type arguments are simply substituted for their corresponding parameters. In the latter case, the type parameters are transformed into term parameters accepting type witnesses at run-time.
+In the former case, type arguments are simply substituted for their corresponding parameters.
+In the latter case, the type parameters are transformed into term parameters accepting type witnesses at run-time.
+
+This document describes how Hylo implements depolymorphization.
+
+> Disclaimer:
+> The syntax of Hylo IR is far from final.
+> Snippets are intended as illustration rather than specification.
+
+## Existentialization
+
+In a nutshell, existentialization consists of replacing type parameters with term parameters while abstracting over any information that cannot be determined at compile-time.
+Consider the following example to illustrate this concept:
+
+```hylo
+fun swap<T is Movable>(a: inout T, b: inout T) {
+  var x = a ; a = b ; b = x
+}
+```
+
+The first step to compile this function is to lower it to Hylo IR.
+Recall a conformance constraint such as `T is Movable` is always compiled to term parameter accepting a conformance witness.
+As a result, the signature of the lowered form of `swap(_:_:)` is as follows:
+
+```hylo
+fun swap(_:_:)<T>(
+  let %p0: Movable<T>, inout %p1: T, inout %p2: T, set %p3: Void
+)
+```
+
+The function is still polymorphic at this stage, taking a type parameter `T`.
+`%p0` accepts a witness of `T`'s conformance to movable.
+`%p1` and `%p2`, correspond to `a` and `b`, respectively.
+Finally, `%p3` denotes the output register of the function, i.e., the place to which its result is written.
+
+Existentialization will produce a copy of this function in which the type parameter `T` is replaced with a term parameters:
+
+```hylo
+fun swap(_:_:).existentialized(
+  let %p0: Type, let %p1: Movable<?0>, inout %p2: ?0, inout %p3: ?0, set %p4: Void
+)
+```
+
+Rather than taking a type parameter, this function accepts a *type witness* as its first parameter `%p0`.
+This type witness will be used to obtain information about the layout of `T`, as we shall see later.
+The following parameters are the same as those of the polymorphic version, just renamed.
+Finally, occurrences of `T` have been replaced with `?0`, which denotes a *skolem* representing some type unknown at compile time.
+
+> From a more formal perspective, existentialization relies on the observation that a term of type `∀⍺.τ`, where `⍺` is a type variable that may occur in `τ`, can be transformed into a term of type `∃⍺.ω → τ` in which the extra parameter witnesses `⍺`'s properties at runtime.
+> One can then [skolemize](https://en.wikipedia.org/wiki/Skolem_normal_form) existentially quantified variable to get rid of the quantifier, resulting in a term of type `ω → τ` in which each occurrence of `⍺` has been replaced with a skolem (aka rigid variable).
+>
+> For example, given a function of type `∀⍺.⍺ → ⍺`, we first construct a function `∃⍺.ω → ⍺ → ⍺` before applying skolemization to replace occurrences of `⍺` by some unique skolem `κ` and end up with a term of type `ω → κ → κ`, which is a monomorphic function.
+>
+> Note that nothing in the type system that links `ω` to neither `⍺` nor `κ`.
+> As a consequence, the compiler cannot verify the type safety of a function after it has been existentialized.
+> This issue is inconvenient but does not invalidate the safety of the source language, since existentialization preserves semantics.
+> Hence, a showing the well-typedness of a polymorphic function provides guarantees about the behavior of its existentialized form, at least in principle.
+> No formal theorem has been proven at the time of this writing.
+
+### Existentializing definitions
+
+Notice that the existentialized form of `swap(_:_:)` is monomorphic, as it no longer accepts any type argument.
+As a result, it can be compiled to LLVM and ultimately machine code.
+Naturally, the definition of the function should also be modified to eliminate uses of the type parameter.
+The details of this transformation are discussed below, using `swap(_:_:)` as a running example.
+
+> From a formal perspective, the mechanisms discussed below describe to the transformations applied to a polymorphic term of type `∀⍺.τ` to produce an monomorphic term of type `ω → τ[⍺ ↦ κ]`.
+
+#### Stack allocations
+
+Storage allocated on the stack is represented by `alloca` instructions in the IR.
+If the type of the allocated storage is concrete and not hidden behind a resilience boundary, the compiler can compute the size and alignment of the allocation at compile-time.
+Otherwise, the size of the allocation is read from a type witness at run-time.
+
+To illustrate, consider the allocation of `x` in the original polymorphic definition of `swap`, which is lowered to the following alloca:
+
+```hylo
+%r0 = alloca T, #preferred
+```
+
+As the function's definition gets existentialized, the compiler will conclude that `?0`'s layout is abstract and thus transform the allocation so that it uses the corresponding type witness instead.
+In our running example, recall that this witness is passed to `%p0`:
+
+```hylo
+%r1 = access [let] %p0
+%r0 = alloca %r1 as ?0, #preferred
+```
+
+A less trivial case arises when the allocation is for a generic type instantiated using a skolem.
+For instance, assume the following instruction was existentialized under the same conditions:
+
+```hylo
+%r0 = alloca T[2], #preferred
+```
+
+Because `%p0` is a witness of `T` (i.e., `?0` in the existentialized signature) and not `T[2]`, the compiler has to first construct a type witness of the latter before it can replace the `alloca`:
+
+```hylo
+%r3 = access [let] %p0
+%r2 = type_witness (<T> T[2])(%r3)
+%r1 = access [let] %r2
+%r0 = alloca %r1 as ?0, #preferred
+```
+
+Here, `%r2` is assigned to a type witness constructed at run-time.
+The semantics of the `type_witness` are discussed later.
+For now, it suffices to understand that it serves to apply a type constructor (i.e., `<T> T[2]`) to the witness of `?0` that is passed to `%p0`.
+Finally, the resulting new type witness is used to transform the `alloca`, as in the previous example.
+
+Note that the more occurrence of a type parameter in a type expression does not determine whether or not that expression denotes a type whose layout is abstract.
+For example, the type `Pointer<T>` does not have an abstract layout because it contains no stored property whose layout depends on `T`.
+Therefore, no change is necessary to existentialize an `alloca` creating a place for storing instances of `Pointer<T>`.
+
+> A non-generic type may also have an abstract layout if it is defined beyond a resilience boundary.
+> In this case, however, the first lowering phase will already have generated `alloca` instructions parameterized by type witnesses.
+
+#### Function calls
+
+Consider the following use of `swap(_:_:)`, in a monomorphic call site:
+
+```hylo
+public fun main() {
+  var m = 2
+  var n = 4
+  swap(&n, &m)
+}
+```
+
+The use elaborates to `swap<Int>` during typing and is compiled to a type application during lowering.
+Specifically, the emitter produces the following IR:
+
+```hylo
+%r0 = type_apply swap(_:_:)<Int>
+%r1 = apply %r0(%w, %m, %n) => %u
+```
+
+`%r0` is the result of applying the lowered form of `swap(_:_:)` to `Int`.
+It occurs as the callee in the second line, which assumes that `%w` is a `let` access on the evidence that `Int` is `Movable`, that `%m` and `%n` are `inout` accesses to the places whose contents are exchanged, and that `%u` is a `set` access to a place for storing a unit value (i.e., the result of the call).
+
+Existentializing this sequence of instructions requires three modifications to the `apply` instruction at the end.
+First, we should substitute a reference to the existentialized form of `swap(_:_:)` for `%r0`.
+Second, we should pass a type witness of `Int` to the additional parameter this function accepts.
+Finally, we should remove the `type_apply` instruction.
+The resulting IR will be as follows:
+
+```hylo
+%r2 = type_witness Int()
+%r3 = access [let] %r2
+%r4 = place_cast %w as let Movable<?0>
+%r5 = place_cast %m as inout ?0
+%r6 = place_cast %n as inout ?0
+%r1 = apply swap(_:_:).existentialized(%r3, %r4, %r5, %r6) => %u
+```
+
+`%r2` is assigned to a value witnessing the properties of the type `Int`, which is used as a type constructor of arity 0.
+This value is passed to the first parameter of the call in the last line, applies the existentialized form of `swap(_:_:)` rather than its type application.
+The casts whose results are assigned to `%r4`, `%r5`, and `%r6` only serve to appease the type system so that the arguments' types match the parameters'.
+
+No other change is necessary.
+When the existentialized form of `swap(_:_:)` is eventually compiled to LLVM, the code generator will conclude that its parameters should be taken by reference since the size of a skolem is unknown at compile-time.
+
+#### Accesses to stored properties
+
+The way in which the stored properties of a type are laid out in memory depends on the sizes and alignment requirements of these properties.
+If the layout of a type is concrete and not hidden behind a resilience boundary, then the address of a particular stored property can be computed at compile time.
+Otherwise, this address must be computed at run-time by reading layout properties from a type witness.
+
+Hylo IR uses a relatively abstract instruction, called `property`, to express raw accesses to these properties.
+While this instruction abstracts over byte offset computation, it still distinguishes between compile-time and run-time address computation.
+To illustrate, consider an overload of `swap(_:_:)` that would operate on the elements of a pair.
+
+```hylo
+struct Pair<A, B> {
+  var first: A
+  var second: B
+}
+
+fun swap<T is Movable>(elements_of pair: Pair<T, T>) {
+  swap(&p.first, &p.second)
+}
+```
+
+The lowered form of `swap(elements_of:)`'s polymorphic definition, the `first` and `second` properties of the pair are accessed as follows:
+
+```hylo
+%r0 = property "first" of %pair as T
+```
+
+Here, `%pair` denotes the pair passed as an argument and `T` is the type of the property whose address is taken.
+As already discussed, `T` will be replaced by `?0` in the function's existentialized form.
+Just like with `allocate` instructions, the compiler will eventually conclude that `?0` is abstract and transform the instruction accordingly.
+Specifically, the property access will be modified to include a run-time witness of the type whose property is being accessed.
+
+```hylo
+%r1 = access [let] %p0
+%r2 = access [let] %p0
+%r3 = type_witness (<T, U> Pair<T, U>)(%r1, %r2)
+%r4 = access [let] %r3
+%r0 = property "first" of (%pair : %r4) as ?0
+```
+
+`%p0` and `%p1` are both accesses to the place containing the witness of `?0`, which are passed to a type constructor to form a witness of `Pair<?0, ?0>`.
+At run-time, this type witness will be used to determine the byte offset of the pair's `first` property.
+
+Accesses to members of a tuple or buffer are transformed similarly, only for another instruction.
+For example, should `Pair` be defined as an alias for `{T, U}` in source code, then the above example would be lowered as follows:
+
+```hylo
+%r1 = access [let] %p0
+%r2 = type_witness (<T> {T, T})(%r1)
+%r3 = access [let] %r2
+%r0 = subfield [0] of (%pair : %r3) as ?0
+```
+
+*The reason why the type constructor takes only one argument here is explained later, along with the discussion that examines the construction of run-time type witnesses.*
+
+#### Witness tables
+
+Before we delve into the details of witness table existentialization, we should first discuss how conformance declarations are lowered from source to IR.
+
+Consider the following program:
+
+```hylo
+trait P {
+  fun foo() -> Int
+  fun bar() -> Int { 1 - self.foo() }
+}
+
+given wa: Bool is P {
+  fun foo() -> Int { if self { 1 } else { 0 } }
+  fun bar() -> Int { if self { 0 } else { 1 } }
+}
+```
+
+A conformance declaration can be understood as a subscript that projects a value witnessing the conformance of a type to a trait.
+In the above example, `wa` declares a conformance of `Bool` to `P`, which can be seen as a subscript projecting an instance of `P<Bool>`.
+The value of this instance is a *witness table* describing how `P`'s requirements are implemented.
+
+Unsurprisingly, then, a conformance declaration is lowered like a subscript:
+
+```hylo
+fun wa() <: let P<Bool> {
+%b0:
+  %r0 = witness_table {
+    $implementation[P.foo for Self: Bool],
+    $implementation[P.bar for Self: Bool]
+  } as P<Bool>
+  %r1 = access [let] %r0
+  %r2 = yield %r1
+  %r3 = end %r1
+  %r4 = return
+}
+
+fun $implementation[P.bar for Self: Bool](
+  let %p0: P<Bool>, let %p1: Bool, set %p2: Int
+) {
+  ...
+  %r3 = apply wa.bar(%r1) => %r2
+  ...
+}
+
+fun wa.bar(let %p0: Bool, set %p1: Int) { ... }
+```
+
+The instruction `witness_table` creates a witness table gathering the functions implementing the trait's requirements.
+Notice that these functions, which are called *implementation interfaces*, have a signature slightly different from those statisfying the trait requirements in the original source.
+Specifically, they accept an additional parameter of type `P<Bool>`, which denotes the witness itself.
+The reason for this mechanism will become more obvious later.
+For now, one can think of an implementation as a method of the witness that forwards calls to some other function.
+For instance, the implementation interface `$implementation[P.foo for Self: Bool]` simply forwards its arguments to `wa.foo`.
+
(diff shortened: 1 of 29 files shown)


dabrahams:
I suspect this work may need separate prose documentation.

Sources/BackEnd/Program+LLVM.swift
```diff
@@ -581,10 +583,9 @@ extension Program {
 
   /// Generates the LLVM IR code corresponding to `i`.
   ///
-  /// The instruction is compiled as a direct call iff the subscriptsubscript being applied is an
-  /// addressor. Otherwise, code dominated by `i` is compiled into a different function that is
-  /// passed as a callback to the subscript. The call to this subscript returns the identifier of
-  /// the basic block to which control-flow should be transferred, if any.
+  /// Code dominated by `i` is compiled into a different function that is passed as a callback to
+  /// the subscript. The call to this subscript returns the identifier of the basic block to which
+  /// control-flow should be transferred, if any.
```
dabrahams, line 588:
```suggestion
  /// control flow should be transferred, if any.
```

Sources/FrontEnd/IR/Mangling/DemangledSymbol.swift
```diff
@@ -148,6 +148,9 @@ internal indirect enum DemangledEntity: Hashable, Sendable {
   /// An IRFunction with `name == .existentialized`.
   case existentialized(DemangledEntity)
 
+  /// An IRFunction with `name == .applied`.
+   case applied(DemangledEntity, Int)
```
tothambrus11:
```suggestion
  case applied(DemangledEntity, Int)
```

Docs/Depolmorphization.md
```diff
@@ -1,9 +1,395 @@
...
+Yet, the signature of the trait requitement did not change and neither did the type of the implementation expected at call sites.
+While existentialization can produce a monomorphic version of this polymorphic function, calling it will require a type witness of the type argument instantiating the conformance.
+Unlike in the previous case, however, this type witness is not constant and thus we cannot simply modify the implementation interface.
+
+In fact, the type witness that should be passed to the existentialized form of `foo` is first passed to the existentialized form of the conformance declaration.
+To illustrate, let us examine a use of `wb` lowered to IR:
+
+```hylo
+%r0 = type_witness Bool()
+%r1 = access [let] %r2
```
RishabhRD, line 339:
```suggestion
%r1 = access [let] %r0
```

Docs/Depolmorphization.md
```diff
@@ -1,9 +1,395 @@
 # Deploymorphization
 
 A function (or type) is polymorphic iff it accepts type parameters and monomorphic otherwise.
-The implementation (aka definition) of a monomorphic function may feature uses of polymorphic constructs that have to be either monomorphized or existentialized before IR can be compiled
-to machine code.
-This transformation is referred to as *depolymorphization*.
+The implementation (aka definition) of a monomorphic function may feature uses of polymorphic constructs that have to be either *monomorphized* or *existentialized* before IR can be compiled to machine code.
```
dabrahams, line 4: It's unfortunate that monomorphization isn't the general term, but rather only one of two methods of making a polymorphic function monomorphic.

I don't know how much we've thought about mixing monomorphization with existentialization (i.e. only some of the type parameters are represented by witnesses).

dabrahams: FWIW, the convention I learned is to **boldface** terms where they are first introduced and defined, and *italicize* them elsewhere when it's important to distinguish that they are referring to the term of art and not some other known meaning.

Docs/Depolmorphization.md
```diff
@@ -1,9 +1,395 @@
...
+These transformations are referred to as *depolymorphization*.
 
 Both monomorphization and existentialization essentially consist of creating a copy of the polymorphic function.
-In the former case, type arguments are simply substituted for their corresponding parameters. In the latter case, the type parameters are transformed into term parameters accepting type witnesses at run-time.
+In the former case, type arguments are simply substituted for their corresponding parameters.
+In the latter case, the type parameters are transformed into term parameters accepting type witnesses at run-time.
+
+This document describes how Hylo implements depolymorphization.
```
dabrahams:
```suggestion
The rest of this document describes how Hylo implements existentialization.
```

Docs/Depolmorphization.md
```diff
@@ -1,9 +1,395 @@
...
+
+> Disclaimer:
+> The syntax of Hylo IR is far from final.
+> Snippets are intended as illustration rather than specification.
+
+## Existentialization
+
+In a nutshell, existentialization consists of replacing type parameters with term parameters while abstracting over any information that cannot be determined at compile-time.
```
dabrahams:
```suggestion
Existentialization replaces type parameters with term parameters that provide any information that cannot be determined at compile-time.
```

Docs/Depolmorphization.md
```diff
@@ -1,9 +1,395 @@
...
-The implementation (aka definition) of a monomorphic function may feature uses of polymorphic constructs that have to be either monomorphized or existentialized before IR can be compiled
-to machine code.
-This transformation is referred to as *depolymorphization*.
+The implementation (aka definition) of a monomorphic function may feature uses of polymorphic constructs that have to be either *monomorphized* or *existentialized* before IR can be compiled to machine code.
+These transformations are referred to as *depolymorphization*.
 
 Both monomorphization and existentialization essentially consist of creating a copy of the polymorphic function.
-In the former case, type arguments are simply substituted for their corresponding parameters. In the latter case, the type parameters are transformed into term parameters accepting type witnesses at run-time.
+In the former case, type arguments are simply substituted for their corresponding parameters.
+In the latter case, the type parameters are transformed into term parameters accepting type witnesses at run-time.
```
dabrahams, line 9:
```suggestion
A polymorphic function cannot be lowered directly to LLVM or machine code.   First, it must be **depolymorphized**, transforming it into functions with the same essential structure as the original function, but no type parameters.  The two methods of depolymorphization are called *monomorphization* and *existentialization*.

**Monomorphization** creates a new function for each set of type arguments passed to the polymorphic function, substituting those arguments for the corresponding parameters.  It is simple and efficient, but is not applicable to all generic code and is incompatible with resilience, since the each new function must be compiled into the module of its caller.

**Existentialization** replaces each type parameter with one or more term parameters representing type and conformance witnesses.  These ordinary function parameters allow the transformed function to be compiled into the module of the callee, with polymorphism supplied by dynamic witness values.
```

Docs/Depolmorphization.md
```diff
@@ -1,9 +1,395 @@
...
+
+```hylo
+fun swap<T is Movable>(a: inout T, b: inout T) {
+  var x = a ; a = b ; b = x
+}
+```
+
+The first step to compile this function is to lower it to Hylo IR.
+Recall a conformance constraint such as `T is Movable` is always compiled to term parameter accepting a conformance witness.
+As a result, the signature of the lowered form of `swap(_:_:)` is as follows:
```
dabrahams:
```suggestion
Because the IR cannot represent constraints directly, Hylo replaces each constraint such as `T is Movable` with a conformance witness parameter:
```

Docs/Depolmorphization.md
```diff
@@ -1,9 +1,395 @@
...
+
+```hylo
+fun swap(_:_:)<T>(
+  let %p0: Movable<T>, inout %p1: T, inout %p2: T, set %p3: Void
+)
+```
+
+The function is still polymorphic at this stage, taking a type parameter `T`.
+`%p0` accepts a witness of `T`'s conformance to movable.
+`%p1` and `%p2`, correspond to `a` and `b`, respectively.
```
dabrahams, line 39: Since we are using only a notional IR, why can't we keep the original names `a` and `b`?  One might ask why that capability is not in the actual IR too!

Docs/Depolmorphization.md
```diff
@@ -1,9 +1,395 @@
...
+As a result, the signature of the lowered form of `swap(_:_:)` is as follows:
+
+```hylo
+fun swap(_:_:)<T>(
+  let %p0: Movable<T>, inout %p1: T, inout %p2: T, set %p3: Void
+)
+```
+
+The function is still polymorphic at this stage, taking a type parameter `T`.
+`%p0` accepts a witness of `T`'s conformance to movable.
```
dabrahams:
```suggestion
`%p0` accepts a witness of `T`'s conformance to `Movable`.
```

Docs/Depolmorphization.md
```diff
@@ -1,9 +1,395 @@
...
+`%p1` and `%p2`, correspond to `a` and `b`, respectively.
+Finally, `%p3` denotes the output register of the function, i.e., the place to which its result is written.
```
dabrahams, line 40:
```suggestion
(Hylo IR also does not directly represent return values; they are instead transformed into `set` parameters like `%p3`)
```

Docs/Depolmorphization.md
```diff
@@ -1,9 +1,395 @@
...
+
+Existentialization will produce a copy of this function in which the type parameter `T` is replaced with a term parameters:
```
dabrahams:
```suggestion
Existentialization then replaces the unconstrained type parameter `T` with a term parameter, `T`'s **type witness**:
```

Docs/Depolmorphization.md
```diff
@@ -1,9 +1,395 @@
...
+
+```hylo
+fun swap(_:_:).existentialized(
+  let %p0: Type, let %p1: Movable<?0>, inout %p2: ?0, inout %p3: ?0, set %p4: Void
+)
+```
+
+Rather than taking a type parameter, this function accepts a *type witness* as its first parameter `%p0`.
```
dabrahams:
```suggestion
```

Docs/Depolmorphization.md
```diff
@@ -1,9 +1,395 @@
...
+This type witness will be used to obtain information about the layout of `T`, as we shall see later.
```
dabrahams:
```suggestion
The type witness supplies information about `T` not given by conformances, such as its layout.
```

Presumably other things like a string description could live there.

Docs/Depolmorphization.md
```diff
@@ -1,9 +1,395 @@
...
+The following parameters are the same as those of the polymorphic version, just renamed.
```
dabrahams, line 51: Again could avoid renaming and needing to explain it, and the reader having to keep track.  Just generate mnemonic names such as `%T`

Docs/Depolmorphization.md
```diff
@@ -1,9 +1,395 @@
...
+Finally, occurrences of `T` have been replaced with `?0`, which denotes a *skolem* representing some type unknown at compile time.
```
dabrahams, line 52: I never quite got the purpose of skolems, but maybe some light begins to dawn.  Am I understanding that the purpose of using a skolem instead of say, `Any`, here is to keep track of type correpondences among the term parameters that would otherwise be erased?  
How is `?0` actually different from `T`?
If it is different, why not name `?0` `?T` instead?

I doubt it would answer my questions above succinctly, but not a bad place to add a wikipedia link for “skolem”

Docs/Depolmorphization.md
```diff
@@ -1,9 +1,395 @@
...
+
+> From a more formal perspective, existentialization relies on the observation that a term of type `∀⍺.τ`, where `⍺` is a type variable that may occur in `τ`, can be transformed into a term of type `∃⍺.ω → τ` in which the extra parameter witnesses `⍺`'s properties at runtime.
```
dabrahams:
```suggestion
> Formally, existentialization transforms a term of type `∀⍺.τ`, where `⍺` is a type variable that may occur in `τ`, into a term of type `∃⍺.ω → τ` in which the extra parameter witnesses `⍺`'s properties at runtime.
```

Docs/Depolmorphization.md
```diff
@@ -1,9 +1,395 @@
...
+> One can then [skolemize](https://en.wikipedia.org/wiki/Skolem_normal_form) existentially quantified variable to get rid of the quantifier, resulting in a term of type `ω → τ` in which each occurrence of `⍺` has been replaced with a skolem (aka rigid variable).
```
dabrahams:
```suggestion
> It then [skolemizes](https://en.wikipedia.org/wiki/Skolem_normal_form) the existentially quantified variable to get rid of the quantifier, resulting in a term of type `ω → τ` in which each occurrence of `⍺` has been replaced with a skolem (aka rigid variable).
```

Docs/Depolmorphization.md
```diff
@@ -1,9 +1,395 @@
...
+>
+> For example, given a function of type `∀⍺.⍺ → ⍺`, we first construct a function `∃⍺.ω → ⍺ → ⍺` before applying skolemization to replace occurrences of `⍺` by some unique skolem `κ` and end up with a term of type `ω → κ → κ`, which is a monomorphic function.
```
dabrahams:
```suggestion
> For example, given a function of type `∀⍺.⍺ → ⍺`, we construct a function `∃⍺.ω → ⍺ → ⍺`, we then skolemize it, replacing occurrences of `⍺` by with a unique skolem `κ`, and end up with a monomorphic function of type `ω → κ → κ`.
```

Docs/Depolmorphization.md
```diff
@@ -1,9 +1,395 @@
...
+>
+> Note that nothing in the type system that links `ω` to neither `⍺` nor `κ`.
```
dabrahams:
```suggestion
> Note that nothing in the type system links `ω` to either `⍺` or `κ`.
```

Docs/Depolmorphization.md
```diff
@@ -1,9 +1,395 @@
...
+> As a consequence, the compiler cannot verify the type safety of a function after it has been existentialized.
```
dabrahams, line 60: That consequence is not obvious based on what I've just read above.  This is essentially a form of type erasure and I can sometimes erase type information without compromising static type safety (maybe worth adding "static" though I know it's implied in a formal context).  The times when I can't, there are essentially equality constraints in the original generic code.  So I know the consequence to be true but I don't get there just by reading your text.

Docs/Depolmorphization.md
kyouko-taiga: - Noted for the convention.
- I agree monomorphization would be a nice general term but I'm afraid the definition has already been claimed 😔 
- I gave some thoughts about mixing existentialization and monomorphization.
  - The two techniques can be used in concert to deal with cases where one type parameter can't be monomorphized but another can. For example `fun f<T, U>() { if g() { f<T[2], U>() } }`. I think we can implement some of the reasoning necessary to choose the right technique with some lightweight analysis on the call graph.
  - There may be other heuristics to consider to figure out which technique is best (e.g., number of run-time calls to `member_offset`). We'll need benchmarks to tell.

Docs/Depolmorphization.md
kyouko-taiga: > Since we are using only a notional IR, why can't we keep the original names a and b? 

Yes, I'll do that in a future pass after I resolve your comments.

> One might ask why that capability is not in the actual IR too!

Once we'll have the ability to parse IR we'll have to support that feature. For now register are simply named after their position in the list containing all instructions.

Docs/Depolmorphization.md
kyouko-taiga: Why the parentheses?

Docs/Depolmorphization.md
kyouko-taiga: Very nice change!

Docs/Depolmorphization.md
kyouko-taiga: Agreed.

Docs/Depolmorphization.md
kyouko-taiga: > Am I understanding that the purpose of using a skolem instead of say, `Any`, here is to keep track of type correpondences among the term parameters that would otherwise be erased?

Yes. We want to "remember" that a the result of a function `∀⍺.⍺ → ⍺` has the same type as that of its argument. That is not something we can do with `Any` (assuming subtyping or a comparable theory).

> How is `?0` actually different from `T`?

It's a very keen question. The short answer is that `?0` is quantified existentially whereas `T` is quantified universally. The distinction matters because the existentially quantified variable represents one specific type (i.e., the one witnessed by some other value) whereas the universally quantified represents any arbitrary type.

The longer answer is that the distinction tells us how these variables can be unified when we're comparing two types to determine whether they match.

There are three sorts of type variables, those that are free, those that are bound, and the skolems:
- Free variables are not introduced by any quantifier and can unify with anything. For instance, `⍺` is free in `⍺ → ⍺` and so it can unify with `Int` to conclude that `⍺ → ⍺` matches `Int -> Int`.
- Bound variables can only unify with themselves, modulo structure-preserving renaming. For instance, `⍺` is bound in`∃⍺.⍺ → ⍺` and we can rename it to `β` before concluding that `∃⍺.⍺ → ⍺` matches `∃β.β → β`. The renaming does not change the meaning of the type.
- Skolems are like bound variables, except that we erased their introducer, which was an existential quantifier. 

The intuition behind skolemization is that if you have a formula `∃⍺.P(⍺)` then you can just pick any name not mentioned in your premise to replace `⍺`; that won't change the meaning of your formula. For instance, if you say "knowing that Paul wears a shirt, there exists someone wearing a skirt", then you can equally say "owning that Paul wears a shirt, Sarah wears a skirt".

So we use skolemization simply to get rid of existential quantifiers.

> If it is different, why not name `?0` `?T` instead?

We could and perhaps that will help this presentation. In the compiler it makes sense to normalize variable names to reduce memory footprint and simplify mangling.

Docs/Depolmorphization.md
kyouko-taiga: Here's an example (that I'll include in the text):

```hylo
fun f<T is Movable>(x: T) -> T {
  var y: T = x
  return x
}
```

Here, the type system can guarantee that the number of bytes allocated for `y` matches the number of bytes in the representation of `T`. But after existentialization, it'll be as though `f` had been defined as follows:

```hylo
fun f(w: Type, x: ?T) -> ?T {
  var y = PointerToMutable[UInt8].allocate(count: w.size(), alignment: w.alignment())
  unsafe y.initialize_pointee(to: x)
  return unsafe y.consume_pointee()
}
```

And in this second version, the type system cannot guarantee that `w.size()` returns enough bits to store an instance of `?T`.

Docs/Depolmorphization.md
dabrahams: Because this part is incidental to the depolymorphization we are describing; it applies to all functions.  We just need it to give a full accounting of the parameters

Docs/Depolmorphization.md
dabrahams: >  _⍺_ is free in _⍺_ → _⍺_ and so it can unify with `Int` to conclude that _⍺_ → _⍺_ matches `Int` -> `Int`.

If this were pattern matching it would be a non-symmetric relation in general (`"x"` matches the pattern `[a-z]` but `"[a-z]"` does not match the pattern `x`).  Is matching for unification similar?

There seems to be some inconsistency here:

> occurrences of `T` have been replaced with `?0`, which denotes a *skolem*…
>
>  `?0` is quantified existentially whereas `T` is quantified universally…
>
>  Skolems are like bound variables, except that we erased their introducer, which was an existential quantifier.

So does `?0` really represent a bound variable, which is an *existentially quantified* skolem?  There are other ways to square the logic here but as written there is a contradiction.

> The intuition behind skolemization is that if you have a formula ∃⍺.P(⍺) then you can just pick any name not mentioned in your premise to replace ⍺; that won't change the meaning of your formula. 

Makes sense that ∃*⍺*.P(*⍺*) means the same thing as ∃*β*.P(*β*), but I don't see how it particularly relates to skolemization.

> For instance, if you say "knowing that Paul wears a shirt, there exists someone wearing a skirt", then you can equally say "owning that Paul wears a shirt, Sarah wears a skirt".

Even assuming that you meant s/owning/knowing/ (without which I don't understand the 2nd sentence)… the first sentence differs from the formula in that there's an additional implication: **wears_shirt(Paul) ⇒** ∃*β*.wears_skirt(*β*).  Maybe that's not significant to the example.  But it still seems to mean something very different from the second sentence, which refers to a specific someone, i.e. wears_shirt(Paul) ⇒ wears_skirt(Sarah).  To make the meaning the same you need to say “there exists some _Sarah_…”, which makes _Sarah_ a variable. The first sentence simply is structured to anonymize _β_.

Docs/Depolmorphization.md
dabrahams: I don't think that functions well as an example for the statement as written.  You're making a statement about the effects of skolemization in a formal context, but the transformation to using dynamic stack allocations for locals is not implied by skolemization; I could skolemize that function in a formal sense without introducing the allocate call.  And if I needed to have it, the allocation could return typed memory that preserves the guarantee in the type system.
(Also, In the transformed code, the failed guarantee doesn't look much like a type safety issue.)

At an abstract level, it seems that replacing `T` with `?T` preserves all the information about type correspondences necessary for type checking a function with one type variable.  I suspect it is only the fact(?) that this process drops additional constraints such as `T == U` from the function representation that makes type checking fail.

I wonder whether it's important to make the statement you are trying to justify.

Docs/Depolmorphization.md
```diff
@@ -1,9 +1,393 @@
 # Deploymorphization
```
tothambrus11:
```suggestion
# Depolymorphization
```

Docs/Depolmorphization.md
kyouko-taiga: > If this were pattern matching it would be a non-symmetric relation in general

Unification is symmetric. One situation where you have probably used unification already is simple algebra. Say I ask you to solve `x + 1 = 12` or `10 = x - 1`. In both cases you will likely use unification to tell me that `x` is 11, no matter if it is on the left or the right of the equal sign.

In this example, `x` is a free variable.

> There seems to be some inconsistency here:

Sorry I miss the inconsistency that you found. `T` is bound to a universal quantifier. `?0` *was* bound to an existential quantifier that we removed from the formula. So it is *like* an existentially bound variable, but since its quantifier is gone we can no longer rename it.

The key part is that a skolem won't unify with anything but itself, like a bound variable (existentially or universally quantified) and unlike a free variable. The difference between a skolem and an existentially quantified variable is just about the presence of the quantifier.

Skolemization is a simple trick and we're even lucky enough to only have to deal with the most trivial cases. Specifically, we never encounter an existential quantifier nested under a universal one. So, [as Wikipedia says](https://en.wikipedia.org/wiki/Skolem_normal_form), we can replace existentially quantified variables by new constants. I decided to name these constants `?0`, `?1`, ... but we could also use `?T`, `?U`, ... or anything else.

To understand the justification, we have to look at deductive systems. Instead of using a fresh constant, we could also replace an existentially quantified variables by any instance in its domain. That's what I wanted to say with Paul and Sarah. I realize my example requires some jumps to be enlightening so I'll try to do better.

First, at the risk of being a little pedantic, note that by "knowing `P`, `Q` is true" I meant `P ⊢ Q` rather than `P ⇒ Q`. That is, I meant provability and not implication. The distinction is subtle but important because some rules of our system can only be expressed in the metatheory, not in the theory itself. If we dig enough we'll find Gödel but I'm afraid I'll be out of my depth very soon. The high bit is that we can devise rules about `P ⊢ Q` that we cannot justify with only intuitionistic second-order logic (i.e., the logic to which most type systems are equivalent).

Now, let's try to prove `P(a) ⊢ ∃β ∈ {a, b} . Q(β)`. Because `β`'s domain is closed, there is a finite number of cases to consider. We can instantiate `β` with `a` and then we get `P(a) ⊢ Q(a)`, or with `b` and then we get `P(a) ⊢ Q(b)`. The trick Thoralf Skolem taught us is that our choice does not matter, even if the two statements right of the turnstile are different. As nicely put by Wikipedia, that's because we're interested in satisfiability, not equivalence. In this particular case, it turns out that neither `P(a) ⊢ Q(x)` nor `P(a) ⊢ Q(y)` holds. That is not surprising: knowing that someone wears a shirt does not tell us identify who's wearing a skirt, even if we can name everyone in the world.

Now, let's try to prove `Q(b) ⊢ ∃β ∈ {a, b} . Q(β)`. Again, we can do the same trick and choose either `Q(b) ⊢ Q(a)` or `Q(b) ⊢ Q(b)`. The second statement holds and that is enough to say that we've succeeded. Again, that is not surprising: knowing that Sarah wears a skirt lets us conclude that there exists someone wearing a skirt.

*Earlier I that "[we] can just pick any name not mentioned in your premise to replace `⍺`". That was inaccurate. The restriction only applies if we're getting rid of an existential quantifier on the **left** of the turnstile, which is not the case in the above example.*

This trick continues to work even if `β` ranges over an infinite domain. Back to our system, that is why we're allowed to say that a term of type `∃β.ω → β` can be typed `ω → ?0`. We can also observe the fact that `P(ω → τ) ⊢ ∃β.P(ω → β)` justifies why we can insert `place_cast` instructions to convince the type system that it's fine to call `ω → ?0 → ?0` with a witness of `Int` and an integer.

dabrahams: Re: the inconsistency, just read the excerpt sI pasted carefully.  I added emphasis.

> a. occurrences of T have been replaced with **?0, which denotes a skolem**…
>
> b. **?0 is quantified existentially** whereas T is quantified universally…
>
> c. Skolems are like bound variables, **except that we erased** their introducer, which was **an existential quantifier.**

It seems to me that one of those 3 statements must be false, i.e.:
1. skolems have the existential quantifiers erased (i.e. are not existentially quantified) (c), and ?0 is existentially quantified (b), but is therefore is _not_ a skolem (¬a)
2. or, skolems have the existential quantifiers erased (c) and ?0 is a skolem (a), but is _not_ existentially quantified (¬b)
3. or, ?0 denotes a skolem (a) and is existentially quantified (b), but skolems have _not_ had the existential quantifiers erased (¬c).

Or, I suppose you might mean something strange by "erased," where you can erase a variable's quantifier but keep the variable quantified, somehow? 🤷 

Based on the additional stuff you wrote, it seems like it must be #2, i.e. statement b is false?

kyouko-taiga: I see. You are right. Statement 2 and 3 are false. Skolems are the result of removing the quantifier of an existentially quantified variable. So they _were_ existentially quantified but aren't anymore. They are constant.

Apologies for the contradiction.

Docs/Depolmorphization.md
kyouko-taiga: The note about the less of well-typedness relates to existentialization, not skolemization. The former impacts the signature and the definition. Rewriting allocas is part of this transformation. To make a proper formal argument we'd have to define a term syntax and its semantics. But that would only give us an example that looks more like System F than Hylo.

There are two points I'm trying to make:
1. Like other forms of erasure, existentialization weakens the safety claims we can make about a term. That is probably not very surprising.
2. Because of the way we transform signatures, we need additional care at call site because nothing in the type system guarantees that we pass the right witness to the right parameter.

The example I gave is about (1). An example of (2) stated in the formal world could be that if one applies the existentialized form `id' : ω → ?0 → ?0` of `id : ∀⍺.⍺ → ⍺`, nothing will catch that `id' String (cast ?0 42)` is ill-typed. I think it's worth mentioning because that is a design choice. We could have made the type of a type witness a dependent type.

Docs/Depolmorphization.md
dabrahams: I thought I already understood unification but…

> Now, let's try to prove `P(a) ⊢ ∃β ∈ {a, b} . Q(β)`. Because `β`'s domain is closed, there is a finite number of cases to consider. We can instantiate `β` with `a` and then we get `P(a) ⊢ Q(a)`, or with `b` and then we get `P(a) ⊢ Q(b)`. 

Two other statements.  The original statement is equivalent to their disjunction `P(a) ⊢ Q(a) ∨ P(a) ⊢ Q(b)`.

> The trick Thoralf Skolem taught us is that our choice does not matter, even if the two statements right of the turnstile are different. 

(I assume you mean the two statements `Q(a)` and `Q(b)`?)
If the goal is a proof (which we haven't yet got), it seems to me it very much matters.  For example, if we know only what we've seen so far plus `P = Q`, a proof can only succeed by exploring `P(a) ⊢ Q(a)`.  (I assume that's what you mean by “our choice”).

> As nicely put by Wikipedia, that's because we're interested in satisfiability, not equivalence. In this particular case, it turns out that neither `P(a) ⊢ Q(x)` nor `P(a) ⊢ Q(y)` holds. 

Whence `x` and `y`?

> That is not surprising: knowing that someone wears a shirt does not tell us identify who's wearing a skirt, even if we can name everyone in the world.

I guess P(x) := x wears a shirt and Q(x) := x wears a skirt.

> Now, let's try to prove `Q(b) ⊢ ∃β ∈ {a, b} . Q(β)`. Again, we can do the same trick and choose either `Q(b) ⊢ Q(a)` or `Q(b) ⊢ Q(b)`. The second statement holds and that is enough to say that we've succeeded. 

But the first does not!  So in what sense does the choice “not matter?”  Clearly one leads to a proof and the other doesn't!

dabrahams: > Statement 2 and 3 are false. 

That would mean statement 1 is true:

1. skolems have the existential quantifiers erased (i.e. are not existentially quantified) (c), and `?0` is existentially quantified (b), but is therefore is not a skolem (¬a)

> Skolems are the result of removing the quantifier of an existentially quantified variable. So they were existentially quantified but aren't anymore. They are constant.

…and therefore `?0` is not a skolem?  Are you sure that's what you mean?

kyouko-taiga: In the narrow case for which we use skolems, they are constants and not existentially quantified variables. It is relevant to remember that they once were existentially quantified because that is what justifies the casts we need to apply an existentialized function.

kyouko-taiga: > I assume you mean the two statements Q(a) and Q(b)?

Yes

> Whence x and y?

Substitute `a` for `x` and `b` for `y`. I missed this renaming while editing my text.

> I guess P(x) := x wears a shirt and Q(x) := x wears a skirt.

Yes

> But the first does not! So in what sense does the choice “not matter?”

Because [contraction](https://en.wikipedia.org/wiki/Structural_rule) will let you make as many wrong choices as you want until you make the right one. Here is a proof of `Q(b) ⊢ ∃β ∈ {a, b} . Q(β)` using [sequent calculus](https://en.wikipedia.org/wiki/Sequent_calculus): 

```
----------------- (I)
Q(b) ⊢ Q(a), Q(b)
-------------------- (∃R)
Q(b) ⊢ Q(a), ∃β.Q(β)
----------------------- (∃R)
Q(b) ⊢ ∃β.Q(β), ∃β.Q(β)
----------------------- CR
Q(b) ⊢ ∃β.Q(β)
```

(We read that from bottom to top).

I started with contraction on the right to duplicate `∃β.Q(β)`. Then I removed the existential quantifier on the right and chose `a` for `β`; that is an application of skolemization. That's the wrong choice but I can make another attempt. This time around I chose `a` for `β`. Bingo! I concluded the proof since `Q(b)` is on both sides of `⊢`.

Since I can apply contraction as many times as a I want, I can enumerate all possible cases in the domain of `β` and necessarily end up making the right pick. So the choice does not matter in that in only changes the shape of terminating proofs. Any proof that did not conclude is the leaf of an exploration from which we can backtrack. So in that sense, the choice does not matter.

If `β` has an infinite domain, then enumeration is impossible. But conceptually the argument is the same.

dabrahams: Can you please give a direct answer to the question [here](https://github.com/hylo-lang/hylo-new/pull/530#discussion_r3919876824)?  I am  still just trying to figure out how to read what you originally posted.  It seems each new thing you write adds new things to be confused about, so unless we nail things down in the order they come, the stack just grows deeper.

>> But the first does not! So in what sense does the choice “not matter?”
>
> Because [contraction](https://en.wikipedia.org/wiki/Structural_rule) will let you make as many wrong choices as you want until you make the right one.

But this property is not special to skolems!  It's just how disjunctions work in proofs.  If you try to prove `P(x) ∨ true` you can discover `P(x)` leads nowhere and try the other side of the disjunction.  So when you say “the trick Thoralf Skolem taught us is that our choice does not matter,” it doesn't tell me anything about why skolems are important.

No need to talk about finite/infinite domains; it's quite obvious why the examples you've given generalize to infinite domains.

Docs/Depolmorphization.md
dabrahams: - Because ?0 is the same type as ?0 when they appear in different places in the same scope, it is not at all obvious to a casual reader that there is any type erasure going on.  So **if** you want to make point 1, IMO it's essential to talk about how type information is actually dropped. AFAICT it happens in relationships between distinctly-named types.  Even straightforward equality constraints don't require such dropping since you can always use the same existential type for `T` and `U` when you have `T == U`.  It's pretty subtle; you need something like `T.X == U`.
- I still say the example you gave doesn't make point 1 well at all because
  - It uses implementation details of our lowering and the point holds even when there's no IR involved
  - It is made in a context where you're speaking formally and the mapping is non-obvious
  - the specific problem you're illustrating would be easily avoided by a small modification to the IR
- I don't think point 1 by itself is interesting at this point of the document.
- You don't make point 2 directly at all.

>  I think it's worth mentioning because that is a design choice. We could have made the type of a type witness a dependent type.

If you don't come out and say that explicitly, mentioning all of this—especially when it has the weaknesses I've pointed out—just creates an opportunity to get tangled up with no benefit.

Why should anyone care that you can't statically type-check the same things in the transformed code if you have proven that the transformation is correct?  It's not like we would ever run an extra type-checking pass IRL, except in a test or debug builds.

Regardless, I think it's possible to make the points you are making much more cleanly.  I'm not sure I see why doing it formally is hard but one simple way would be to disentangle this from lowering to IR.  Just describe existentialization in terms of Hylo code.  It's easy to show why this can't be statically type checked after existentialization.:

```
fun f<T, U where T.X == U, U is Equatable>(a: T, b: U) -> Bool { a.x == b }
```

If you want to then make  point 2 by saying “obviously you have to get the transformation right” I guess that's easy too… though “you need to do it right” is kinda table stakes for everything innit?

Docs/Depolmorphization.md
dabrahams: In fact, if you take the suggestion I made elsewhere, to separate the description of existentialization from lowering, this issue goes away too.  Just show transformed Hylo code.

Docs/Depolmorphization.md
dabrahams: if you take the suggestion I made elsewhere, to separate the description of existentialization from lowering by  transforming Hylo code, it's more justifiable to use mnemonic names in the transformed code.

Sources/FrontEnd/IR/Mangling/ManglingEncoding.swift
```diff
@@ -567,17 +556,25 @@ internal struct ManglingEncoding: Sendable {
...
+      append(.plateauDeclaration, tagging: s, with: n, to: &output)
     }
   }
 
+  /// Writes an application of `o` tagging `n` with `i` to `output`.
+  private mutating func append(
+    _ o: ManglingOperator, tagging n: IRFunction.Name, with i: Int,
+    to output: inout ManglingContext
+  ) {
+    output.add(operator: .plateauDeclaration)
```
tothambrus11, line 573:
```suggestion
    output.add(operator: o)
```
We should add a roundtrip test for this.

Sources/BackEnd/Program+LLVM.swift
```diff
@@ -774,26 +766,80 @@ extension Program {
...
+    // Store the captures, if any.
+    if !s.captures.isEmpty {
+      // The computation of an individual capture's offset is justified by the assertions made
+      // during the computation of the storage's size and alignment.
+      let stride = ctx.module.llvm.layout.pointerSize
+      for (i, c) in s.captures.enumerated() {
+        let x0 = codegen(c, in: &ctx)
+        let x1 = ctx.module.llvm.insertGetElementPointerInBounds(
+          of: table, typed: tableRawType,
+          indices: [entriesSize + (i * stride)], indexType: ctx.module.llvm.i32,
```
tothambrus11, line 815: If indices count in  bytes, the `typed` should be `i8`.

There should be some test cases testing various number of captures.

Tests/CompilerTests/CompilerTests.swift
```diff
@@ -5,20 +5,17 @@ import StandardLibrary
 import Utilities
 import StableCollections
 import Subprocess
-import Testing
+import XCTest
 
 /// `true` iff intermediate compilation artifacts shall be saved for successful tests.
-private let alwaysSaveArtifacts: Bool = false
+private let alwaysSaveArtifacts: Bool = true
```
tothambrus11, line 11:
```suggestion
private let alwaysSaveArtifacts: Bool = false
```

Sources/FrontEnd/IR/IREmitter.swift
```diff
@@ -912,18 +922,41 @@ internal struct IREmitter {
     /// The IR values in the representation of `self.`
     private let operands: [IRValue]
 
+    /// If `self` is bound, the index of the value in `operands` denoting its receiver; otherwise,
+    /// the length of `operands`.
+    let receiver: IRValue?
```
tothambrus11, line 927: Why do we need to store the length of `operands`? Is there a better way than repurposing this property?

Sources/FrontEnd/IR/Instructions/IRWitnessTable.swift
```diff
@@ -1,6 +1,16 @@
...
 
 /// Creates a witness table.
+///
+///     witness_table
+///       ['<' <arguments... : type> '>']
+///       ['[' <captures... : value> ']']
+///       (<entries... : value>) as <witness : type>
+///
+/// `witness_table` assembles an object witnessing of the conformance of a type to some trait by
+/// gathering `entries`, `aruguments`, and `captures`, which list the implementations the trait's
```
tothambrus11, line 11:
```suggestion
/// gathering `entries`, `arguments`, and `captures`, which list the implementations the trait's
```

Sources/FrontEnd/IR/Instructions/IRWitnessTable.swift
```diff
@@ -1,6 +1,16 @@
 import Archivist
 
 /// Creates a witness table.
+///
+///     witness_table
+///       ['<' <arguments... : type> '>']
+///       ['[' <captures... : value> ']']
+///       (<entries... : value>) as <witness : type>
+///
+/// `witness_table` assembles an object witnessing of the conformance of a type to some trait by
```
tothambrus11, line 10:
```suggestion
/// `witness_table` assembles an object witnessing the conformance of a type to some trait by
```

Sources/FrontEnd/IR/Instructions/IRWitnessTable.swift
tothambrus11, line 11:
```suggestion
/// gathering `entries`, `aruguments`, and `captures`, which list the implementations of the trait's
```

Sources/BackEnd/Program+LLVM.swift
```diff
@@ -774,26 +766,80 @@ extension Program {
...
+    let tableRawType = ctx.module.llvm.arrayType(storageSize, ctx.module.llvm.i8)
+    let storage = ctx.module.llvm.insertAlloca(tableRawType, atEntryOf: ctx.result.value)
+    ctx.module.llvm.setAlignment(storageAlignment, for: storage)
 
-    unimplemented(if: !entries.allSatisfy(\.unsafe[].isConstant),
-      """
-      Runtime-defined IRWitnessTable operand. https://github.com/hylo-lang/hylo-new/issues/342
+    // Store the entries.
+    let entries = s.entries.map({ (x) in codegen(x, in: &ctx) })
+    let requirements = ctx.module.llvm.structConstant(of: entriesType, aggregating: entries)
```
tothambrus11, line 796: IIRC `structConstant` requires all entries to be constants, that's why the `unimplemented` assertions was there to provide a better diagnostic rather than some internal LLVM crash/UB. Can we add a test case that exercises the non-constant case?

Sources/FrontEnd/IR/IREmitter+Deploymorphization.swift
```diff
@@ -184,27 +219,81 @@ extension IREmitter {
...
+    _ poly: IRFunction
+  ) -> IRFunction.ID {
+    // Existentialize the polymorphic function.
+    let mono = demandExistentialized(poly)
+
+    // Declare the function forwarding the call to the existentialized function.
+    let types = poly.signature.typeParameters
+    let terms = poly.signature.termParameters
+    var applied = IRFunction(
+      name: .applied(program[module].ir[mono].name, 0), // TODO compute a discriminator
```
tothambrus11, line 261: What would that discriminate between?

Sources/BackEnd/SubscriptDecomposition.swift
```diff
@@ -601,7 +601,7 @@ extension IRFunction {
   /// Returns the relevance of `i`, which is an instruction in the prologue of a slide or plateau.
   fileprivate func relevanceInSlideOrPlateau(_ i: AnyInstructionIdentity) -> InstructionRelevance {
     switch tag(of: i) {
-    case IRAlloca.self, IRProject.self:
+    case IRAlloca.self, IRProject.self, IRWitnessTable.self:
```
tothambrus11, line 604: Now IRWitnessTable.self is in two branches, we should remove one of them.

tothambrus11 (changes requested):
Nice work!
I would like to see a lot more tests that exercise the new behavior before merging. Now there are many things that can break silently.

Docs/Depolmorphization.md
dabrahams: @kyouko-taiga I realize this discussion is getting difficult but if we wait much longer to press forward and resolve it I fear we will lose the context and all the progress we've made so far.

PR #531, opened 2026-08-29
Support for compiling the frontend to WASM

tothambrus11:
I corrected some incorrect assumptions regarding the size of `Int`, and upgraded BigInt to support the WASM32 target. I split the Utilities module into two, so that Utilities only includes the non-hosted things (E.g. subprocess creation is not supported in WASM).

This is the first step towards making an online playground and struct layout explorer at https://abiexplorer.org


Sources/FrontEnd/Syntax/ScopeIdentity.swift
```diff
@@ -34,7 +34,7 @@ public struct ScopeIdentity: Hashable, Sendable {
 
   /// `true` iff `self` represents a file.
   public var isFile: Bool {
-    representation.offset == UInt32.max
+    (representation.bits >> 32) == UInt64(UInt32.max)
```
tothambrus11, line 37: The Int value of `offset` overflows because of the extra sign bit. This is a case where using `UInt` for array indices would have been better.

Sources/FrontEnd/Types/TypeStore.swift
```diff
@@ -703,17 +705,17 @@ public struct TypeStore: Sendable {
   /// Projects the type identified by `n`.
   internal subscript(n: AnyTypeIdentity) -> any TypeTree {
     get {
-      switch n.offset {
-      case AnyTypeIdentity.error.offset:
+      switch n {
+      case AnyTypeIdentity.error:
```
tothambrus11, line 709: Are identities having the same offsets imply having the same properties?

kyouko-taiga: Identities with the same offset are equal. The other bits encode properties of the identified type tree.

PR #532, opened 2026-08-30
Interpreter: remove machine independent details from TargetABI

RishabhRD:
The size of `MachineType` is mostly target-independent i.e., it is determined by type itself instead of target except the pointer-sized types. The old `TargetABI` protocol allowed to customize the type of these target-independent types too and thus layouts for these target independent types needed to be passed dynamically despite of them being statically known.

[This Comment](https://github.com/hylo-lang/hylo-new/pull/527#discussion_r3844263237) on #527 shows an example.


Sources/Interpreter/TargetABI.swift
```diff
@@ -20,33 +47,13 @@ struct UnrealABI: TargetABI {
...
-      case .float16: 16
-      case .float32: 32
-      case .float64: 64
-      case .float128: 128
-      case .ptr: bitsInAWord
-      }
-    return layout(bitWidth: bitWidth)
-  }
+  /// The size of a pointer in bytes.
+  var pointerSize: Int { bitsInAWord / 8 }
```
tothambrus11: I would rename `word` as it can mean too many things. It's definitely not necessarily equal the pointer's width.

https://tratt.net/laurie/blog/2021/static_integer_types.html#native_word_width

(This can be done in a later PR, it shouldn't hold back this one.)

Sources/Interpreter/TargetABI.swift
```diff
@@ -3,8 +3,35 @@ import FrontEnd
...
+    case .float64: 8
+    case .float128: 16
+    case .ptr: pointerSize
+    }
+  }
+
   /// Returns the layout of `t`.
-  func layout(_ t: MachineType) -> TypeLayout.Bytes
+  public func layout(_ t: MachineType) -> TypeLayout.Bytes {
+    .init(alignment: alignment(t), size: size(t))
```
tothambrus11, line 31: An ABI should be able to decide the required alignment for its types. E.g. on 8-bit AVR, all integer types have alignment 1:
https://abiexplorer.org/#2.PcfRasJAEAXQXxmur6PYpBWZtH3zLxbCZhM0uBlDdjZRxH8vWPTpcO6YIQUjQbDqNcTcdvSdrO3VNqdfp06TTTkYHa5-GGNHd6dEuVfb10a-eu1jVxs175ZFbRTe3X3WRm3l9FE5BSNCEMBI1kJw1FyUYBgEfp7WWc96WfQlGCMEYAwNZMtI3T_Lk2V8coVgPQwh_3izXm_7LzAmiOYYGfMCgfkm4fEH

RishabhRD: I don't understand the point here. Isn't [this line](https://github.com/hylo-lang/hylo-new/pull/532/changes#diff-ceb336307445a8ebf438596b7adda96d3adb4f7976354674ac92115b20837538R9) enabling that? And this layout method just uses the alignment target ABI provides.

Sources/Interpreter/TargetABI.swift
RishabhRD: This is there for being consistent with `MachineType` [definition](https://github.com/hylo-lang/hylo-new/blob/cd58bbb9b2c65f281c6d3ad6ff857435a9a9b24a/Sources/FrontEnd/Types/MachineType.swift#L13) which defines a word to have same width as pointer.

Sources/Interpreter/TargetABI.swift
tothambrus11: Ah I see, my bad. It's done correctly.

Sources/Interpreter/TargetABI.swift
tothambrus11: Ah, that's used for `Int`. We discussed that we should have `Int` be the equivalent of C++'s `ssize_t`, so that it's the type appropriate for indexing an array. I would rename the builtin type to `size` in the future.

Sources/Interpreter/TargetABI.swift
```diff
@@ -3,8 +3,35 @@ import FrontEnd
...
+  var pointerSize: Int { get }
+
+}
+
+extension TargetABI {
+
+  /// Returns the size, in bytes, of `t` on `self`.
+  public func size(_ t: MachineType) -> Int {
+    switch t {
+    case .i(let w): Int((w + 7) / 8)
```
tothambrus11: This may deserve a named function on BinaryInteger:

```swift
extension BinaryInteger {

  /// Returns the smallest `n` such that `n * divisor >= self`.
  ///
  /// - Requires: `self >= 0` and `divisor > 0`.  
  func dividedRoundingUp(by divisor: Self) -> Self {
     let (quotient, remainder) = self.quotientAndRemainder(dividingBy: divisor)
     return quotient + (remainder > 0 ? 1 : 0)
  }

}
```

Sources/Interpreter/TargetABI.swift
```diff
@@ -20,33 +47,13 @@ struct UnrealABI: TargetABI {
...
-    precondition(
-      bitWidth > 0 && bitWidth.nonzeroBitCount == 1,
-      "bit width \(bitWidth) is not a power of 2.")
-    let sizeInBytes = (bitWidth + 7) / 8
-    return .init(
-      alignment: min(sizeInBytes, maxAlignment),
-      size: sizeInBytes)
-  }
+  /// The alignment of `.i(1)` and `.i(8)` is required to be one byte.
+  func alignment(_ t: MachineType) -> Int { min(size(t), maxAlignment) }
```
tothambrus11: I don't see how the implentation ensures what the documentation requires. Why do we have `maxAlignment`? Are we concerned that `size(t)` may ever return an alginment that's too large?

RishabhRD: > I don't see how the implentation ensures what the documentation requires.

What exactly is unclear? The `UnrealABI` design is alignment = size in bytes till a maximum limit. And documentation requires alignment of `.i(1)` and `.i(8)` to be 1 byte and they are one byte as their size is 1 byte. Am I missing something in your question?

> Are we concerned that size(t) may ever return an alginment that's too large?

Yes. If there exists something like `.i(256)` in future. (It is an "unreal" abi :wink:) It was an explicit design choice.

Sources/Interpreter/TargetABI.swift
```diff
@@ -3,8 +3,35 @@ import FrontEnd
...
+    case .float32: 4
+    case .float64: 8
+    case .float128: 16
+    case .ptr: pointerSize
+    }
+  }
+
   /// Returns the layout of `t`.
-  func layout(_ t: MachineType) -> TypeLayout.Bytes
+  public func layout(_ t: MachineType) -> TypeLayout.Bytes {
```
dabrahams: (Aside: This does look weird.  Maybe I didn't choose the best name with `Bytes` but I don't have a better idea)

Sources/Interpreter/TargetABI.swift
```diff
@@ -20,33 +47,13 @@ struct UnrealABI: TargetABI {
...
-  private func layout(bitWidth: Int) -> TypeLayout.Bytes {
-    precondition(
-      bitWidth > 0 && bitWidth.nonzeroBitCount == 1,
-      "bit width \(bitWidth) is not a power of 2.")
-    let sizeInBytes = (bitWidth + 7) / 8
-    return .init(
-      alignment: min(sizeInBytes, maxAlignment),
-      size: sizeInBytes)
-  }
+  /// The alignment of `.i(1)` and `.i(8)` is required to be one byte.
```
dabrahams: I'm not sure how to read this line. Is it intended to be a precondition?  Is it really just an implementation comment that explains why the current implementation is OK?

Sources/Utilities/BinaryInteger+Extensions.swift
```diff
@@ -6,4 +6,12 @@ extension BinaryInteger {
     return (r == 0) ? self : self + (n - r)
   }
 
+  /// Returns the smallest `n` such that `n * divisor >= self`.
+  ///
+  /// - Requires: `self >= 0` and `divisor > 0`.
```
dabrahams, line 11: Should you add precondition checks for these, or at least assertions?

dabrahams (approved):
See comments

Sources/Interpreter/TargetABI.swift
tothambrus11: I'd call it `Footprint` or `StorageRequirements`.

Sources/Interpreter/TargetABI.swift
dabrahams: I wonder if `leastMultiple(exceeding:)` would be better?

Sources/Interpreter/TargetABI.swift
dabrahams: Size can mean even more things than Word.  Maybe we should use `bitsInAPointer` instead.

Sources/Interpreter/TargetABI.swift
tothambrus11: Ah I see, this is a concrete property of the UnrealABI. Could you add this design decision to the type's or size's documentation?

I would leave out the docs about `.i(1)` since we know that the minimum alignment in any Hylo program is 1. That is not specific to UnrealABI or this function. There is nothing special about i(8) that distinguishes it from i(16), so I would leave that out from the description.

Sources/Interpreter/TargetABI.swift
dabrahams: The latter is long but exceedingly clear.  I'd support that (if you add the missing "t")!

Sources/Interpreter/TargetABI.swift
dabrahams: Nothing other than this comment says it couldn't have 2 byte alignment, right?

tothambrus11: We should just add that types with size smaller than one byte have alignment 1.

Sources/Interpreter/TargetABI.swift
RishabhRD: I exactly copied the `TargetABI` protocol's documentation which states this requirement, but this kind of doc on implementation for sure looks weird.

I should have phrased it like:

```swift
  /// Returns the ABI-required alignment of t, guaranteeing one-byte alignment for
  /// .i(1) and .i(8).
  func alignment(_ t: MachineType) -> Int { min(size(t), maxAlignment) }
```

Sources/Interpreter/TargetABI.swift
RishabhRD: I feel `dividedRoundingUp(by: )` is faster for me to understand what function does.

Sources/Interpreter/TargetABI.swift
RishabhRD: > Could you add this design decision to the type's or size's documentation?

Should we? I feel like "alignment is size till an upper limit" is an implementation detail. I don't want the clients of `UnrealABI` to depend on this. My thought process is start with a weaker postcondition and strengthen it only when it is necessary.

Sources/Interpreter/TargetABI.swift
dabrahams: How about updating in both places to

```
/// Returns the alignment of `t`, which is 1 for `.i(1)` and `.i(8)`.
```

?

I don't think “ABI-required” is adding anything.

Sources/Interpreter/TargetABI.swift
dabrahams: I don't think it's as good. "3 divided, rounding up by 4" is the only way to make it grammatical and is confusing and not the intended meaning whereas ”4's least multiple exceeding 3“ is perfectly grammatical and not confusing

RishabhRD: I see. Although, `exceeding` somewhat implies strictly greater relationship, which is not what we want. Also, I guess we want *multiplier* not multiple for our purposes. How about `leastMultiplier(reaching:)`?

dabrahams: You're right about exceeding, I overlooked that. Should be multiple, not multiplier. 6 is a multiple of 3. Consider `atLeast` instead of reaching. Well, that's a little awkward. `firstMultiple(thatIsAtLeast:)` or `leastMultiple(greaterThanOrEqualTo:)` work for me

Sources/Interpreter/TargetABI.swift
tothambrus11: I think most clients will use an ABI through the TargetABI protocol, and UnrealABI is just a specific instance of ABI which on its own does something very specific. I would document explicitly what that specific thing is, so that member methods can be implemented coherently.

Sources/Interpreter/TargetABI.swift
RishabhRD: > Should be multiple, not multiplier. 6 is a multiple of 3.

But we wanted to abstract `(w + 7) / 8`, which I understand as computing the least multiplier of 8 `k` such that `8 * k >= w`. In my view, for `w = 8`, the multiplier is 1 (which we want), but the multiple is 8. Is there a gap in my understanding here?

dabrahams: Oh, you're quite right!  It took me a long time to see that.  But then the names I'm proposing don't work because the multiplier is not at least the argument value.

After turning many possibilities over in my head, maybe the solution is `w.roundingUpDivision(by: 8)`

dabrahams: or `w.roundingUpQuotient(whenDividedBy: 8)` or `w.roundingUpQuotient(divisor: 8)` but many people get dividend/divisor confused, or `w.roundingUp(dividingBy: 8)`.

I guess `8.minMultiplier(suchThatProductIsAtLeast: w)` or `8.minMultiplier(whereProductIsAtLeast: w)` also work.

I like `w.roundingUp(afterDividingBy: 8)` for clarity.  It's not like this is going to get used a lot, and it reads very naturally.

PR #536, opened 2026-08-31
Naming improvements from @tothambrus.

dabrahams:
Plus some documentation tweaks.

### DIFF
--- a/Sources/Interpreter/TypeLayout.swift
+++ b/Sources/Interpreter/TypeLayout.swift
@@ -63,8 +63,7 @@ struct TypeLayout: Regular {
 
 extension UnsignedInteger {
 
-  /// Returns the least multiple of `n` greater than or equal to
-  /// `self`.
+  /// Returns the least multiple of `n` greater than or equal to `self`.
   ///
   /// - Precondition: `n > 0`.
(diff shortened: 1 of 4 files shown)


Sources/Interpreter/TypeLayout.swift
```diff
@@ -63,7 +63,8 @@ struct TypeLayout: Regular {
 
 extension UnsignedInteger {
 
-  /// Returns `self` rounded up to the nearest multiple of `n`.
+  /// Returns the least multiple of `n` greater than or equal to
+  /// `self`.
```
tothambrus11:
```suggestion
  /// Returns the least multiple of `n` greater than or equal to `self`.
```

tothambrus11 (changes requested):
Thanks, LGTM.

Left a comment about formatting.

Sources/Interpreter/TypeLayout.swift
dabrahams: Rilly? Wotevs!

dabrahams:
IMO you should have approved it with the comment. "Request changes" usually means to me someone wants to review again afterwards.

PR #541, opened 2026-09-04
Interpreter: support memory copy

RishabhRD:
This is required for implementing function calls


Sources/Interpreter/Interpreter.swift
```diff
@@ -262,9 +262,10 @@ public struct Interpreter {
       _ = x
     case let x as IRLoad:
       let v = try load(from: x.source)
-      return .initializeRegister(to: .init(v))
+      return initializeRegister(to: v)
```
dabrahams: Isn't “initialize” kind of superfluous given that all instructions have a result that goes in a new register?  Maybe it should just be `.result(v)`?

Sources/Interpreter/Interpreter.swift
```diff
@@ -341,6 +342,15 @@ public struct Interpreter {
     try memory.store(self[v], at: access(of: p))
   }
 
+  /// Copies the bytes of object at address `source` to address `destination`.
+  ///
+  /// - Precondition: `source` and `destination` are non-overlapping.
```
dabrahams: What do you think of this?

```suggestion
  /// Copies the bytes of the object at `source` to `destination`.
  ///
  /// - Precondition: `source` and `destination` are non-overlapping addresses.
```

Hmm, IIRC addresses are points in memory with no extent and can't overlap.  Did we change that?  I know we discussed it but don't remember the resolution.  I don't see any type information here so presumably that's encoded in source and destination somehow… but then I see we have `TypedAddress` so I'm a little confused.

Sources/Interpreter/Memory.swift
```diff
@@ -222,6 +222,33 @@ struct Memory {
       yield &allocation[i]!
     }
   }
+
+  /// Returns the result of calling `body` with raw buffer pointer to bytes in `p`.
```
dabrahams: Don't repeat type information here.


```suggestion
  /// Returns the result of calling `body` on the bytes in `p`.
```

Sources/Interpreter/Memory.swift
```diff
@@ -222,6 +222,33 @@ struct Memory {
...
+    _ p: TypedAddress, havingLayout l: TypeLayout, _ body: (UnsafeRawBufferPointer) -> R
+  ) -> R {
+    let o = self[p.allocation].baseOffset
+    return self[p.allocation].storage.withUnsafeBytes {
+      let s = $0.baseAddress!.advanced(by: o + p.offset)
+      return body(UnsafeRawBufferPointer(start: s, count: l.size))
+    }
+  }
+
+  /// Returns the result of calling `body` with mutable raw buffer pointer to bytes in `p`.
```
dabrahams:
```suggestion
  /// Returns the result of calling `body` on the bytes in `p`.
```

Sources/Interpreter/Memory.swift
```diff
@@ -350,4 +377,34 @@ extension Memory {
     store(v, at: p.location)
   }
 
+  /// Copies the bytes of object at `source` to `destination`.
```
dabrahams:
```suggestion
  /// Copies the bytes of the object at `source` to `destination`.
```

Sources/Interpreter/Memory.swift
```diff
@@ -350,4 +377,34 @@ extension Memory {
...
+    let l = layout(source.type)
+    self.withUnsafeBytes(source, havingLayout: l) { a in
+      self.withUnsafeMutableBytes(destination, havingLayout: l) {
+        var b = $0
+        b.copyElements(from: a)
+      }
+    }
+  }
+
+  /// Copies the bytes of object at `source` to `destination`.
```
dabrahams:
```suggestion
  /// Copies the bytes of the object at `source` to `destination`.
```

Sources/Interpreter/Memory.swift
```diff
@@ -350,4 +377,34 @@ extension Memory {
...
   }
 
+  /// Copies the bytes of object at `source` to `destination`.
+  ///
+  /// - Precondition: `source` and `destination` are non-overlapping.
+  private mutating func copy(
+    _ source: Memory.TypedAddress,
+    to destination: Memory.TypedAddress
+  ) {
+    precondition(source.type == destination.type)
```
dabrahams, line 387: Are you sure this is legit?  If the destination is uninitialized memory, it may not be legal to use copyElements here.  You would have to initialize the bytes.  IIRC there are bulk initialization methods on pointers that you can use and you could dispense with `copyElements`, which would be a win regardless of whether the current code is legal.

dabrahams (approved):
Some thoughts inline

Sources/Interpreter/Interpreter.swift
RishabhRD: I’m okay with `.result`, but one thing that worries me is that other `InstructionEpilogue` cases, like `.jump`, could also be considered a result of an instruction. It’s just not a value-producing result, but rather a control-flow/action result.

Sources/Interpreter/Interpreter.swift
RishabhRD: > Hmm, IIRC addresses are points in memory with no extent and can't overlap. 

Yes, using the word "address" was a mistake from my side. It is indeed a "place". I know that we are using `TypedAddress` name instead of `Place`, but saying "typed address" in documentation is a little awkward.

> I don't see any type information here so presumably that's encoded in source and destination somehow

From our past discussion, `IRValue` is a notional index of a value. `source` and `destination` identify the typed address of that value.

> but then I see we have TypedAddress so I'm a little confused.

This weak type information function exists for readability at call site, otherwise at call site I would need to write:

```swift
copy(access(to: x.source), to: access(to: x.target))
```

Thus, it just delegates the call to function that has enough type information.

Sources/Interpreter/Memory.swift
RishabhRD: Used `copyMemory` of pointer as destination is always initialized.

Sources/Interpreter/Interpreter.swift
dabrahams: `.value` or `.register`?  I just don't like that it's imperative

Sources/Interpreter/Interpreter.swift
dabrahams: Ah the memory knows what types are  there.

dabrahams:
```suggestion
  /// Copies the bytes of the object at `source` to `destination`.
  ///
  /// - Precondition: `source` and `destination` are non-overlapping places.
```

Sources/Interpreter/Memory.swift
dabrahams: Keep in mind that the alignment requirements of the ABI run in the interpreter may not match those of the host, so you really need to be copying raw bytes…and, for that matter reading/writing raw bytes everywhere except in registers.  I'm not sure what `copyMemory` does…

RishabhRD: > you really need to be copying raw bytes…and, for that matter reading/writing raw bytes everywhere except in registers

Yes, I am aware of this. `copyMemory(from source)` copies the raw bytes of `source`, except if `[source, source + source.count)` is bound to a type `T`. For our current code, load and save of integers would create problems, because we do host-dependent reads and writes. However, I raised #544 to fix this issue.

dabrahams: I'm confused.  How does being bound to a type change what needs to be done?

RishabhRD: Sorry, my previous phrasing was inaccurate. Let me retry.

`copyMemory(from: source)` copies raw bytes from source. The documentation has preconditions based on the type that the destination (`self`) memory is bound to. In our case, since the `storage` is `[UInt8]`, those bytes are bound to `UInt8`, so those preconditions are always satisfied.

The problem is with reading/writing integers in `storage`. Currently, that relies on:

```swift
     storage.withUnsafeMutableBytes { p in
        body((p.baseAddress! + baseOffset + a).assumingMemoryBound(to: T.self))
      }
```
Even if we set the alignment issue aside, `assumingMemoryBound(to:)` requires the memory to already be bound to `T`. But that's never the case here: our `storage` is `[UInt8]`, so the memory is always bound to `UInt8`, not `T`. Thus using `assumingMemoryBound(to: T.self)` here is UB.

#544 fixes both the alignment issue and this memory-binding issue.

NOTE: I am specifically talking about integers because currently we only support integer builtin types.

RishabhRD:
As all the comments are resolved, merging this PR.

PR #544, opened 2026-09-04
Interpreter: Interpreting integers independent of host machine

RishabhRD:
Currently, while reading or writing integers in memory we relied on `assumingMemoryBound(to:)`. However, host alignment might be different than ABI alignment. So, this is an machine dependent UB.

Also, endianness is a property of ABI that was currently missing and is important for reading or writing integers independent of host machine.

The PR fixes the above issues, by providing endianness as an ABI property and doing host independent interpretation of integers.

### DIFF
--- a/Sources/Interpreter/Endianness.swift
+++ b/Sources/Interpreter/Endianness.swift
@@ -1,0 +1,9 @@
+/// The order in which the bytes of multi-byte values are arranged in memory.
+public enum Endianness {
+
+  /// Places the least significant byte at the lowest address.
+  case little
+
+  /// Places the most significant byte at the lowest address.
+  case big
+}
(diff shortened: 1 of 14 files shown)


dabrahams:
Does this handle pointers too?  All the `Builtin` types need to be covered by this (except, I guess, bytes).

RishabhRD:
Currently pointers are not supported. Also, pointers need additional design decisions. Because, size of pointer can vary with ABI, but we need to store `AllocationID + offset` instead of raw value. My initial idea for pointers is to store the pointer data out of line instead of in `storage` array.

tothambrus11:
I think the interpreter can have one concrete pointer size, as dicated by the target platform we are interpreting code for. The frontend-level target selection is still missing for that.

PR #547, opened 2026-09-07
Specify alignment for memcpy, enable cross-compilation testing for ARM32

tothambrus11:
In LLVM 20, an ARM optimization [made an invalid assumption](https://github.com/llvm/llvm-project/blob/47addd4540b4c393e478ba92bea2589e330c57fb/llvm/lib/Target/ARM/ARMTargetTransformInfo.cpp#L1170-L1171) that alignment will be provided for memcpy. This is fixed in https://github.com/llvm/llvm-project/pull/138356, but until we upgrade, we must supply memcpy's alignment attribute. It will also likely help optimizations anyway, so it's not just a temporary fix.

I extended the test suite to test on ARM 32, using qemu to emulate the foreign processor. I had to disable the LLVM IR assertions when testing using cross-compilation, since that changes the alignment and size of everything. I think we could add per-target assertion file extensions, such as `.host.llvm-ir.expected`, `.pointer-64.llvm-ir.expected` or `.arch-amd64.llvm-ir.expected`.


Tests/CompilerTests/CompilerTests.swift
```diff
@@ -23,6 +23,19 @@ struct CompilerTests {
...
+  private static let crossCompilationTarget = Host.environment["HYLO_TEST_TARGET"]
+
+  /// The qemu executable running the compiled test programs, if applicable.
+  ///
+  /// The program's path is appended to this command. For instance, setting it to "qemu-arm" runs
+  /// the test programs under qemu's user-mode emulator for 32-bit ARM.
+  private static var testRunner: Executable? {
+    Host.environment["HYLO_TEST_RUNNER"].map({ (r) in
+      r.contains("/") ? .path(.init(r)) : .name(r)
+    })
```
kyouko-taiga:
```suggestion
    Host.environment["HYLO_TEST_RUNNER"].map { (r) in
      r.contains("/") ? .path(.init(r)) : .name(r)
    }
```

kyouko-taiga (approved):

Issue #545, opened 2026-09-05
hc with no parameters produce errors


JoMazM:
Hi,

I managed to build hylo-new compiler. Calling the compiler with parameters usually works. For example:
hc --version
hc --help
hc file.hylo

Calling hc with no parameters gives the following errors:

When clang is found in the path:
```
/usr/bin/ld: /usr/lib64/gcc/x86_64-pc-linux-gnu/16/../../../../lib64/Scrt1.o: in function `_start':
(.text+0x1b): undefined reference to `main'
clang: error: linker command failed with exit code 1 (use -v to see invocation)
```

When clang is not in the path:
```
../hylo-new/.build/x86_64-unknown-linux-gnu/debug/hc
Error: Executable "clang" is not found or cannot be executed.
```

I wonder why does hylo compiler calls clang when nothing was passed as parameter.

Also, can you explain how clang is used by the hylo compiler?

Thanks :D

tothambrus11:
Thanks for catching this! The command line driver interprets any number of extra arguments that are not flags as files. We haven't checked for whether that's an empty list. That should be an easy fix in CommandLine's `validate()` function.

We use clang as our linker driver at the moment. It knows how to locate the required C standard library / runtime. We are open to suggestions on how we could improve this. You can also invoke `hc` without linking, and then you can use a custom linker. See https://hylo-lang.org/docs/user/tooling/build-systems/ (also very much open to feedback!)

PR #548, opened 2026-09-07
Emit an error when hc is invoked with no arguments

JoMazM:
This fixes the following issue:

https://github.com/hylo-lang/hylo-new/issues/545

Now, It will print the message below when it is invoked with no arguments:
```
Error: Expected argument
Usage: hc [<options>] [<inputs> ...]
  See 'hc --help' for more information.
```


tothambrus11:
Thanks for the fix, LGTM!
Could you add a test case in https://github.com/hylo-lang/hylo-new/blob/main/Tests/CommandLineTests/CommandLineValidationTests.swift?

Sources/hc/CommandLine.swift
```diff
@@ -199,6 +199,9 @@ private typealias Module = FrontEnd.Module
 
   /// Checks that the parsed arguments form a consistent configuration.
   public func validate() throws {
+    if (inputs.isEmpty) {
+      throw ValidationError("Expected argument")
```
```suggestion
      throw ValidationError("expected argument")
```

tothambrus11:
I see a current test is failing that's asserting bad exit code on no input. We can change that test case. You can run the tests locally via `swift test` or make a PR inside your own fork to test the changes.

Sources/hc/CommandLine.swift
JoMazM: I will make this change and see how to update the tests. I'm still getting familiar with swift in general and the project. It may take some time

tothambrus11: Thanks, no hurry. Feel free to ping me on Slack if you need any help, I'm happy to chat.

For learning Swift, I used its language tour which was quite helpful when I started.

JoMazM:
I moved the validation to run as there are tests that only call parse without validate. I also included a new test. Let me know if you have a better idea to implement it

tothambrus11:
> I moved the validation to run as there are tests that only call parse without validate. 

Validation logic is supposed to live inside `validate`. Note that the only test that should ever trigger this behavior is something that we add now (or the previous obsolete test that just asserted an exit code; we can remove that).

JoMazM:
Hi @tothambrus11 ,


I'm re-taking again this issue. When I move the check to validate as below, 

 the tests below fail due to using hc directly effectively bypassing validate().
[  func testModuleEmission() ](https://github.com/hylo-lang/hylo-new/blob/7485e844dae3e4f9db835bf337118dae3399ec78/Tests/CommandLineTests/CommandLineValidationTests.swift#L6) 

│Test Suite 'Selected tests' started at 2026-09-14 17:10:51.902
│Test Suite 'CommandLineValidationTests' started at 2026-09-14 17:10:51.903
│Test Case 'CommandLineValidationTests.testCachingOptions' started at 2026-09-14 17:10:51.903
│Test Case 'CommandLineValidationTests.testCachingOptions' failed (0.026 seconds
│Test Case 'CommandLineValidationTests.testImportAndLinking' started at 2026-09-14 17:10:51.929
│Test Case 'CommandLineValidationTests.testImportAndLinking' failed (0.026 seconds)
│Test Case 'CommandLineValidationTests.testInterfaceHashEmission' started at 2026-09-14 17:10:51.955
│Test Case 'CommandLineValidationTests.testInterfaceHashEmission' failed (0.035 seconds)
│Test Case 'CommandLineValidationTests.testModuleEmission' started at 2026-09-14 17:10:51.990
│Test Case 'CommandLineValidationTests.testModuleEmission' failed (0.058 seconds)
│Test Case 'CommandLineValidationTests.testObjectToStandardOutput' started at 2026-09-14 17:10:52.048
│Test Case 'CommandLineValidationTests.testObjectToStandardOutput' failed (0.021 seconds)
│Test Case 'CommandLineValidationTests.testSelfImport' started at 2026-09-14 17:10:52.069 
|Test Case 'CommandLineValidationTests.testSelfImport' failed (0.066 seconds)

WHat would be your suggestion to address these failures? Notice that I added new test to CommandLineEndToEndTests.swift

tothambrus11:
Hi, thanks for looking into this! 
It turned out that the existing test suite missed to supply a file to compile in many existing tests, so I just corrected them now to reflect the new expectations. I also moved the end to end CLI test to a CommandLineValidationTest, as the latter can be faster, and slightly shorter to write. Keeping it as an CLIEndToEndTest would also have its value, it's higher fidelity, but I'm confident that there is not much special about this specific error output that distinguishes it from other, already end-to-end-tested error conditions.

For the future: the issues you pasted were likely just the summaries, and the detailed test results can be read if you scroll up a bit. Filtering for the failed test suite can clear up the logs a bit: `swift test --filter=CommandLineValidationTests`. Or you can use VSCode's Swift extension, which has nice failure previews.

